PR #36, opened 2021-07-08
Initial bikeshed file

mukul-p:


mukul-p:
For those who are new - This would be (snipped) output of the current config - 
image

Jack-Works (approved):

index.bs
```diff
@@ -0,0 +1,11 @@
+<pre class='metadata'>
+Title: Web Extensions Draft
```
hober: s/ Draft//

index.bs
```diff
@@ -0,0 +1,11 @@
+<pre class='metadata'>
+Title: Web Extensions Draft
+Shortname: webextensions
+Level: 1
+Status: w3c/ED
```
hober: s|w3c/ED|CG-DRAFT|

index.bs
```diff
@@ -0,0 +1,11 @@
+<pre class='metadata'>
+Title: Web Extensions Draft
+Shortname: webextensions
+Level: 1
+Status: w3c/ED
+Group: w3c/webextensions
```
hober: s|w3c/webextensions|WECG|

index.bs
```diff
@@ -0,0 +1,11 @@
+<pre class='metadata'>
+Title: Web Extensions Draft
+Shortname: webextensions
+Level: 1
+Status: w3c/ED
+Group: w3c/webextensions
+URL: https://github.com/w3c/webextensions
```
hober: the URL should be https://w3c.github.io/webextensions/

index.bs
```diff
@@ -0,0 +1,11 @@
+<pre class='metadata'>
+Title: Web Extensions Draft
+Shortname: webextensions
+Level: 1
+Status: w3c/ED
+Group: w3c/webextensions
+URL: https://github.com/w3c/webextensions
+Editor: Mukul Purohit, Microsoft Corporation https://www.microsoft.com, mpurohit@microsoft.com
+Editor: Tomislav Jovanovic, Mozilla, https://www.mozilla.org/, tjovanovic@mozilla.com
```
hober: These two Editor lines use different syntax; have you tried this in Bikeshed and do they produce the same result?

index.bs
hober: Note that this will require a corresponding change to Bikeshed, which I'll submit a PR for today.

index.bs
hober: (It's unnecessary to say that it's a draft in the title, as the Status metadata item establishes that.)

hober (changes requested):

index.bs
hober: https://github.com/tabatkins/bikeshed/pull/2111

hober: nota bene https://github.com/tabatkins/bikeshed/pull/2111#issuecomment-880984568

index.bs
mukul-p: Indeed there was a difference (because of one comma). Thanks

hober (approved):
Looks good to me!

PR #37, opened 2021-07-08
Publish minutes of 2021-07-08 meeting

Rob--W:
Generated from https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit

Using a process similar to #23.

During this meeting, we discussed or mentioned #31, #13, #30, #36, #19, #11, #25.

PR #42, opened 2021-07-15
Publish meeting schedule + time zone converter

Rob--W:
As a follow-up to #37 (https://github.com/w3c/webextensions/blob/main/_minutes/2021-07-08-wecg.md#meeting-notes), here is the schedule of the next meetings. I also included an inclusive time zone converter that is able to convert to dates across the whole globe.

### DIFF
--- a/_minutes/README.md
+++ b/_minutes/README.md
@@ -1,0 +1,23 @@
+# Meetings
+
+The [WebExtensions Community group](https://www.w3.org/community/webextensions/) meets virtually every other week, for one hour.
+See the group's mailing list for [instructions to join the meeting](https://lists.w3.org/Archives/Member/internal-webextensions/2021Jun/0000.html).
+
+To accomodate attendees across the globe, the recurring meeting alternates between two time slots:
+
+* Thursday 8 AM PDT / Thursday 3 PM UTC
+* Thursday 11 PM PDT / Friday 6 AM UTC
+* To convert to your local time zone, see https://everytimezone.com/
+
+After the end of each meeting, meeting notes are published here.
+
+
+## Upcoming meetings
+
+* 2021-07-22 at 11 PM PDT = https://everytimezone.com/?t=60f8b500,708
+* 2021-08-05 at 8 AM PDT = https://everytimezone.com/?t=610b2a00,384
+
+## Past meetings
+
+* 2021-07-08 ([minutes](2021-07-08-wecg.md))
+* 2021-06-24 ([minutes](2021-06-24-wecg.md))


xeenon (approved):

_minutes/README.md
```diff
@@ -0,0 +1,23 @@
+# Meetings
+
+The [WebExtensions Community group](https://www.w3.org/community/webextensions/) meets virtually every other week, for one hour.
+Instructions to join the meeting are sent to the group's mailing list at https://lists.w3.org/Archives/Member/internal-webextensions/
```
dotproto:
```suggestion
Instructions to join the meeting are sent to the group's mailing list at https://lists.w3.org/Archives/Member/internal-webextensions/2021Jun/0000.html
```

dotproto: Might be better to link to the specific email so folks don't have to go digging through the archives.

_minutes/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+To accomodate attendees across the globe, the recurring meeting alternates between two time slots:
+
+* Thursday 8 AM PDT / Thursday 3 PM UTC
+* Thursday 11 PM PDT / Friday 6 AM UTC
+* To convert to your local time zone, see https://everytimezone.com/
+
+After the end of each meeting, meeting notes are published here.
```
dotproto, line 12: Should we include anything here about the process? For example, "the meeting's scribe will submit a PR with the minutes. Attendees are encouraged to review and request edits as appropriate. The PR will be merged after N days."

dotproto (approved):

_minutes/README.md
Rob--W: I linked to the archive itself for now, to be consistent with your blog post at https://www.w3.org/community/webextensions/2021/06/18/wecg-public-meeting-schedule/, where you also linked to the mailing list. We were supposedly going to send an e-mail before every meeting, and perhaps we should do that in case folks aren't following our repository.

_minutes/README.md
Rob--W: Probably not needed. We haven't really established a process. And in the past two meetings, nobody asked for revisions of the minutes on Github.

dotproto: Fair point :)

 In that case, perhaps we should merge very quickly after and address corrections through additional PRs.

_minutes/README.md
dotproto: I would have linked directly to the email if I could have, but the email wasn't ready. As I recall we wanted to get the blog post out as soon as possible, then followed up with the internal email to provide details that could not be publicly shared.

I think a regular email a few days before the meeting may be a good reminder for group members & serve as confirmation that the meeting is taking place as planned, even if it is consistent. In that case, failure to send an email would be a signal that our normal process has been interrupted. The main drawback is that this creates manual work for group organizers (even if it is relatively small edits to an existing template).

Alternatively, we could have a static doc in the repo that outlines the meeting schedule and future planned meetings. The Credentials CG does something like this in their [announcements repo](https://github.com/w3c-ccg/announcements/tree/8240293b264a6b60e6401ecb933feb8b8ac8b5b3). I use a static link here because their upcoming events table is currently empty.

dotproto:
Looks like we don't have any outstanding concerns, so I'm going to merge this PR.

Issue #45, opened 2021-07-23
Participating in TPAC 2021 


dotproto:
This week the group's chairs received an email announcing that W3C has begun organizing TPAC 2021. Like last year, this will be a completely remote event. The following comes from the announcement email.

> TPAC 2021 will be virtual, and will be held from 18 to 29 October.
> \*The week from 25 to 29 October will be dedicated to the Joint and Group 
> meetings.*
> The format we propose focuses on collaborative meetings that bring together the community 
> and create momentum and brainstorming.

Full details on this event can be found at https://www.w3.org/wiki/TPAC/2021/GroupMeetings. Guidance on the role TPAC serves and recommendations for participation can be found at https://www.w3.org/Guide/meetings/tpac.html.

This is a good opportunity for us to publicise our CG and to begin collaborating with other with other groups relevant to our work. Let's use this issue to begin identifying the groups we want to reach out to, topics we want to discuss, conversations we'd like to organize, etc.

xeenon:
Topics we should consider talking about at TPAC included Service Workers, Web App Manifest, Content Security Policy, Web Transport, and JavaScript for Isolated Worlds / Content Scripts (which would be TC39 and not W3C / TPAC.)

xeenon:
WebDriver BD might be interested in Isolated Worlds and could be at TPAC too.

dotproto:
As discussed in today's meeting, the [GroupMeetings](https://www.w3.org/wiki/TPAC/2021/GroupMeetings) page states, "Until 10 September please fill in the tables below with your preferred meeting dates and time. If flexible, please state a second option."

dotproto:
About an hour ago I completed the process of sending emails to the chairs of the various working groups we're interested in meeting with during TPAC. I'll follow up here as we hear back.

Rob--W:
TPAC 2021 is in the past, TPAC 2022 will be discussed at #232.

Issue #30, opened 2021-06-30
Set up Bikeshed spec pipeline


dotproto:
As [suggested](https://github.com/w3c/webextensions/issues/13#issuecomment-871735082) in #13, we should…

- [x] Add Bikeshed to this repo
- [x] Make sure that the output of spec builds are clearly tagged as draft/WIP
- [x] Set up GitHub page publishing
- [x] Set up a build action
- [x] Set up a deploy action

Once we have this, we can begin drafting drafting the discussing the `manifest.json` specification.

dotproto:
Reference GitHub deploy workflow https://github.com/w3ctag/security-questionnaire/blob/main/.github/workflows/deploy.yml

dotproto:
As noted in today's meeting, we did this a while back.

Issue #13, opened 2021-06-17
Where do we start for a draft?


xeenon:
We need to discuss where we should start documenting a spec draft.

* Should we start at the `manifest.json` format or other another area of the common core of APIs?
* Do we inherit anything from the old CG (issue #1)?
* Should we contribute WebIDL files for the APIs?

therealglazou:
Maybe we first need to discuss how and where we discuss? We have four possible chat sources:

1. GH issues
2. matrix
3. CG mailing-lists
4. regular confcalls ?

dotproto:
@therealglazou, there's a bit of context lost in the transition from our pre-announce efforts to the public group. The primary channels we're currently planning to use are GitHub issues, regular conference calls, and Matrix. We just [published a blog post](https://www.w3.org/community/webextensions/2021/06/18/wecg-public-meeting-schedule/) that outlines the meeting schedule.

* Issues for specific topic and long-running conversations. 
* Conference calls to discuss topics that require higher bandwidth conversations than issues typically allow
* Matrix for general chat, quick questions, pings, etc. We may ask participants to open issues here for tracking purposes. 

We allude to this in the [contributing guide](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md), but we can probably make that clearer.

dotproto:
> Should we start at the `manifest.json` format or other another area of the common core of APIs?

I don't have strong opinions on where to start. As the entry point for extensions, `manifest.json` feels like a natural starting point. Alternatively, we could go breadth first and start collecting the topics that we want to pursue in more detail. This could potentially allow us to parallelize work and identify areas where we need further discussion.

I'd love to hear @zombie and @mukul-p's thoughts here. 

> Do we inherit anything from the old CG (issue #1)?

In a [comment on that issue](https://github.com/w3c/webextensions/issues/1#issuecomment-864184142) I suggested one way that we might handle the previous group's assets.

I don't think we should start from the [Browser Extensions](https://browserext.github.io/browserext/) draft spec as I want to be judicious about the initial common core we pursue. That said, I do think it makes sense to borrow targeted pieces of document where appropriate to accelerate our efforts.

With respect to our drafting efforts, we could do something like the "breadth first" approach I suggested above by extracting the heading structure from the [Browser Extensions](https://browserext.github.io/browserext/) draft spec and using that outline to help guide our conversations. 

> Should we contribute WebIDL files for the APIs?

[...]

PR #46, opened 2021-07-23
Publish minutes of 2021-07-22 meeting

Rob--W:
Generated from https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit

Using a process similar to #23.

During this meeting, we discussed or mentioned #13, #30, #36, and agreed to file new issues (about tracking work on manifest file (TBD) and participating in TPAC at #45).

### DIFF
--- a/_minutes/2021-07-22-wecg.md
+++ b/_minutes/2021-07-22-wecg.md
@@ -1,0 +1,96 @@
+# WECG Meetings 2021, Public Notes, July 22
+
+* Chair: Simeon Vincent
+* Scribes: Rob Wu
+
+Time: 11 PM PDT = https://everytimezone.com/?t=60f8b500,708
+
+Call-in details: https://lists.w3.org/Archives/Member/internal-webextensions/2021Jun/0000.html
+
+Zoom issues?  ping @zombie (Tomislav Jovanovic, Mozilla) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+## Agenda: [github issues](https://github.com/w3c/webextensions/issues)
+
+* [Issue 13](https://github.com/w3c/webextensions/issues/13): Where do we start for a draft? \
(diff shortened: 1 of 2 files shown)


Rob--W:
So far I used the PDT timezone as the primary point of reference (with UTC and time zone converters). Should we continue to do that in this way, or should we start using UTC instead? If we use UTC, then the document should be attributed to `2021-07-23` instead of `2021-07-22`.

mukul-p:
> So far I used the PDT timezone as the primary point of reference (with UTC and time zone converters). Should we continue to do that in this way, or should we start using UTC instead? If we use UTC, then the document should be attributed to `2021-07-23` instead of `2021-07-22`.

UTC would be better choice. We can check what is the precedent.

dotproto:
As I recall, when we last discussed time zones we decided to use Pacific time because the browser member reps either worked in or worked closely with people in that time zone. Targeting the Pacific time zone allowed us to set a single time (8 AM Pacific) without regard for [daylight savings time](https://en.wikipedia.org/wiki/Daylight_saving_time). 

If we were to use UTC, we would need to decide how to handle daylight savings: do we keep maintain a fixed UTC time or adjust the meeting to accommodate daylight savings? If so, what DST region do we accommodate (observation dates vary by location)?

Rob--W:
I'm fine with using PDT as a reference (since we based our meeting schedule on it, and conversions to other time zones are provided in the schedule and meeting notes).

@mukul-p Any objections?

dotproto:
Ah, I was thinking more about how we schedule rather than canonical datetime references. My personal preference would be to schedule around Pacific but for fixed dates and times to use UTC. @xeenon, do you have any thoughts?

dotproto (approved):
LGTM. I'm good with either holding off on this merge until we resolve the UTC question or accepting it as is and updating the timestamp later if necessary.

PR #48, opened 2021-07-26
First draft for the specification of the manifest file

### LINKED ISSUE #47: Kickstart spec work on manifest.json
The previous community group had [a section on manifest.json in their draft](https://browserext.github.io/browserext/#manifest-json). I propose that we use it as a base to discuss what we want/don’t want for our own draft, or even copy (some part of) it if we feel like it’s appropriate.

- Do we agree on the structure of this section?
- Do we want to start from that and adapt it?

For example, I’m not sure that the [5.4 The activeTab permission](https://browserext.github.io/browserext/#active-tab) section has exactly its place in this section. Maybe we would like to describe permissions in another section entirely, especially since the list of permissions varies a lot and depend on API namespace available on each platform.

Keep in mind that the goal is to have some minimal, useful text in the spec, to start iterating on it. :slightly_smiling_face:

ariasuni:
Fix #47 

I started writing the spec for the manifest. I believe we could start by briefly describing the Manifest file, and list the keys and their basic characteristics. Any slightly more advanced thing (version comparison, permissions, key translations, etc.) can be added later.

I’d like to know if I’m going in the right direction before doing more work, so please comment and tell me what you think.

I went for a short list with links to detailed section instead of a table, I think it’s much better to have a short list then the possibly long details in a dedicated section. I also replaced «browsers» par «implementors», since I think it’s more clear and accurate. I took inspiration from the [Web Application Manifest Editor’s Draft](https://w3c.github.io/manifest/) for a few things.

Here’s a screenshot:

image

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -8,4 +8,182 @@ URL: https://w3c.github.io/webextensions
 Editor: Mukul Purohit, Microsoft Corporation https://www.microsoft.com, mpurohit@microsoft.com
 Editor: Tomislav Jovanovic, Mozilla https://www.mozilla.org/, tjovanovic@mozilla.com
 Abstract: [Placeholder] Abstract.
+Markup Shorthands: markdown yes
 </pre>
+
+# Manifest
+
+A WebExtension must have a manifest file at its root directory.
+
+## Manifest file
+
+A manifest file is a [[!JSON]] document named `manifest.json`. Malformed JSON files are not supported.
+
+## Manifest keys
+
+If manifest keys that are not defined in this specification are specified, implementors must ignore those keys.
+
+If manifest keys that are defined in this specification are specified with a different JSON type than defined in this specification, implementors must ignore those keys.
+
+The following keys must be considered valid:
+
+* <a href="#key-manifest_version">`manifest_version`</a>: required.
+* <a href="#key-name">`name`</a>: required.
+* <a href="#key-version">`version`</a>: required.
+* <a href="#key-default_locale">`default_locale`</a>: required under some conditions.
+* <a href="#key-action">`action`</a>: optional
+* <a href="#key-author">`author`</a>: optional
+* <a href="#key-background">`background`</a>: optional
+* <a href="#key-browser_action">`browser_action`</a>: optional
+* <a href="#key-chrome_settings_overrides">`chrome_settings_overrides`</a>: optional
+* <a href="#key-chrome_url_overrides">`chrome_url_overrides`</a>: optional
+* <a href="#key-commands">`commands`</a>: optional
+* <a href="#key-content_scripts">`content_scripts`</a>: optional
+* <a href="#key-content_security_policy">`content_security_policy`</a>: optional
+* <a href="#key-description">`description`</a>: optional
+* <a href="#key-developer">`developer`</a>: optional
+* <a href="#key-devtools_page">`devtools_page`</a>: optional
+* <a href="#key-externally_connectable">`externally_connectable`</a>: optional
+* <a href="#key-homepage_url">`homepage_url`</a>: optional
+* <a href="#key-icons">`icons`</a>: optional
+* <a href="#key-incognito">`incognito`</a>: optional
+* <a href="#key-omnibox">`omnibox`</a>: optional
+* <a href="#key-optional_permissions">`optional_permissions`</a>: optional
+* <a href="#key-options_page">`options_page`</a>: optional
+* <a href="#key-options_ui">`options_ui`</a>: optional
+* <a href="#key-page_action">`page_action`</a>: optional
+* <a href="#key-permissions">`permissions`</a>: optional
+* <a href="#key-short_name">`short_name`</a>: optional
+* <a href="#key-storage">`storage`</a>: optional
+* <a href="#key-theme">`theme`</a>: optional
+* <a href="#key-user_scripts">`user_scripts`</a>: optional
+* <a href="#key-version_name">`version_name`</a>: optional
+* <a href="#key-web_accessible_resources">`web_accessible_resources`</a>: optional
+
+### Key `manifest_version`
+
+This key must be present.
+
+### Key `name`
+
+This key must be present.
+
+### Key `version`
+
+This key must be present.
+
+### Key `default_locale`
+
+This key must be present if the `_locales` subdirectory is present, must be absent otherwise.
+
+### Key `action`
+
+This key may be present.
+
+### Key `author`
+
+This key may be present.
+
+### Key `background`
+
+This key may be present.
+
+### Key `browser_action`
+
+This key may be present.
+
+### Key `chrome_settings_overrides`
+
+This key may be present.
+
+### Key `chrome_url_overrides`
+
+This key may be present.
+
+### Key `commands`
+
+This key may be present.
+
+### Key `content_scripts`
+
+This key may be present.
+
+### Key `content_security_policy`
+
+This key may be present.
+
+### Key `description`
+
+This key may be present.
+
+### Key `developer`
+
+This key may be present.
+
+### Key `devtools_page`
+
+This key may be present.
+
+### Key `externally_connectable`
+
+This key may be present.
+
+### Key `homepage_url`
+
+This key may be present.
+
+### Key `icons`
+
+This key may be present.
+
+### Key `incognito`
+
+This key may be present.
+
+### Key `omnibox`
+
+This key may be present.
+
+### Key `optional_permissions`
+
+This key may be present.
+
+### Key `options_page`
+
+This key may be present.
+
+### Key `options_ui`
+
+This key may be present.
+
+### Key `page_action`
+
+This key may be present.
+
+### Key `permissions`
+
+This key may be present.
+
+### Key `short_name`
+
+This key may be present.
+
+### Key `storage`
+
+This key may be present.
+
+### Key `theme`
+
+This key may be present.
+
+### Key `user_scripts`
+
+This key may be present.
+
+### Key `version_name`
+
+This key may be present.
+
+### Key `web_accessible_resources`
+
+This key may be present.
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+index.html


index.bs
```diff
@@ -1,11 +1,62 @@
...
+URL: https://w3c.github.io/webextensions
+Editor: Mukul Purohit, Microsoft Corporation https://www.microsoft.com, mpurohit@microsoft.com
+Editor: Tomislav Jovanovic, Mozilla https://www.mozilla.org/, tjovanovic@mozilla.com
+Abstract: [Placeholder] Abstract.
+Markup Shorthands: markdown yes
+</pre>
+
+# Manifest
+
+A WebExtension must have a manifest file at its root directory.
```
Jack-Works, line 16: Should we explain the basic idea of how a web extension is loaded into a browser to make `its root directory` have its context?

index.bs
```diff
@@ -1,11 +1,62 @@
...
+
+## Manifest keys
+
+If manifest keys that are not defined in this specification are specified, implementors may ignore those keys.
+
+If manifest keys that are defined in this specification are specified with a different JSON type than defined in this specification, implementors may ignore those keys.
+
+A manifest file must have these keys properly specified:
+
+* <a href="#key-name">`name`</a>: required.
```
Jack-Works, line 31: I think we should follow https://datatracker.ietf.org/doc/html/rfc2119 for words like MAY REQUIRE MUST

index.bs
```diff
@@ -1,11 +1,62 @@
...
+* <a href="#key-version">`version`</a>: required.
+* <a href="#key-default_locale">`default_locale`</a>: required under some conditions.
```
Jack-Works, line 33:
```suggestion
* <a href="#key-default_locale">`default_locale`</a>: See <a href="#key-default_locale">`default_locale`</a>.
```

index.bs
```diff
@@ -1,11 +1,62 @@
...
+
+### Key `description`
+
+This key may be present. Its value must be a JSON string.
+
+It should be used.
+
+### Key `icons`
+
+This key may be present. It value must be a JSON object, whose keys and values are JSON strings.
```
Jack-Works: Should we use JSON Schema to specify those things?

Jack-Works (changes requested):

index.bs
ariasuni: I think we should but later probably, since we don’t have a reference text yet.

index.bs
ariasuni: I deliberately used a shorter formulation so that someone writing a WebExtensions can have a clear overview, with the formal definitions in the sections themselves. I also didn’t want to repeat myself.

index.bs
ariasuni: I’d argue that the spec needs a separate section outside of the manifest definition to define this, and the fact that it can either be a directory or a zip for example.

xeenon: We should define the directory structure in a separate section which would cover the locale directories. However, per this CG’s charter, we have decided the package format (zip, app, etc) are out of scope.

index.bs
```diff
@@ -1,11 +1,62 @@
...
+
+A WebExtension must have a manifest file at its root directory.
+
+## Manifest file
+
+A manifest file is a [[!JSON]] document named `manifest.json`. Malformed JSON files are not supported.
+
+## Manifest keys
+
+If manifest keys that are not defined in this specification are specified, implementors may ignore those keys.
```
xeenon: "…implementors _should_ ignore those keys." since we want a developer to be able to show one manifest to any browser without worry it will be flagged as invalid in some browsers.

index.bs
xeenon: Wouldn't the JSON schema for the manifest be a separate document, and this would be more prose?

xeenon (commented):
This is a good start. Feel free to continue as much as you can. We can also do this iteratively across multiple PRs for more bite-sized reviews.

index.bs
```diff
@@ -1,11 +1,62 @@
...
+* <a href="#key-version">`version`</a>: required.
+* <a href="#key-default_locale">`default_locale`</a>: required under some conditions.
+* <a href="#key-description">`description`</a>: optional.
+* <a href="#key-icons">`description`</a>: optional.
+
+### Key `name`
+
+This key must be present. Its value must be a JSON string.
+
+It should be used as the user-facing name of the extension.
```
oliverdunk: Would something like this be a little clearer? I'm not sure if it's closer to normal spec language or not:

> The browser should use this to identify the extension when displaying it to a user.

index.bs
```diff
@@ -1,11 +1,62 @@
...
+
+Note: TODO describe how versions should be compared
+
+### Key `default_locale`
+
+This key must be present if the `_locales` subdirectory is present, must be absent otherwise. Its value must be a JSON string.
+
+### Key `description`
+
+This key may be present. Its value must be a JSON string.
```
oliverdunk: Should we explain what happens if it isn't present?

oliverdunk (commented):
Exciting to break ground here! I didn't have much to say but took a look just in case it helped. @ariasuni, do you think this needs to stay a draft? I agree with @xeenon that we should merge little and often - partly because I'd love to help, but don't want to conflict.

index.bs
dotproto: @xeenon, I'm actually unclear about that at the moment. I've seen references in the Web Application Manifest issues to parsing content out of the spec file. I'm currently reaching out to Googlers that have contributed to the Web Application Manifest spec to gather information on how that WG has approached JSON specification and integration with browser tooling.

ariasuni:
Ok so I’ve pushed something more minimalist for starters. Hopefully we can figure out how we want to define types, how to describe each key individually, if we want to add some keys, etc.

I think we should open a few issues so we can discuss the questions that were raised here. Does that sound good?

index.bs
```diff
@@ -1,11 +1,205 @@
...
+If manifest keys that are defined in this specification are specified with a different JSON type than defined in this specification, implementors must ignore those keys.
+
+The following keys must be considered valid:
+
+* <a href="#key-manifest_version">`manifest_version`</a>: required.
+* <a href="#key-name">`name`</a>: required.
+* <a href="#key-version">`version`</a>: required.
+* <a href="#key-default_locale">`default_locale`</a>: required under some conditions.
+* <a href="#key-description">`description`</a>: optional.
+* <a href="#key-icons">`description`</a>: optional.
```
xeenon: Should be `icons` or removed since it is also include a second time below.

index.bs
```diff
@@ -1,11 +1,205 @@
...
+* <a href="#key-author">`author`</a>: optional
+* <a href="#key-background">`background`</a>: optional
+* <a href="#key-browser_action">`browser_action`</a>: optional
```
xeenon, line 37: This raises the question, should we document `browser_action` and `page_action` or document the Manifest v3 `action`?

index.bs
```diff
@@ -1,11 +1,205 @@
...
+* <a href="#key-chrome_settings_overrides">`chrome_settings_overrides`</a>: optional
+* <a href="#key-chrome_url_overrides">`chrome_url_overrides`</a>: optional
+* <a href="#key-commands">`commands`</a>: optional
+* <a href="#key-content_scripts">`content_scripts`</a>: optional
+* <a href="#key-content_security_policy">`content_security_policy`</a>: optional
+* <a href="#key-default_locale">`default_locale`</a>: optional
+* <a href="#key-description">`description`</a>: optional
```
xeenon, line 43: `description` is in here twice.

index.bs
```diff
@@ -1,11 +1,205 @@
...
+* <a href="#key-developer">`developer`</a>: optional
+* <a href="#key-devtools_page">`devtools_page`</a>: optional
+* <a href="#key-externally_connectable">`externally_connectable`</a>: optional
+* <a href="#key-homepage_url">`homepage_url`</a>: optional
+* <a href="#key-icons">`icons`</a>: optional
```
xeenon, line 48: `icons` is in here twice.

index.bs
```diff
@@ -1,11 +1,205 @@
...
+* <a href="#key-incognito">`incognito`</a>: optional
+* <a href="#key-omnibox">`omnibox`</a>: optional
+* <a href="#key-optional_permissions">`optional_permissions`</a>: optional
+* <a href="#key-options_page">`options_page`</a>: optional
+* <a href="#key-options_ui">`options_ui`</a>: optional
+* <a href="#key-page_action">`page_action`</a>: optional
```
xeenon, line 54: Same comment as `browser_action`.

index.bs
ariasuni: I think that’s a good idea. Let’s do that after merging this, maybe create an issue?

index.bs
ariasuni: I removed this from the PR, description will need to be in a separate one.

index.bs
ariasuni: I changed it to «must» without thinking too much about it, I think this should discussed for real afterwards, maybe in a separate issue as well?

index.bs
ariasuni: I don’t know Manifest v3 very much so I just put what was existing on several browsers right now. This should indeed be properly discussed…

index.bs
ariasuni: see my comment above: https://github.com/w3c/webextensions/pull/48#discussion_r678683402

index.bs
mukul-p: We can drop the aspect of location of manifest file in this section and define the directory structure later (as suggested by xeenon).

index.bs
```diff
@@ -1,11 +1,193 @@
...
+* <a href="#key-default_locale">`default_locale`</a>: required under some conditions.
+* <a href="#key-author">`author`</a>: optional
+* <a href="#key-background">`background`</a>: optional
+* <a href="#key-browser_action">`browser_action`</a>: optional
+* <a href="#key-chrome_settings_overrides">`chrome_settings_overrides`</a>: optional
+* <a href="#key-chrome_url_overrides">`chrome_url_overrides`</a>: optional
+* <a href="#key-commands">`commands`</a>: optional
+* <a href="#key-content_scripts">`content_scripts`</a>: optional
+* <a href="#key-content_security_policy">`content_security_policy`</a>: optional
+* <a href="#key-default_locale">`default_locale`</a>: optional
```
mukul-p: default_locale is repeated (already defined at no 4)

index.bs
mukul-p: Created https://github.com/w3c/webextensions/issues/50 for discussion

xeenon (approved):

index.bs
dotproto: @Jack-Works, I don't quite follow. What's wrong with this block? It appears to be following the guidance in 2119.

index.bs
dotproto: I'm a little torn. On one hand it doesn't matter where the manifest file itself is located, but on the other it _does_ contain information about filesystem pathing for the other resources it mentions. For the moment I'm okay with the current copy.

index.bs
dotproto: +1 to more issues :)

index.bs
dotproto: I synced with my team on this specific question and the border issue of how we're thinking about Manifest V2 and Manifest V3 in the context of this spec. I'll be following up in #50 soon.

Short answer: include `action`, `browser_action`, and `page_action`.

dotproto (approved):

index.bs
ariasuni: Ok so for the sake of merging this quickly, I’ll let it there (we can certainly move this sentence in another, more appropriate section later).

index.bs
ariasuni: Could you open an issue about this? I believe the question is if we can use a JSON schema file as part of the spec so the «core» text is lighter, right?

index.bs
ariasuni: ah, I never realized it also defined «required» and «optional» also.

index.bs
ariasuni: done!

ariasuni: Should I add `host_permissions` as well then? Since it’s the other manifest key that’s not in manifest v2, as far as I know.

index.bs
dotproto: Will do.

Issue #53, opened 2021-08-05
Decompose manifest.json spec work into action items members can PR


dotproto:
As discussed in our 2021-08-05 meeting, we should decompose the manifest.json specification work in allow group members to work in parallel on the spec.

Issue #50, opened 2021-07-29
Document `browser_action` and `page_action` or document `action` (Manifest v3) in the manifest-schema


mukul-p:
should we document browser_action and page_action or document the Manifest v3 action

Techdojo:
I think we should do both but tag them (and any other MV2 / MV3 specific sections) as appropriate.

dotproto:
I think it's probably best to specify all 3: `browser_action` and `page_action` for Manifest V2 and `action` for Manifest V3. 

More broadly, though, we should probably discuss how our work integrates with Manifest V2, Manifest V3, and whatever may come after that. The Chrome team is primarily interested in pursuing the future of the browser extension platform, but also recognizes that the future is built on the platform that we have today.

To that end our preference is to specify both Manifest V2 and Manifest V3 features, but to limit Manifest V2 inclusions to capabilities that a) have an equivalent in Manifest V3 and b) are common across Manifest V2 implementations. For example, this would mean that when specifying Manifest V2 capabilities we would omit both Chrome's [system.display API](https://developer.chrome.com/docs/extensions/reference/system_display/) and Firefox's [search API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/search).

zombie:
I agree with Simeon, and along the same lines we should treat MV3 features similarly: for now document those that a) have V2 equivalents, and b) once there's a at least a second broadly compatible implementation. In this concrete example, this would include the `action` property, implemented in Chrome and (experimentally, behind a pref) [in Firefox 91](https://bugzilla.mozilla.org/show_bug.cgi?id=1706398) to be released next week. :)

As we progress along with those specifications and implementations, a) would matter less and b) should become the main criteria.

Issue #43, opened 2021-07-20
Proposal for Manifest v3: Have less restrictive security measures for force-installed extensions


nir-walkme:
Manifest v3 introduces some security improvements which are great for consumers. They reduce the risk that an extension installed from the store will harm the user. This is great as most users lack the knowledge and tools to assess the security level of an extension they install from the store.

On the contrary, in an enterprise use case, the Manifest v3 changes limit the functionality of existing extensions. There are many use cases where an extension requires elevated permissions. For example, extensions used for QA, security or digital adoption.
In the enterprise world, extensions are pushed to the end users using an IT policy (either from the store or from a self-hosted location) and the extensions usually go through a security review before being pushed.

Therefore we suggest treating extensions installed manually from the store differently than extensions force-installed.
This is already the case for the blocking version of the webRequest API (https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#when-use-blocking-webrequest) so it appears that our line of thought is in the right direction.

As a general rule, we suggest that any security related restriction in the extension would be excluded from force-installed extensions, or at least there would be a discussion about it.

More specifically, our ask is to allow us to load remotely hosted code into a website's page when using force-installed extensions. We believe this has much use in enterprise use cases.

Rob--W:
@nir-walkme This topic was discussed during the last meeting (https://github.com/w3c/webextensions/blob/main/_minutes/2021-08-05-wecg.md), but we ran out of time. Did you get a satisfactory answer or is there anything else that you'd like to ask/discuss?

nir-walkme:
@Rob--W Thank you for the discussion yesterday. I feel we made progress but did not finalize this issue.

To re-clarify my request following yesterday's discussion:
Every browser has its own way to deploy an extension using IT administrator policy ("Enterprise use case"). The actual deployment method is out of scope of my request.
Every browser "knows" if the extension was installed manually from an extension store (For example: addons.mozilla.org or "chrome web store") or if it was installed via IT policy.

-------------------------------------

So I propose:
1. Add the following API:
GetExtensionInstallMethod() - Returns "Store" / "Enterprise"
This would assist with the next proposal

2. Deprecations in Manifest v3 (or any future version) that were done for security/performance reasons would not be deprecated if the extension was installed by IT administrator policy.
The rationale:
* Extensions installed by IT undergo security and performance review.
* These deprecations break existing extensions which causes high business impact both for the extension developers and customers.
[...]

Issue #21, opened 2021-06-24
How should `sidebarAction` be considered, since it’s only implemented in two browsers?


ariasuni:
Both Firefox and Opera implements a mostly non-overlapping API under the [sidebarAction](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/sidebarAction) namespace (agreeing on the [`sidebar_action`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/sidebar_action) key of the `manifest.json`).

Considering that most browsers don’t implement it, should this be described somehow, and considered part of the main spec while being optional, or an extension of it? Is two implementors sufficient to be considered something we want to describe, for interoperability purposes? Anyway, I guess it could be good to have guidelines for vendor-specific extensions of the WebExtensions API so that the whole stays coherent (and fit for a potential later standardization).

dotproto:
My first impression is that it probably makes sense for the specification effort to approach this as an optional module that individual vendors may choose to adopt. I think existing implementations may provide enough reference material to begin drafting a spec for this namespace, but we may not have enough input from implementers to include it a report (a formal document published by the group).

therealglazou:
Let me just comment that a feature having two publicly shipped interoperable implementations will pass Candidate Recommendation's exit criteria and then qualifies for standardization...

carlosjeurissen:
Naver Whale also implemented sidebar_action. Tho it's using a different syntax. For example, it requires the use of default_page instead of default_panel.

Even tho it tolerates the syntax of Opera and Firefox. Unlike those two browsers, it can't run alongside other action declarations (action, browser_action, page_action).

See: https://developers.whale.naver.com/api/extensions/sidebarAction/

dotproto:
Agreed that we should not specify vendor-specific APIs in order to keep the main specification focused on cross-browser compatibility. Such capabilities may be added to the spec in the future if there's sufficient interest from other browsers.

carlosjeurissen:
Unifying some form of sidebar api is tracked here:
https://github.com/w3c/webextensions/issues/128

PR #54, opened 2021-08-05
Publish minutes of 2021-08-05 meeting

Rob--W:
Generated from https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit

Using a process similar to #23.

During this meeting, we discussed #50, #21, #13, #43, #45. We have also agreed to new file issues (considering a common test suite to support interoperability; decomposing the manifest.json work in actionable tasks that other WECG members can work on as #53).

### DIFF
--- a/_minutes/2021-08-05-wecg.md
+++ b/_minutes/2021-08-05-wecg.md
@@ -1,0 +1,153 @@
+# WECG Meetings 2021, Public Notes, Aug 5
+
+* Chair: Timothy Hatcher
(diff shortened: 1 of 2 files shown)


Rob--W:
I did not only publish meeting notes, but also adjusted the link to the meeting instructions as suggested at the end of the meeting.

xeenon (approved):

Issue #60, opened 2021-08-20
Add `name` and `short_name` descriptions


xeenon:
Add some prose for the `name` and `short_name` manifest keys.

dotproto:
I was looking at the `name` property yesterday and noticed that [MDN notes](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/name) "Google Chrome and Microsoft Edge restrict the name length to 45 characters." It appears that both Chromium and Firefox allow actually allow arbitrary length strings, but CWS and AMO do not allow names longer than 45 characters. I did not test other stores or browsers.

In #24, @twschiller suggested that when specifying the manifest we should "also consider any relevant web store factors." Should we note in the specification that name length is limited to 45 characters?

dotproto:
> Should we note in the specification that name length is limited to 45 characters?

I was originally inlined to say no, as browsers allow longer strings. As I've been considering this, though, it is a practical limit of the ecosystem and therefore something that extension authors need to be cognizant of. Going back to the [Priority of Constituencies](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies), it seems like a requirement worth documenting.

ariasuni:
MDN also advise to not go over 12 characters for [`short_name`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/short_name).

Maybe we could add a note, or a «The `name` value should be kept under 45 characters.»? At least I don’t think it needs in the current PR, we could think about it and add what we want later.

PR #61, opened 2021-08-20
Add the `name` and `short_name` prose.

xeenon:
Fixes issue #60.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -66,7 +66,7 @@ This key must be present.
 
 ### Key `name`
 
-Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See [[#key-short_name]] about declaring a shorter name for some user interface contexts.
+Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also <a href="key-short_name">`short_name` key</a>.
 
 This key must be present. This property can be localized.
 
--- a/index.bs
+++ b/index.bs
@@ -66,7 +66,7 @@ This key must be present.
 
 ### Key `name`
 
-Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also <a href="key-short_name">`short_name` key</a>.
+Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also [`short_name` key](#key-short_name).
 
 This key must be present. This property can be localized.
 
--- a/index.bs
+++ b/index.bs
@@ -168,7 +168,7 @@ This key may be present.
 
 ### Key `short_name`
 
-Short name of the extension used in the browser’s user interface when the [name](#key-name) is too long. If `short_name` is not included, `name` will be used and may be truncated by the browser if needed.
+The short name of the extension. This value should be used in contexts where [`name`](#key-name) is too long to use in full. If `short_name` is not provided, manifest consumers should use a truncated version of `name`.
 
 This key may be present. This property can be localized.
 
--- a/index.bs
+++ b/index.bs
@@ -66,7 +66,9 @@ This key must be present.
 
 ### Key `name`
 
-This key must be present.
+Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See [[#key-short_name]] about declaring a shorter name for some user interface contexts.
+
+This key must be present. This property can be localized.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -166,7 +168,9 @@ This key may be present.
 
 ### Key `short_name`
 
-This key may be present.
+Short name of the extension used in the browser’s user interface when the [name](#key-name) is too long. If `short_name` is not included, `name` will be used and may be truncated by the browser if needed.
```
ariasuni:
```suggestion
Short name of the extension used in the browser’s user interface. This should be used instead of the <a href="key-name">`name` key</a>, when the latter could be truncated.
```

index.bs
```diff
@@ -66,7 +66,9 @@ This key must be present.
 
 ### Key `name`
 
-This key must be present.
+Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See [[#key-short_name]] about declaring a shorter name for some user interface contexts.
+
+This key must be present. This property can be localized.
```
ariasuni: I think we should avoid describing how one key should be used in the description in another.

```suggestion
Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension.

This key must be present. This property can be localized.

See also <a href="key-short_name">`short_name` key</a>.
```

ariasuni (changes requested):
I added some suggestions which I feel improves slightly your initial wording.

index.bs
dotproto: I like @ariasuni's suggestion. For the sake of discussion, though, here's is an alternate framing that's a little closer to the original text.

```suggestion
Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also <a href="key-short_name">`short_name` key</a>.

This key must be present. This property can be localized.
```

@ariasuni's suggestion has an implicit pattern where "see also" links are added to the end of the description. This allows for a more consistent way for readers to see what other topics are associated with a given section of text. My suggestion has an implicit pattern where a concept is explained, then immediately followed by links to related topics. This allows for more contextual association between the concept being explained and the link.

I have a slight preference for my own suggestion, but I wouldn't object if I'm in the minority.

index.bs
dotproto: Between the two, I find @xeenon's original text easier to understand than this suggestion. I do like that this suggestion shortens the first sentence, but I only understood what the second sentence was saying when I tried to break it down and rephrase it in my head.

```suggestion
The short name of the extension. This value should be used in contexts where [`name`](#key-name) is too long to use in full. If `short_name` is not provided, manifest consumers should use a truncated version of `name`.
```

This formation attempts to iterate on @xeenon's copy by making each sentence focus on a specific aspect of the property. I also wanted to allow for non-browser uses such as extension stores, administrative interfaces, or security tools.

dotproto (approved):
I'm happy with the PR in its current state, but it would be best to resolve the outstanding discussions before we merge.

index.bs
xeenon: Is "manifest consumers" here meant to be general enough to apply to stores, etc? I wonder if there is a better name for that, but nothing is coming to mind right now.

index.bs
```diff
@@ -66,7 +66,9 @@ This key must be present.
 
 ### Key `name`
 
-This key must be present.
+Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also <a href="key-short_name">`short_name` key</a>.
```
dotproto:
```suggestion
Name of the extension used in the browser’s user interface. This should be the full name used to identify the extension. See also [`short_name` key](#key-short_name).
```

Sorry, my suggestion accidentally introduced an HTML style link.

ariasuni: Does this work, though? As far as I know, Bikeshed’s Markdown support is very limited and doesn’t handle nested syntax, e.g. code inside link.

dotproto: Oh shoot, I didn't realize! Thank you for calling that out. I just tested locally and you are correct; the inline code block was processed as expected, but the surrounding link was not.

I'll throw together a new PR to fix the stuff we just broke 😅

making787 (commented):
Okay

PR #80, opened 2021-09-15
Create initial deploy.yml

mukul-p:
Using SpecProd to verify the bikeshed file (https://w3c.github.io/spec-prod/)


dotproto:
@mukul-p, is deploy.yml meant to be used with Github Actions? If so, the [Actions reference docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows) state that workflows must be stored in the `.github/workflows/` directory.

deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+      # Runs a set of commands using the spec-prod (https://w3c.github.io/spec-prod/)
+      # the document is built and validated as a check in the pull request.
+      # the document is built and validated, and then deployed to gh-pages branch, when a commit is pushed to the main branch.
+      - uses: w3c/spec-prod@v2
+        with:
+          TOOLCHAIN: bikeshed
+          VALIDATE_MARKUP: true
+          GH_PAGES_BRANCH: gh-pages
+          DESTINATION: specification/index.html
+          # Deployment will be available at: https://<org>.github.io/<repo>/specification/
```
dotproto:
```suggestion
          # Deployment will be available at: https://w3c.github.io/webextensions/specification/
          DESTINATION: specification/index.html
```

Move the comment above the output destination and make the URL specific to our repo.

deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a set of commands using the spec-prod (https://w3c.github.io/spec-prod/)
+      # the document is built and validated as a check in the pull request.
+      # the document is built and validated, and then deployed to gh-pages branch, when a commit is pushed to the main branch.
+      - uses: w3c/spec-prod@v2
+        with:
+          TOOLCHAIN: bikeshed
+          VALIDATE_MARKUP: true
```
dotproto:
```suggestion
```

I think we should either specify all args to be explicit or trust in the defaults provided with w3c/spec-prod@2. Since we have have not specified most flags, I'd suggest that we drop these ones as well.

deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "main"
+  main:
+    name: Build, Validate and Deploy
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
dotproto:
```suggestion
    runs-on: ubuntu-20.04
```

I imagine using `ubuntu-latest` is probably fine, but since spec-prod [Getting started](https://w3c.github.io/spec-prod/#getting-started) instructions explicitly uses I'd suggest we follow their lead.

deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "main"
+  main:
+    name: Build, Validate and Deploy
```
dotproto:
```suggestion
    name: Build, Validate, and Deploy
```

I'm a fan of the [serial comma](https://en.wikipedia.org/wiki/Serial_comma) ;)

dotproto (changes requested):
In addition to my [last comment](https://github.com/w3c/webextensions/pull/80#issuecomment-920510027), I have a couple suggestions on how we might clean this file up a little.

mukul-p:
> @mukul-p, is deploy.yml meant to be used with Github Actions? If so, the [Actions reference docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows) state that workflows must be stored in the `.github/workflows/` directory.

Never trust the (Web) UI :). Use cmd-line or equivalent.

dotproto (approved):

PR #102, opened 2021-10-12
Document basic file structure

### LINKED ISSUE #64: Specify the hierarchy of files
As far as I know, a WebExtensions is a hierarchy a files with:

- As its root, a `manifest.json`
- As its root also, an optional `_locales` directory
- All other files referenced by the manifest or the APIs (e.g. in `tabs.executeScript`) must be in the root or in a subdirectory.

A note could be added that since several browsers distribute extensions as zip, the limitation of this compression format should be taken into account (no need to add more details).

Sounds simple enough to describe, if I didn’t forget anything.

oliverdunk:
Adds basic documentation for the file structure of a WebExtension.

Resolves #64

***
<a href="https://pr-preview.s3.amazonaws.com/1Password/webextensions/pull/102.html" title="Last updated on Oct 12, 2021, 10:04 PM UTC (f459fb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webextensions/102/1a8004b...1Password:f459fb1.html" title="Last updated on Oct 12, 2021, 10:04 PM UTC (f459fb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11,6 +11,24 @@ Abstract: [Placeholder] Abstract.
 Markup Shorthands: markdown yes
 </pre>
 
+# File structure
+
+Once unpacked from the distribution format, a WebExtension is a directory containing a number of files.
+
+Note: In some operating systems, filenames are case insensitive. This can lead to naming collisions.
+
+## manifest.json
+
+A <a href="#manifest">Manifest</a> file.
+
+## _locales subdirectory
+
+An optional directory containing strings as defined in <a href="#localization">localization</a>.
+
+## Other files
+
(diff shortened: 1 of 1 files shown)


Jack-Works:
File names start with `_` expected `_locales` are preserved by Chrome. If you have a `_file.js`, it will be rejected to load. I think we should document that.

Rob--W:
Files/directories starting with `_` do not have a special treatment in Firefox.

How about Safari?

While it makes sense to document the (widely?) supported `_locales/` directory, it may not make as much sense to document something that is only restricted in one browser. It may however make sense to add a comment like "Names starting with `_` are reserved for future use and should not be used by extensions."

mukul-p (approved):

Rob--W (approved):

PR #124, opened 2021-10-29
Publish minutes of 2021-10-28 meeting 

dotproto:
List of issues/PRs in order of appearance in the input:
- Issues
  - #109
  - #119
  - #103
  - #112
  - #44
- PRs: -
- Mentioned issues without link to issue: -

PR #129, opened 2021-11-01
Replace outdated meeting link with calendar list

Rob--W:
~Also added link to last meeting's notes, and~ replaced the link to a past meeting with the calendar list.

@xeenon Could you update https://www.w3.org/groups/cg/webextensions/calendar so that it's in the PDT time zone? It's currently fixed at 3 PM UTC, but after the summer time ends (last weekend for EU, upcoming weekend for US), the meeting is set to happen at 4 PM UTC instead.

Edit: meeting times have been fixed in #124, so this PR is just to update the old link.

### DIFF
--- a/_minutes/README.md
+++ b/_minutes/README.md
@@ -1,7 +1,7 @@
 # Meetings
 
 The [WebExtensions Community group](https://www.w3.org/community/webextensions/) meets virtually every other week, for one hour.
-The instructions to join the meeting and agenda are available at https://www.w3.org/events/meetings/7fc25ca5-a50c-498c-82e5-f48fc96e1637.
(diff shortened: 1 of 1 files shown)


dotproto:
Just updated the recurring event on the W3C calendar to "08:00 - 09:00 America/Los Angeles".

dotproto (approved):
This change makes the link more technically correct, but it's also a bit of a usability loss. The user now has to manually navigate to view the next confirmed event in order to see the meeting details.

Ideally we'd be able to link to the next confirmed event using a URL like `https://www.w3.org/groups/cg/webextensions/calendar/next`. Mentioned this use case her: w3c/calendar#67

PR #186, opened 2022-03-17
Add mock for browser.secureStorage proposal

### LINKED ISSUE #180: browser.secureStorage: Create polyfill for implementation
As mentioned at the [last WECG meeting](https://github.com/w3c/webextensions/pull/171) a polyfill would be a good way for people to experiment with the secureStorage API and get a feel for if the API surface works in their extension. We should look at building something based on what we have today, possibly alongside the proposal in this repo.

oliverdunk:
As suggested in the last meeting, this PR adds a polyfill which shows the expected API behaviour of the browser.secureStorage proposal. This includes the `remove` call added in #183.

For obvious reasons, we can't currently access hardware backed secure storage. As a result, this polyfill uses localStorage and is for proof of concept purposes only.

Resolves #180

### DIFF
--- a/proposals/secure-storage-polyfill.js
+++ b/proposals/secure-storage-polyfill.js
@@ -1,9 +1,9 @@
 let global;
 
-if (window.browser?.runtime.id) {
-  global = window.browser;
-} else if (window.chrome?.runtime.id) {
-  global = window.chrome;
+if (globalThis.browser?.runtime.id) {
+  global = browser;
+} else if (globalThis.chrome?.runtime.id) {
+  global = chrome;
 } else {
   throw new Error(
     "browser.secureStorage polyfill must be run in extension contexts"
@@ -87,11 +87,11 @@ const secureStorage = {
 };
 
 // Attach to browser namespace in Firefox/Safari
-if (window.browser?.runtime?.id) {
-  window.browser.secureStorage = secureStorage;
+if (globalThis.browser?.runtime?.id) {
+  browser.secureStorage = secureStorage;
 }
 
 // Attach to chrome namespace in all browsers
-if (window.chrome?.runtime?.id) {
-  window.chrome.secureStorage = secureStorage;
+if (globalThis.chrome?.runtime?.id) {
+  chrome.secureStorage = secureStorage;
 }
--- a/proposals/secure-storage-polyfill.js
+++ b/proposals/secure-storage-polyfill.js
@@ -11,7 +11,7 @@ const RECOGNISED_AUTH_METHODS = [
   "BIOMETRY_FINGERPRINT",
 ];
 
-window.secureStorage = {
+const secureStorage = {
   getInfo: () => {
     return {
       type: "MACOS_KEYCHAIN",
@@ -65,3 +65,19 @@ window.secureStorage = {
     return localStorage.removeItem(id);
   },
 };
+
+let isExtension = false;
+
+if (window.browser?.runtime?.id) {
+  window.browser.secureStorage = secureStorage;
+  isExtension = true;
+}
+
+if (window.chrome?.runtime?.id) {
+  window.chrome.secureStorage = secureStorage;
+  isExtension = true;
+}
+
+if (!isExtension) {
+  window.secureStorage = secureStorage;
+}
--- a/proposals/secure-storage-polyfill.js
+++ b/proposals/secure-storage-polyfill.js
@@ -1,0 +1,67 @@
+// Existing browser APIs don't give us access to the system's secure storage,
+// meaning all data in this polyfill is stored in less secure mechanisms.
+console.warn(
+  "Warning: browser.secureStorage polyfill loaded. This proof of concept stores data insecurely and should not be used in production."
+);
+
+const RECOGNISED_AUTH_METHODS = [
+  "PIN",
+  "PASSWORD",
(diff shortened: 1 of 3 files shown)


proposals/secure-storage-polyfill.js
```diff
@@ -0,0 +1,67 @@
...
+);
+
+const RECOGNISED_AUTH_METHODS = [
+  "PIN",
+  "PASSWORD",
+  "BIOMETRY_FACE",
+  "BIOMETRY_FINGERPRINT",
+];
+
+window.secureStorage = {
```
xeenon: Should this go on the `browser` and `chrome` objects to be an accurate polyfill?

proposals/secure-storage-polyfill.js
```diff
@@ -0,0 +1,67 @@
...
+  getInfo: () => {
+    return {
+      type: "MACOS_KEYCHAIN",
+      availableAuthentication: RECOGNISED_AUTH_METHODS,
+    };
+  },
+  store: (request) => {
```
chrmod: all those methods should probably be async

proposals/secure-storage.md
```diff
@@ -100,6 +100,8 @@ browser.secureStorage.store({
 });
 ```
 
+The authentication array is optional. If omitted, the secret is available without biometrics but is still stored in the hardware backed location.
```
bradcush: Instead of "without biometrics" I think it would be more accurate to say without "a recognized auth method"

dotproto (commented):
For the sake of this polyfill, I think we should make two main changes.

1) Use `browser.storage.local` or `indexedDB` instead of `localStorage` to make the polyfill usable in (extension service) worker environments 
2) Encrypt data before putting it in storage and decrypt it after pulling it out

For 2, I believe we can use SubtleCrypto to handle encryption. For the moment I'm thinking this would be a relatively simple password based encryption scheme. The password could be held in a variable or `storeage.session` where available. After quickly poking around, it looks like [this article](https://bradyjoslin.com/blog/encryption-webcrypto/) describes how (or at least a way) to implement this with SubtleCrypto.

proposals/secure-storage-polyfill.js
oliverdunk: Good call! I assumed without checking that these would be read-only, but I was totally wrong. I've updated this to attach to both of those where available.

proposals/secure-storage-polyfill.js
oliverdunk: Definitely a good call, I've added the async keyword for each of these.

proposals/secure-storage.md
oliverdunk: I've tweaked the wording here a little, let me know what you think. I also opened an issue just now that may impact how this works going forward: https://github.com/w3c/webextensions/issues/190

oliverdunk:
> For the sake of this polyfill, I think we should make two main changes.
> 
> 1. Use `browser.storage.local` or `indexedDB` instead of `localStorage` to make the polyfill usable in (extension service) worker environments
> 2. Encrypt data before putting it in storage and decrypt it after pulling it out
> 
> For 2, I believe we can use SubtleCrypto to handle encryption. For the moment I'm thinking this would be a relatively simple password based encryption scheme. The password could be held in a variable or `storeage.session` where available. After quickly poking around, it looks like [this article](https://bradyjoslin.com/blog/encryption-webcrypto/) describes how (or at least a way) to implement this with SubtleCrypto.

Thanks for taking a look @dotproto!

I definitely agree on supporting extension service workers, so I've moved to `browser.storage.local`. Let me know what you think.

On encrypting data, I think I can add this, but I'm curious what the benefit is. Do you think it just makes things a bit more real and usable? Of course if we include the password in the polyfill it doesn't really achieve anything and we're not doing anything that isn't already known to be possible.

dotproto:
> Do you think it just makes things a bit more real and usable?

That was the idea.

As I understand it a polyfill (or a [prollyfill](https://twitter.com/slexaxton/status/257543702124306432) in this case) is a lib that provides the features of a spec/standard that isn't implemented in a browser. As such, I assumed that this library was aiming to both provide a functional API surface developers could experiment with as well as implement as much of the spec's functionality as reasonably possible.

If you want to instead focus on just the API surface without worrying much about functionality, I'd suggest we tweak the framing of this library to say it's a demo, fake, double, or perhaps something else entirely. (Or even [poppyfill](https://twitter.com/dotproto/status/1509684481334022161)?)

oliverdunk:
@dotproto: I'm going to bring this up at WECG tomorrow but I think I'd rather take the path of changing the framing than add a crypto implementation. Having _some_ sort of implementation is useful to see it working, but trying to implement something that is cryptographically sound (even if we're using Web APIs which do most of that for us) seems like an easy way to get distracted, and also gives the impression that this is real more than I would like to without more scrutiny. Poppyfill all the way!

oliverdunk:
We discussed this at the last meeting (https://github.com/w3c/webextensions/pull/215) and agreed that we should merge this as-is after renaming it from a "polyfill" to a "mock", to indicate that there is no cryptographic backing. I've gone ahead and done that so this should be good for another look :)

xeenon (approved):

dotproto (approved):
Sorry for the delay on this

Issue #119, opened 2021-10-25
Proposal: add `optional_host_permissions`


carlosjeurissen:
In mv3, the idea is to only have `host_permissions` and make them optional by default. This prevents vendors from deciding whether or not some `host_permissions` should be requested upon installation or only later in the process. And also prevents extension authors from telling the stores, browsers and users what `host_permissions` are needed for the core functionality and which ones are `optional_host_permissions`.

The proposal here is to add a new property in addition to `host_permissions` named `optional_host_permissions` allowing extension authors to define which host permissions are optional and which are mandatory for the core functionality. This would then allow some vendors to request the required host permissions upon installation, and allow other vendors to also treat the required `host_permissions` as being optional.

For example, Apple currently requests host permissions just after installation from the user. Yet the developer may not want to ask all host permissions right away upon installation.

One of the reasons for this proposal is the fact not all extensions act on websites / tabs, but need host permissions for functionality provided separately in an extension popup completely independent from browser tabs.

An issue in the chromium issue tracker can be found here:
https://bugs.chromium.org/p/chromium/issues/detail?id=1265064

### Status
Apple Safari: in favor, not yet implemented.
Mozilla Firefox: in favor, not yet implemented.
Google Chrome: intend to support this, will work on it soon.

carlosjeurissen:
Based on meeting 2021-10-28 (from meeting notes):
_Apple (timothy): in favor
Google (simeon): missing optional_host_permission was too optimistic about timelines of rolling out MV3 and permissions UI. Personally in favor of optional, but can’t commit on behalf of the team. Expect that._

Issue has also been added to chromiums issue tracker:
https://bugs.chromium.org/p/chromium/issues/detail?id=1265064

Jack-Works:
When I asked for missing `optional_host_permission` one year ago, the response team from the Chrome team means that all `host_permissions` are optional so there's no need for an "optional" one

carlosjeurissen:
> When I asked for missing `optional_host_permission` one year ago, the response team from the Chrome team means that all `host_permissions` are optional so there's no need for an "optional" one

Can totally imagine that. As Simeon mentioned in meeting 2021-10-28, they expected all of them to be optional. Yet not every change in Chrome has arrived as soon as they expected. So now they are more open to it. Even if in the future, some browsers consider "host_permissions" as being optional, it can still be a great addition for the reasons stated in my original post.

dotproto:
[...]

PR #221, opened 2022-05-31
Add host_permissions and optional_host_permissions

carlosjeurissen:
As each vendor agreed to add `optional_host_permissions` and thus also `host_permissions` in manifest v3, it's time to add it to this repo.

Background issue about supporting `optional_host_permissions`: https://github.com/w3c/webextensions/issues/119

***
<a href="https://pr-preview.s3.amazonaws.com/carlosjeurissen/webextensions/pull/221.html" title="Last updated on Jun 1, 2022, 7:02 AM UTC (8c0d835)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webextensions/221/953709d...carlosjeurissen:8c0d835.html" title="Last updated on Jun 1, 2022, 7:02 AM UTC (8c0d835)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -63,6 +63,10 @@ The following keys must be considered valid:
 * <a href="#key-devtools_page">`devtools_page`</a>: optional
 * <a href="#key-externally_connectable">`externally_connectable`</a>: optional
 
+The following keys must be considered valid in manifest v3:
+* <a href="#key-host_permissions">`host_permissions`</a>: optional
+* <a href="#key-optional_host_permissions">`optional_host_permissions`</a>: optional
+
 ### Key `manifest_version`
 
 This key must be present.
@@ -76,6 +80,22 @@ This key must be present. This property can be localized.
 ### Key `version`
 
 This key must be present.
+
+### Key `permissions`
+
+This key may be present.
+
+### Key `optional_permissions`
+
+This key may be present.
+
+### Key `host_permissions`
+
+This key may be present.
+
+### Key `optional_host_permissions`
+
+This key may be present.
 
 ### Key `default_locale`
 
@@ -105,15 +125,7 @@ This key may be present.
 
 This key may be present.
 
-### Key `optional_permissions`
-
-This key may be present.
-
 ### Key `options_ui`
-
-This key may be present.
-
-### Key `permissions`
 
 This key may be present.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -63,6 +63,10 @@ The following keys must be considered valid:
 * <a href="#key-devtools_page">`devtools_page`</a>: optional
 * <a href="#key-externally_connectable">`externally_connectable`</a>: optional
 
+The following keys must be considered valid in manifest v3:
```
dotproto:
```suggestion
The following keys must be considered valid in Manifest V3:

```

* Insert a newline between the list and paragraph for consistency.
* "Manifest V3" is a proper noun, so it should be uppercased. The canonical casing Chrome has been using is to capitalize both the M and V. I'd prefer to default to that pattern unless and until we adopt a different pattern in this group.

index.bs
```diff
@@ -77,6 +81,22 @@ This key must be present. This property can be localized.
...
+This key may be present.
+
+### Key `host_permissions`
+
+This key may be present.
+
+### Key `optional_host_permissions`
+
+This key may be present.
+
```
dotproto, line 100: Nit: We may want to consider sorting the list of keys alphabetically or by another well-defined scheme.

dotproto (changes requested):
I suggested a small change, but beyond that this LGTM.

index.bs
carlosjeurissen: The list was already not sorted alphabetically. Grouped the permissions-related keys together. Logical grouping seems more useful than alphabetically sorting the list.

carlosjeurissen: Changed the casing of the Manifest V3 mention.

dotproto: I prefaced this with "nit" because it was a small, relatively unimportant observation. I was mostly just thinking out loud. I mostly agree with the intent behind grouping related keys.

That said,  but for readers that aren't clear on the association between proximity-based groups I worry that it can end up feeling like a random collection rather than an intentionally curated list. One way we could address this is by making groups part of the heading structure of the document, but I don't feel strongly enough about it to tackle that concern in this PR or likely even this year.

dotproto (approved):
LGTM. Thanks for taking this on, @carlosjeurissen.

xeenon (approved):

dotproto (approved):
Sorry for the delay!

PR #222, opened 2022-06-02
Move list of meetings into a collapsible section

dotproto:
This PR aims to make the list of previous meetings on `_minutes/README.md` less overwhelming by breaking flat list of all previous meetings into groups based on the year they occurred and collapsing these groups by default. This keeps the readily available while minimizing the amount of information readers are confronted with when they visit the README.

### DIFF
--- a/_minutes/README.md
+++ b/_minutes/README.md
@@ -21,6 +21,18 @@ After the end of each meeting, meeting notes are published here.
 * 2022-06-09 ([minutes](2022-06-09-wecg.md))
 * 2022-05-26 ([minutes](2022-05-26-wecg.md))
 * 2022-05-12 ([minutes](2022-05-12-wecg.md))
+
+<details>
+<summary><strong>All past meeting notes</strong></summary>
+
+**2022**
+
+* 2022-07-21 ([minutes](2022-07-21-wecg.md))
+* 2022-07-07 ([minutes](2022-07-07-wecg.md))
+* 2022-06-23 ([minutes](2022-06-23-wecg.md))
+* 2022-06-09 ([minutes](2022-06-09-wecg.md))
+* 2022-05-26 ([minutes](2022-05-26-wecg.md))
+* 2022-05-12 ([minutes](2022-05-12-wecg.md))
 * 2022-04-28 ([minutes](2022-04-28-wecg.md))
 * 2022-04-14 ([minutes](2022-04-14-wecg.md))
 * 2022-03-31 ([minutes](2022-03-31-wecg.md))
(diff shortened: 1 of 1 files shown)


xeenon (approved):

Rob--W (commented):
The meeting notes are already ordered, with the most recent one being at the top for visibility.

If the issue is mainly the long list, could we just add `</details>` at the end, and `<details><summary>Older meeting notes</summary>` after e.g. the 5 most recent meetings? Then we can easily "archive" older meeting notes without impacting the `git blame` history (i.e. it is easy to find the PR to which the commit was associated).

Rob--W (changes requested):
This PR has merge conflicts. But if the previous feedback is addressed (after a rebase), it could be merged.

mukul-p (approved):

dotproto:
@Rob--W I think is in a better state. I combined the collapsible sections as you suggested. I also kept the per-year sub-lists as I find that they help group the items visually and I think they also make it a little easier for screen reader users to navigate the content.

Rob--W:
The diff looks as if you've added new meeting notes, but I'm pretty sure that you have just rebased. Could you squash all changes and rebase it to the current main branch?

```sh
git checkout collapsible-meetings

git fetch origin main

git rebase -i origin/main
# ^ Replace the last two "pick" with "squash".

git commit --am --reset-author -m 'Move list of meetings into a collapsible section'
```

_minutes/README.md
```diff
@@ -10,12 +10,22 @@ After the end of each meeting, meeting notes are published here.
...
 ## Past meetings
 
+* 2022-07-07 ([minutes](2022-07-07-wecg.md))
+* 2022-06-23 ([minutes](2022-06-23-wecg.md))
+* 2022-06-09 ([minutes](2022-06-09-wecg.md))
+* 2022-05-26 ([minutes](2022-05-26-wecg.md))
 * 2022-05-12 ([minutes](2022-05-12-wecg.md))
+
+<details>
+<summary>Older meeting notes</summary><br>
```
Rob--W: While the current PR preserves the `git blame` history of the meetings that we have had so far, it will force changes on new ones, when the "oldest recent" meeting note gets moved to the section below.

To preserve the `git blame` history, could we rename "Older meeting notes" to something like "All (past) meeting notes", which has a full list of all meeting notes (the current list)? We would basically end up with the list before this PR, plus a volatile copy of the full list above the "archive" (i.e. the list of 5 meeting notes).

dotproto: Ah, I hadn't considered preserving blame. I'm happy to run with this suggestion; I'm tentatively planning to use "All past meeting notes".

Rob--W (approved):

PR #215, opened 2022-05-26
Publish minutes of 2022-05-26 meeting

Rob--W:
Generated from https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit using the tool and process from #105.

During this meeting we discussed or mentioned #208, #214, #212, #162, #190, PR #186. We had 20 minutes left at the end of the meeting, where we discussed more topics, notably on the work towards making host permissions opt-in across browsers. Mozilla has also announced a Developer Preview of Firefox's MV3 implementation at https://blog.mozilla.org/addons/2022/05/18/manifest-v3-in-firefox-recap-next-steps/.

PR #226, opened 2022-06-08
Update browser.secureStorage proposal with authentication levels

### LINKED ISSUE #190: browser.secureStorage: Discuss how to handle platforms with limited control over authentication methods
Currently, the [browser.secureStorage](https://github.com/w3c/webextensions/blob/main/proposals/secure-storage.md) proposal relies on an authentication array which the developer can use to require specific types of authentication (for example, only face or fingerprint based biometrics).

However, I recently found out that with some implementations (like Windows Hello), it isn't currently possible to specify this and you must request authentication without any control over if biometrics or another method like a PIN code is used.

This is already the case for desktop apps using Windows Hello, so isn't a huge concern, but we should find a way of making it clear to developers that we can't guarantee what auth method secrets are protected by. My first thought is some sort of "*" authentication method that you are forced to pass in on Windows but I'm open to better suggestions.

cc/ @mukul-p - I think this is a Windows API limitation but maybe you have thoughts/there is a way of doing this.

oliverdunk:
Following discussion at the last meeting (https://github.com/w3c/webextensions/pull/215) and in #190, this PR updates the browser.secureStorage proposal to improve two aspects of the API:

- Clearer, more generic authentication levels are now used which provide a higher level API. This allows the developer to request that a secret is protected by `BIOMETRICS` for example, without making things too specific and allowing them to specify a particular piece of hardware to use.
- There is now documentation on how to handle implementations like Windows Hello where the OS does not provide the option for authentication methods to be specified. This is through the introduction of a new `ANY` authentication level.

The `implementation` key has also been added which the browser can use to share more data with the developer about the OS-specific hardware it might use. This is not standardised and is purely informational.

Resolves #190

### DIFF
--- a/proposals/secure-storage.md
+++ b/proposals/secure-storage.md
@@ -38,34 +38,81 @@ Response:
 ```
 {
   type: "MACOS_KEYCHAIN",
-  availableAuthentication: [
-    "PIN",
-    "PASSWORD",
-    "BIOMETRY_FACE",
-    "BIOMETRY_FINGERPRINT"
+  supportedAuthenticationLevels: [
+    {
+      name: "BIOMETRIC",
+      implementation: [
+        "MACOS_KEYCHAIN_FACE",
+        "MACOS_KEYCHAIN_TOUCHID"
+      ]
+    },
+    {
+      name: "DEVICE_CREDENTIAL",
+      implementation: [
+        "MACOS_ACCOUNT_PASSWORD"
+      ]
+    }
   ]
 }
 ```
 
+Name can be either `BIOMETRIC` or `DEVICE_CREDENTIAL`, indicating which authentication levels are supported for a `browser.secureStorage.store` call. These are intentionally vague descriptors that allow the API to be used with some level of abstraction away from the hardware.
+
+Implementation is an array of browser-specific descriptors that a developer can use to learn more about what specific implementations the browser will use. These are not standardised and therefore are not used elsewhere in the API. They simply exist so the developer can get a better idea about what will be used if they request a secret to be stored with a particular authentication level.
+
+On platforms like Windows, where the OS provides an API like Windows Hello but does not allow specific methods to be requested, the response will look as follows:
+
+```
+{
+  type: "WINDOWS_HELLO",
+  supportedAuthenticationLevels: [
+    {
+      name: "ANY",
+      implementation: []
+    }
+  ]
+}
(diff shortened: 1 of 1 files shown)


dotproto (approved):
Since this proposal is authored and edited by @oliverdunk, I'm inclined to accept whatever changes he makes in order to reduce the friction of iterating on this proposal. Discussions of specific concerns with the current draft can take place in appropriately tagged issues. What do you think @xeenon, @zombie, @mukul-p, and @Rob--W?

oliverdunk:
> Since this proposal is authored and edited by @oliverdunk, I'm inclined to accept whatever changes he makes in order to reduce the friction of iterating on this proposal. Discussions of specific concerns with the current draft can take place in appropriately tagged issues. What do you think @xeenon, @zombie, @mukul-p, and @Rob--W?

I'm of course in favor of merging things, appreciate you trying to keep things moving! That said - don't feel a rush to merge this. There's no point merging things with no intent of implementing them so I see getting feedback here as really important.

proposals/secure-storage.md
```diff
@@ -38,35 +38,82 @@ Response:
...
-  availableAuthentication: [
-    "PIN",
-    "PASSWORD",
-    "BIOMETRY_FACE",
-    "BIOMETRY_FINGERPRINT"
+  supportedAuthenticationLevels: [
+    {
+      name: "BIOMETRIC",
+      implementation: [
+        "MACOS_KEYCHAIN_FACE",
```
xeenon, line 45: `MACOS_KEYCHAIN_FACEID`

mukul-p (approved):

proposals/secure-storage.md
```diff
@@ -38,35 +38,82 @@ Response:
...
+        "MACOS_KEYCHAIN_TOUCHID"
```
xeenon, line 46: This would apply to iOS as well. I would prefer to not bake the platform into the contents like this. I much preferred the previous generic wording.

oliverdunk:
I synced with Simeon and got some feedback here. I'm going to make a new draft based on that soon.

dotproto:
@oliverdunk would you like to merge these changes as is or create a new PR for with the updates you're working on?

oliverdunk:
@dotproto: Let's close this for now and I'll re-open once I have a new draft. Thanks for checking in!

Issue #267, opened 2022-08-26
Proposal: declarative way for opening an extension page in a new tab when click the extension icon


hanguokai:
Popup is a temporary small window. Sometimes, developers want another behavior: open a page in a new tab when users click the extension icon.

At present, developers must declare a background script(service worker), and add one line of code in it:
`browser.action.onClicked.addListener(_ => chrome.tabs.create({url: "index.html"}));`

My proposal likes below:
```
  // manifest.json
  "action": {
    "default_tab_page": "index.html",
  }
```
```
  // action api
  browser.action.getTabPage();
  browser.action.setTabPage('index.html');
```
Unlike popup page, tab page has no per-tab state. When click the extension icon, just open that page in a new tab.

fregante:
This is already trivial to implement, but it's also a very simple request. I would not mind such interface.

Rob--W:
Discussed during the meeting; Since the feature can already trivially be implemented with a background worker/script, it is not necessary to add the functionality to the extension API. Extension API additions are better reserved for features that are difficult to achieve otherwise.

carlosjeurissen:
In the current form, it indeed seems quite trivial to implement something like this.

However, when exploring further options / features, this could start becoming useful. Think about the browser being smart enough to focus the `full tab page` if it is already opened, instead of opening a new one. And/or have this `full tab page` be a pinned tab right away when opening it.

Rob--W:
> In the current form, it indeed seems quite trivial to implement something like this.
> 
> However, when exploring further options / features, this could start becoming useful. Think about the browser being smart enough to focus the `full tab page` if it is already opened, instead of opening a new one. And/or have this `full tab page` be a pinned tab right away when opening it.

If that is the desired result, then we could repurpose the existing default_popup and setPopup APIs along with a new property `open_in_tab` (similar to `options_ui`). Clicking the popup and responding to a click are mutually exclusive anyway.

But unlike options pages and (browser) action panels, opening a new tab is already easily possible with existing extension APIs, which was the reason for closing this feature request.

The participants in the meeting, including representatives of Apple (Timothy), Google (Simeon) and Mozilla (me), were all not against the functionality itself, but after weighing implementation/maintenance effort and balancing that against the benefits to developers, we decided to not add a new API requested here.

Issue #264, opened 2022-08-24
Supply credentials for proxy authorization in MV3


erba2061:
Issue https://bugs.chromium.org/p/chromium/issues/detail?id=1135492

### Suggestions

- Declarative `authCredentials` property in [ProxyServer](https://developer.chrome.com/docs/extensions/reference/proxy/#type-ProxyServer) config 
  - Firefox supports similar implementation with `proxyAuthorizationHeader` in [ProxyInfo](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/proxy/ProxyInfo)
- Allow to modify/add `Proxy-Authorization` header on CONNECT requests
  - https://bugs.chromium.org/p/chromium/issues/detail?id=1352397
  - https://bugs.chromium.org/p/chromium/issues/detail?id=129572

### Background

Blocking webRequest API will not be supported in favour of DNR API. Adding blocking `webRequest.onAuthRequired` listener is the only way to supply credentials for proxy servers in Chromium browsers.

Note that browser sign-in dialog box (asks user to enter proxy credentials) is not sufficient:
  - User does not know which credentials he/she should enter. Proxy server authorization credentials are usually different from account credentials used to login to use extension features
  - Proxy returns 407 (Proxy Authentication Required) more than once during browsing time, causing dialogs to popup at random times

### Impact

Privacy extensions won't be able to provide proxy services to users without consistent authorization solution in MV3.

olso:
Have the same issue

erosman:
There could be an alternative approach to proxy authentication in MV3 without `webRequest.onAuthRequired`, by using DNR.

Currently DNR can not be applied to proxy. If it would be possible to apply to proxy then, a set of rules can be created using `declarativeNetRequest.updateDynamicRules` to pass authentication to the proxy.

Scenario: `Browser -> proxyurl.com -> target.com`
It appears that the `urlFilter` applies to the `target.com`.

```js
const removeRuleIds = [1];
const addRules = [
  {
    id: 1,
    priority: 1,
    action: {
      type: 'modifyHeaders',
      responseHeaders: [
        {
          header: 'Proxy-Authorization',
          operation: 'set',
          value: 'Basic ' + btoa(username + ':' + password)
        }
      ]
    },
    condition: {urlFilter: '||proxyurl.com'}
  }
];

chrome.declarativeNetRequest.updateDynamicRules({removeRuleIds, addRules});
```

[...]

Issue #263, opened 2022-08-24
undefined value in StorageArea.set()


regseb:
Hi,

I noticed a difference between Firefox and Chromium on the [`StorageArea.set()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set) method.

- In Firefox, `StorageArea.set()` removes item when value is `undefined`.
- In Chromium, `StorageArea.set()` ignore item when value is `undefined`.

## Expected

I propose a third solution where the value is saved in the storage.

_This pseudo-code use [assert.deepStrictEqual()](https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message) from Node.js._

```JavaScript
const data = { foo: undefined };
chrome.storage.local.set(data);
chrome.storage.local.get(Object.keys(data), (d) => assert.deepStrictEqual(d, data));
```

## Testcase

- _manifest.json_:

```JSON
{
    "manifest_version": 2,
    "name": "Test case",
    "version": "1.0.0",
    "background": {
        "scripts": ["background.js"]
    },
    "permissions": ["storage"]
}
```

- _background.js_:

```JavaScript
chrome.storage.local.clear(() => {
    chrome.storage.local.set({ foo: "bar" });
    chrome.storage.local.get("foo", (v) => console.log(v));
    chrome.storage.local.set({ foo: undefined });
    chrome.storage.local.get("foo", (v) => console.log(v));
    chrome.storage.local.set({ foo: "baz" });
    chrome.storage.local.get("foo", (v) => console.log(v));
});
```

### Firefox

> { foo: "bar" }
> {  }
> { foo: "baz" }

### Chromium

> { foo: "bar" }
> { foo: "bar" }
> { foo: "baz" }

Rob--W:
Relevant bug reports:

- Bug 1499842 - "chrome.storage.sync.set({'existingKey': undefined}): inconsistency with Chromium"
  https://bugzilla.mozilla.org/show_bug.cgi?id=1499842
- Issue 896503: chrome.storage.sync.set({'existingKey': undefined}): inconsistency with Firefox
  https://bugs.chromium.org/p/chromium/issues/detail?id=896503

xeenon:
Safari matches Chrome here (the key is ignored and will not be set or changed). A `null` value can be used in Safari for doing this.

Rob--W:
[...]

Issue #262, opened 2022-08-24
Inconsistency: `oldValue` in `storage.StorageChange`


regseb:
Hi,

I noticed a difference between Firefox and Chromium on the `storage.onChanged` event.

- In Firefox, [`storage.StorageChange`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageChange) always has an `oldValue` property (with `undefined` when there was not old value).
- In Chromium, `storage.StorageChange` has not an `oldValue` property when there was not old value.

## Testcase

- _manifest.json_:

```JSON
{
    "manifest_version": 2,
    "name": "Test case",
    "version": "1.0.0",
    "background": {
        "scripts": ["background.js"]
    },
    "permissions": ["storage"]
}
```

- _background.js_:

```JavaScript
chrome.storage.onChanged.addListener((changes) => {
    console.log(changes.foo);
});
chrome.storage.local.clear(() => {
    chrome.storage.local.set({ foo: "bar" });
    chrome.storage.local.set({ foo: "baz" });
});
```

### Firefox

> { oldValue: undefined, newValue: "bar" }
> { oldValue: "bar", newValue: "baz" }

### Chromium

> { newValue: "bar" }
> { newValue: "baz", oldValue: "bar" }

xeenon:
Safari matches Chrome here.

xeenon:
We resolved in the meeting to close this as it fundamentally does not matter much, fetching `change.oldValue` will result in `undefined` either way. The only impact is key iteration and we decided that was not important enough to warrant a change in browsers here.

carlosjeurissen:
@regseb if this actually causes practical issues when you want to distinguish a new installation from a storage key set to an empty value you can consider using `null` instead of `undefined`. `null` is handled in all browsers when setting storage.

PR #268, opened 2022-09-01
Publish minutes of 2022-09-01 meeting

Rob--W:

During this meeting we discussed or mentioned #260, #261, #262, #263, #264, #267, #244, #232, #242, #162.

I've also fixed some timezone references (PST -> PDT). Unfortunately, the everytimezone links were also off-by-one for many months, but they should be good now.

### DIFF
--- a/_minutes/2022-09-01-wecg.md
+++ b/_minutes/2022-09-01-wecg.md
@@ -1,0 +1,155 @@
+# WECG Meetings 2022, Public Notes, Sep 1
+
+ * Chair: Timothy Hatcher
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=630ff600,384
+Call-in details: [WebExtensions CG, 1st September 2022](https://www.w3.org/events/meetings/d7bbce8f-549f-46ea-b440-ea6902f8707c/20220901T080000)
+Zoom issues? Ping `@robwu` (Rob Wu) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+
+## Agenda: [github issues](https://github.com/w3c/webextensions/issues)
+
+The meeting will start at 3 minutes after the hour.
+
+ * **Carry-over from previous meetings**
+   * None
+ * **Other new issues**
+   * [todd] Optional permissions affordances for enterprise security/compliance use cases
+     * [Issue 260](https://github.com/w3c/webextensions/issues/260): "Runtime allowed hosts" on Enterprise to pre-grant optional host permissions without user action
+     * [Issue 261](https://github.com/w3c/webextensions/issues/261): API to retrieve user email or well-known identifier on Enterprise deployments
+   * [Issue 262](https://github.com/w3c/webextensions/issues/262): Inconsistency: oldValue in storage.StorageChange inconsistency
+   * [Issue 263](https://github.com/w3c/webextensions/issues/263): undefined value in StorageArea.set()
+   * [Issue 264](https://github.com/w3c/webextensions/issues/264): Supply credentials for proxy authorization in MV3
+   * [Issue 267](https://github.com/w3c/webextensions/issues/267): Proposal: declarative way for opening an extension page in a new tab when click the extension icon
+ * **Open discussion queue (add yourself at the bottom)**
+   * None
+ * **Check-in on ongoing issues**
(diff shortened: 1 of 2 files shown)


hanguokai:
Was #258 discussed at this meeting?

I don't see #252 discussed at this meeting. Why @carlosjeurissen added label `accepted: chrome` and `accepted: firefox` on it?

I see chrome and safari don't opposed to #267 . Why @Rob--W added label `rejected: chrome` and `rejected: safari` on it?

carlosjeurissen:
@hanguokai Thanks for closely following the meeting notes and processes in the group.

#252 was discussed in the previous meeting (Aug 4, 2022), in which Chrome and Safari expressed being in favour.

As for #267, neutral tags might also work here. @xeenon proposed to be able to tag issues with a neutral `no-objections` tag. See: https://github.com/w3c/webextensions/issues/244#issuecomment-1234680010

Rob--W:
> Was #258 discussed at this meeting?

It wasn't part of the agenda, so it wasn't covered. I don't know if omission from the agenda by @xeenon was intentional.

> I don't see #252 discussed at this meeting. Why @carlosjeurissen added label `accepted: chrome` and `accepted: firefox` on it?

This was dicussed during the previous meeting. Notes from that meeting are at
https://github.com/w3c/webextensions/blob/main/_minutes/2022-08-04-wecg.md

The labels were recently added because they are new, see #244 for details on labels. 

> I see chrome and safari don't opposed to #267 . Why @Rob--W added label `rejected: chrome` and `rejected: safari` on it?

The "Resolution" comment in the meeting notes was added after we all reached consensus on closing the issue.

hanguokai:
> The "Resolution" comment in the meeting notes was added after we all reached consensus on closing the issue.

Closing it is OK. But I'm talking about labels here, you could close it without "reject" labels.

> This was dicussed during the previous meeting. Notes from that meeting are at
> https://github.com/w3c/webextensions/blob/main/_minutes/2022-08-04-wecg.md
> #252 was discussed in the previous meeting (Aug 4, 2022), in which Chrome and Safari expressed being in favour.

In 2022-08-04-wecg.md, about #252 , 

> "[simeon] From my perspective, a Chrome issue would be welcome."

Like millions of other Chromium issues, I think this sentence means this is a beginning and does not mean acceptance.

In today meeting note, about #267 , 

> "[simeon] In general in favor of making developers lifes easier. No strong objections, but in terms of relative priority this feels like something as an Available issue."

I think this likes above sentence, that means you can add an issue too, but will not implement soon.

> As for #267, neutral tags might also work here. @xeenon proposed to be able to tag issues with a neutral no-objections tag. See: #244 (comment)

This is better than "accept" or "reject". I understand, from browser venders point of view, there are still a lot of important issues that remain unresolved. Therefore, it is often difficult to say whether it will be implemented or when it will be implemented.

Rob--W:
> > This was dicussed during the previous meeting. Notes from that meeting are at
> > https://github.com/w3c/webextensions/blob/main/_minutes/2022-08-04-wecg.md
> > #252 was discussed in the previous meeting (Aug 4, 2022), in which Chrome and Safari expressed being in favour.
> 
> In 2022-08-04-wecg.md, about #252 ,
> 
> > "[simeon] From my perspective, a Chrome issue would be welcome."
> 
> Like millions of other Chromium issues, I think this sentence means this is a beginning and does not mean acceptance.

We've just renamed "accepted" to "supportive" to more closely match what we're intending.

> > The "Resolution" comment in the meeting notes was added after we all reached consensus on closing the issue.
> 
> Closing it is OK. But I'm talking about labels here, you could close it without "reject" labels.

In a similar vein, we've renamed "rejected" to the less strongly worded "opposed".

> In today meeting note, about #267 ,
> 
> > "[simeon] In general in favor of making developers lifes easier. No strong objections, but in terms of relative priority this feels like something as an Available issue."
> 
> I think this likes above sentence, that means you can add an issue too, but will not implement soon.

The meeting notes reflect what has been said in the meeting. Because it is a discussion, viewpoints can change after having expressed and processed the initial statements. In this specific example, we were initially indifferent/neutral, but after weighing the options we decided to oppose the feature request (elaborated in #267).

dotproto:
Most of the discussion on this issue seems to be about how labels have been applied to issues rather than the minutes themselves. As such, I think it may be best to (1) move discussion of the labels themselves to issue #244 and (2) move discussion of a whether given label was correctly applied to that issue's comments.

xeenon (approved):

Issue #328, opened 2022-11-21
Inconsistency: How should tabs.onUpdated behave with the History API?


oliverdunk:
Try the following:

1. Open https://example.com.
2. In an extension, add a `browser.tabs.onUpdated` listener.
3. In the console of the web page, run `history.pushState({}, "", "/test")`.

Currently:

- Chrome fires both a "loading" event and a "completed" one in immediate succession.
- Firefox fires a single "completed" event.
- Safari fires no events.

I think an event makes sense because the tab has updated. However, it would be useful to identify this as a slightly different type of event, since extensions may otherwise accidentally inject multiple times on the same page if they are unaware of this edge case.

Related Chromium bug: https://bugs.chromium.org/p/chromium/issues/detail?id=106429

xeenon:
Filed https://bugs.webkit.org/show_bug.cgi?id=248177 for Safari/WebKit.

xeenon:
I think Firefox's single "completed" event makes the most sense, because no loading is involved in `history.pushState()` changes.

Sxderp:
I'm not sure if that's sufficient. As the OP mentioned, doing that may cause the extension to inject / perform work multiple times for a page. Maybe that's on the developer to implement a tracking mechanism? Tracking becomes a little harder due to ServiceWorker implementation and the inability to rely on cache variables (still haven't seen vendor solutions for this).

xeenon:
`browser.storage.session` can be used by a service worker for cache variables.

M-SAI-SOORYA:
You are correct that different browsers handle the history.pushState event differently, which can cause inconsistencies in how extensions behave when listening for browser.tabs.onUpdated events.

To address this issue, you can check the changeInfo.url property of the browser.tabs.onUpdated event to determine if the page URL has changed. This property will be undefined for events that do not involve a URL change, such as when a page's title changes or when a page finishes loading.

By checking the changeInfo.url property, you can ensure that your extension only injects code once per URL change, regardless of how the browser handles the history.pushState event.

Issue #327, opened 2022-11-16
Update WECG meeting schedule through January 2023


dotproto:
**Overview**

Over the next few weeks there are a number of holidays observed across the globe that impact the availability of attendees. We need to go through the tentative meeting schedule and make adjustments to account for these issues. 

**Additional information**

#314 covers a similar issue from 2021.

dotproto:
Assuming our regular bi-weekly caydence, I've gathered a list of meeting dates and noted potential conflicts. 

| Date | Notes | 
| --- | --- |
| 2022-11-24 | Martyrdom of Guru Tegh Bahadur Sahib<br>Thanksgiving (USA) | 
| 2022-12-08 | _None_ |
| 2022-12-22 | Proximity to winter holiday period & common vacation times |
| 2023-01-05 | Proximity to winter holiday period & common vacation times |
| 2023-01-19 | _None_ |

Based on this, I am planning cancel the 2022-11-24 meeting. Since 2022-12-22 does not have a direct conflict, I plan to sync with the other organizers before our next public session and (tentatively) add an agenda item to our next public session to discuss.

dotproto:
I had a chance to check in with the editors and chairs to discuss upcoming meeting schedules. After some discussion, we decided to proceed with the following plan:

- The 2022-11-24 meeting will be held as scheduled. @zombie will chair this meeting.
- The 2022-12-22 meeting will be canceled due to vacations during the winter holidays.
- All of the other meetings will remain on the calendar as originally planned.

**Action items**

- [x] Cancel the 2022-12-22 session in the official WECG calendar. ([event](https://www.w3.org/events/meetings/d7bbce8f-549f-46ea-b440-ea6902f8707c/20221222T080000))

dotproto:
Test

dotproto:
Test

Issue #326, opened 2022-11-16
Any plans to implement "tab" context menu?


NV:
"tab" context menu is only implemented in Firefox. Are Chrome and Safari supportive of this API? If not, are there any issues with it?

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/ContextType

I'm working on an [extension that auto-archives old tabs](https://tab-cellar.app/). It also has an option to manually archive (close and save the state) and unload (discard) a tab. These actions should really be context menu items for a tab, but currently, I have to use "action"/"page_action" instead.

dotproto:
Chrome doesn't have any active plans here. Actually, to the best of my knowledge no one has formally requested this capability before now. I just opened https://crbug.com/1385529 to track this feature request.

rdcronin:
As @dotproto mentioned, this is tracked in Chromium (new URL: https://issues.chromium.org/issues/40246822).  I'm personally supportive of this for Chrome.  We're not actively working on it, but may in the future.

xeenon:
[WebKit has support](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/Shared/Extensions/WebExtensionMenuItemContextType.h) for this now, and will be in an future Safari release.

oliverdunk:
@xeenon I tried this in Safari 26.3.1 and I wasn't able to see my item in the tab strip. Is it possible there is some additional UI work needed to link up the API?

oliverdunk:
Chrome is also investigating this in https://crrev.com/c/7689933. We still need to go through some additional API design and security reviews before landing the change but hopefully we can ship support soon.

carlosjeurissen:
@oliverdunk Just tested in Safari 26.4.0, with https://jeurissen.co/webext-demos/context-menus-menu-types-product, and the `tab` context menu does show up when right clicking tabs. This seems to work in both the normal tab layout and the compact tab layout.

Glad to hear Chrome is in the process of adding this feature!

oliverdunk:
Thanks Carlos! Hmm, I updated and now everything is working. @xeenon Did this get added recently, or do you think I went wrong in my test somewhere?

xeenon:
> Thanks Carlos! Hmm, I updated and now everything is working. [@xeenon](https://github.com/xeenon) Did this get added recently, or do you think I went wrong in my test somewhere?

It is only supported on Mac currently. It was added in Safari 15.4, since the WebKit web extension APIs.

[...]

Issue #325, opened 2022-11-15
Popup should not close on file navigation initiated by the popup


AidanWelch:
Presently, if trying to either saveAs or input a file to a browser extension you have to create a separate tab, and it cannot be done in popup.  Clicking within the file navigation initiated by the popup should not close the popup, allowing the popup script to actually be used for file inputting and outputting.

This does slightly expose a risk of a malicious popup spamming file navigation popups, but that doesn't justify removing the functionality entirely.  This change would improve both user and developer experience.

AidanWelch:
On further testing, Chrome actually does behave this way, but Firefox does not.

hanguokai:
Hi @AidanWelch 

It looks like a Chrome bug. There was a [discussion](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/aZH3sUbLU9g/m/ctAddRPaAAAJ) and a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1158230) for it. But it seems to have been fixed. I can't reproduce this problem in a simple example. If you can, please [file a bug](https://crbug.com/new) and close this issue (because browser-specific bugs are not discussed here).

AidanWelch:
> Hi @AidanWelch
> 
> It looks like a Chrome bug. There was a [discussion](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/aZH3sUbLU9g/m/ctAddRPaAAAJ) and a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1158230) for it. But it seems to have been fixed. I can't reproduce this problem in a simple example. If you can, please [file a bug](https://crbug.com/new) and close this issue (because browser-specific bugs are not discussed here).

The issue doesn't occur on Chrome, and as far as I know it isn't an intended feature of Firefox as it doesn't happen in Firefox.  I think it is instead an asynchrony in the standard.  (Chrome behaves how I described, Firefox doesn't)

hanguokai:
> Chrome behaves how I described, Firefox doesn't

Do you means it works on Chrome but doesn't work on Firefox? Then it is a Firefox bug. Normally, using `<input type=file>` or download action in the popup should not cause the popup to be closed.

AidanWelch:
> Do you means it works on Chrome but doesn't work on Firefox? 

Yeah

> Then it is a Firefox bug. Normally, using `<input type=file>` or download action in the popup should not cause the popup to be closed.

[...]

PR #329, opened 2022-11-24
Publish minutes of 2022-11-24 meeting

Rob--W:

During this meeting we discussed or mentioned #320, #311, #317, #318, #319, #325, #326, #328, #97, #274, #327, #279, #258. I also summarized the result of the user script meeting (whose notes were published at #324). Many topics were nominated but postponed to the next meeting because the representatives from Google and Apple were absent due to a US holiday (Thanksgiving).


dotproto:
If all else fails I can review & approve, but since I didn't attend I'd be more comfortable if @zombie or @mukul-p could.

AidanWelch:
The discussion mentions #235 not #325, just to correct a minor typo

PR #333, opened 2022-12-08
Add IDL files for Safari as reference.

xeenon:
This adds the `interfaces/safari` directory and the IDL files used for the JavaScript interfaces for Web Extensions. These can be used as a reference for specification work and to identify inconsistencies.


dotproto (approved):
Thanks, @xeenon!

Issue #294, opened 2022-10-16
Proposal: one-time message passing addressed to a specific documentId or contextId


bershanskiy:
## Summary

Currently, there are two types of messaging: one-time messages and persistent ports. Persistent ports have some efficiency implications while one-time messages work only in contexts clearly associated with a tab and messages sent from background to content scripts can accidentally be delivered to a wrong content script if there is a race between tab update and message sending.

### Ergonomics: not all contexts have associated tab

One-time messaging is routed based on `tabId` and `frameId`, which, of course, exist only for tab contexts. These attributes do not exist on messages sent from [Sidebars](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Sidebars) (tested on Firefox and Opera), Extension Actions, and Extension options. This necessitates some trickery for initiating sending messages from background to these contexts (e.g., background context can open a short-lived persistent port, saving message to storage for recipient to get `onChanged` event or similar place, etc.) Therefore, background _can_ initiate messaging, but it is very not ergonomic.

### Confused recipient problem: `tabId` and `frameId` are not globally unique, frames can load different documents

Update: Chromium supports this via `documentId` option on `tabs.sendMessage()`.

Currently, some browsers use simple counters for `tabId` and even `frameId` (e.g., Firefox) while others started to generate somewhat unique numbers (e.g., Chromium). However, all browsers have to set main frame id to `0` following prior convention. This means that background can accidentally mean to send message to a particular context (tab's top frame or a subframe) but this context can be replaced with another one (page navigation or reload) and message would be delivered to the new context.

Also, a minor nit: in the current framework it is conceptually impossible to communicate with contexts which are being unloaded or prerendered. I'm not sure how useful that would be in practice, but previously I had cases where I need to clean up data in `storage.StorageArea` associated with a particular tab/frame when said frame is destroyed and I had to choose between developer ergonomics and efficiency (e.g., using non-passive `unload` handler is easy and produces maintainable code, but breaks BFCache). Conceptually, it would be nice if it was possible to navigate away from a frame but still be able to communicate with document for a short while (e.g., 10 seconds) to finalize some stuff.

## Solution
Browsers already assign unique identifiers to document contexts: Chromium calls these `documentId` while Firefox calls them `contextId`. It would be nice to be able to route messages based on these ids. I treat `documentId` and `contextId` as interchangeable identifiers, since most contexts are documents anyway (except for background service workers which don't really need these ids).

## API proposal
[...]

Issue #293, opened 2022-10-15
Transfer huge binary data buffer between service worker and content script (more efficiently)


tophf:
Currently it's impossible to transfer 1GB ArrayBuffer/Blob/Uint8Array and the like between an extension background script (service worker) and the extension's content script both instantly (~1ms for Blob, 1sec for buffers) and directly.

**Current workarounds**

1. The fastest workaround is to create a web_accessible_resources iframe -> then use navigator.serviceWorker messaging and [secure](https://github.com/w3c/webextensions/issues/77#issuecomment-1243308962) message passing with the parent window, all the while using the last parameter of postMessage to enable instant transfers. This workaround is fragile because the web page can delete the iframe, even if we hide it inside shadow DOM. It also adds a considerable overhead to create and initialize the iframe, which is a waste of time (at least 50ms), CPU, and memory.

2. Sending binary data via extension sendMessage API via structured clone algorithm. Chrome can't do it yet, Firefox can. This is quite slow and blocks both processes (extension and the web page) for the duration of the internal serialization/deserialization, which can be long for a huge binary data, thus introducing janks and lags.

**Better solutions wanted**

1. URL.createObjectURL + asynchronous fetch in the content script doesn't block the page. Currently it is disabled in service workers due to [concerns with the lifetime of the blob](https://github.com/w3c/ServiceWorker/issues/688). The web doesn't suffer from this restriction because a web SW is only used by the same origin pages/workers which all can use postMessage or self.onfetch + Response API. The extension's SW usage patterns have nothing in common with the web SW in 99.9% of cases, so we need this restriction lifted - just for the duration of SW lifetime.

2. Add `transfer` parameter to chrome.tabs.sendMessage or at least to long-lived messaging via port.postMessage, that transfers ownership of the data instantly same as in the web postMessage's [transferables](https://developer.mozilla.org/en-US/docs/Glossary/Transferable_objects).

Rob--W:
We've discussed this issue, not the solutions as phrased, but more the general capability request.

The overall sentiment is that the browser vendors are in favor of the better-designed capabilities from the web platform. I'm tentatively marking this as `follow-up: follow-up`, so that @dotproto can synchronize with his team to confirm whether his personal opinion is also shared by the team.

[...]

Issue #134, opened 2021-11-10
Proposal: Limited Event Pages for MV3


zombie:
As part of our work on MV3, Mozilla is working on service workers for extensions in Firefox, both for broad compatibility reasons, and because they’re an event-driven environment with defined lifetimes, and already part of the Web Platform with good cross-browser support.

Chrome’s current implementation of MV3 introduces _Background Service Worker_ as a replacement for the (persistent) background page. However, we’ve found that this approach can’t fully support [various use cases](https://github.com/w3c/webextensions/issues/72) we consider important, especially around DOM-related features and APIs.

Additionally, the worker environment is unfamiliar to regular web developers, and we’ve heard repeated complaints from our community that completely rewriting extensions would be an unreasonable burden for the thousands of solo developers and small shops whose  [existing extensions](https://addons.mozilla.org/en-US/firefox/extensions/) currently work just fine.

Overall we like many aspects of MV3, and would prefer a speedy transition, so we don’t want our answer to that capability gap to be extending MV2 support indefinitely.  Instead, we’re proposing:

## Limited Event Pages

These would be similar to existing non-persistent or “event” pages from MV2, with a few additional limitations to align with MV3.  They’re a familiar concept for developers and broadly compatible across browsers.  They would help with both of our main goals: enabling extensions which can be [suspended and resumed as needed](https://github.com/w3c/webextensions/issues/51#issuecomment-892702868), and providing functionality related to the DOM.

One limitation would be that _inactive_ [Ports](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port) would no longer prevent the Event page from suspending, something Safari for iOS is already doing.  Since we’re still in the design phase, we might discover additional limitations as we progress with prototyping.

Otherwise, the same policies and lifetime rules that hold for service workers should be applied to event pages.  It seems that there’s a hard 5 minute cutoff in Chromium, but there are no such platform limitations in Gecko, so we’re unlikely to be applying arbitrary time limits, at least on desktop.

## MV3 and the future

This is a good opportunity for compromise, with a promising cost-benefit ratio for the short and medium term.

It would afford us time to seek solutions for the missing functionality through the WECG (as part of “MV4”), in the best way possible: either pursuing them through the Web Platform, or by designing extension-specific APIs that work for both browser vendors and extension developers.

And as a result, by the time we’re ready for MV4, the whole ecosystem will probably be in a better position to transition fully to service workers.

[...]

Issue #91, opened 2021-09-30
Proposal: `scripting.executeScript` without specifying `tabId`


carlosjeurissen:
### Background
1) currently `scripting.executeScript` / `tabs.executeScript` require one to define the tabId of the main_frame or sub_frame to inject the content script in.
2) extension popups have their tabId set to -1 (no tabId). This makes sense as they are no tabs, but puts limitations like the inability to use executeScript for iframes within extension popups.

### Proposal
Allow to only specify a `frameId` as target when running `scripting.executeScript` / `tabs.executeScript`. This basically means either `tabId` or `frameId` will be required when running `executeScript` or `insertCSS`.

### Reasoning
This will allow dynamic contentscript to be injected into frames within non-tab environments like extension popups, and to allow injection of script when it's faster / more convenient to not specify a tabId.

### Requirement
All frameIds should be unique for this to be implemented. See:
- https://github.com/w3c/webextensions/issues/90

carlosjeurissen:
As discussed during the 2021-11-11 meeting, this can be tricky as of right now frameId is not unique in all contexts. A potential alternative proposal was to use a window object.

You can think about a syntax like this:
```
browserOrChrome.scripting.executeScript(
    {
      target: {window: someFrame.contentWindow},
      files: ['script.js'],
    },
    () => { ... });
```

carlosjeurissen:
As with https://github.com/w3c/webextensions/issues/12#issuecomment-942826300

With the introduction of documentId this seems to become possible.

erosman:
See also: 
[Proposal: Default to Active Tab in scripting.executeScript()](https://github.com/w3c/webextensions/issues/431)

rdcronin:
We discussed this at the WECG meet-up.

We're agreed that this is useful functionality.  There's one tricky bit, which is that currently, the main frame is always `0` (whereas subframes are globally-unique).  While we've talked about changing that, it's a very breaking change and not something we're actively looking at.

We tentatively agred that we're supportive of:
- Allowing non-zero frameId without a tabId (new)
- Allowing any documentId without a tabId (new)
- Allowing any frameId _with_ a tabId (as today)

This would also allow extensions to inject in non-tab contexts, as @carlosjeurissen  mentioned in the issue report.

We also pointed out that this does introduce a _requirement_ that non-main-frame IDs are globally unique.  This is already an attribute in most Chromium-based browsers, Firefox, and Safari, but is now a requirement for this functionality.

[...]

PR #334, opened 2022-12-08
New API proposal : extension.getContexts()

justinlulejian:
@dotproto

### DIFF
--- a/proposals/extension_get_contexts.md
+++ b/proposals/extension_get_contexts.md
@@ -137,7 +137,7 @@ methods, one synchronous, and one asynchronous API methods:
     thread). If called from a service worker, it would likely only return the
     service worker `Context`.
 
-### Content Script Contexts
+### Content Script Contexts <a name="content-scripts-fw">Content Script Contexts</a>
 
 [Content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/)
 run in a separate
@@ -148,6 +148,16 @@ querying all Renderer processes in the browser to find content scripts. However,
 extensions could find it useful to have this `Context` metadata for places where
 the extension is executing javascript in a web page. It could be justified in
 the future if extension developers would have a significant need.
+
+### Messaging APIs Support `Context`s as Targets
+
+This new API now provides the ability to consider specifying targets via
+`Context`s to messaging APIs like `runtime.sendMessage()` or
+`tabs.sendMessage()`. With this support it would avoid
+[various workarounds](https://github.com/w3c/webextensions/pull/334#issuecomment-1343255899)
+needed to target specific pages with messages. `tabs.sendMessage()` accepting
+`Context`s would require work in retrieving content script `Context`s explained
+[above](#content-scripts-fw).
 
 ## <a name="footnotes">Footnotes</a>
 
--- a/proposals/extension_get_contexts.md
+++ b/proposals/extension_get_contexts.md
@@ -1,0 +1,161 @@
+# New API: extension.getContexts()
+
+## Background / Summary
+
+Chromium currently has the
+<code>[extension.getViews()](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+API method which allows an extension to get information about the views the
+extension is running in. Common reasons for calling this method are: determining
+if a toolbar popup is open, reading and/or modifying a settings page, and etc.
+Chromium's implementation of Manifest V2 (MV2) allows for
+[background scripts (pages)](https://developer.chrome.com/docs/extensions/mv2/background_pages/)
+to call this API. This is possible because these scripts run on our main thread
+which allows them easy access to the Javascript
+[Window objects](http://go/mdn/API/Window#instance_properties) provided by
+<code>extension.getViews()</code>.
+
+## Problem
+
+However, with Chromium’s migration to Manifest V3 (MV3), background pages
+[no longer exist](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/);
+instead, an extension’s background context is
+[service worker](https://developer.chrome.com/docs/workbox/service-worker-overview/)-based.
+For technical reasons<sup>[1](#footnotes)</sup>, service workers cannot access
+the [HTMLWindow](http://go/mdn/API/Window#instance_properties) objects that
+`extension.getViews()` provides and it is not feasible to implement that with
+our browser design. Due to this we cannot provide access to the JS context for
+these views, but we can allow an extension to query for them and potentially
+target them for messaging purposes.
+
+## Solution
+
+Considering the above situation, we’d like to propose a new (Manifest V3 only)
+method to asynchronously provide metadata about the view that is still useful
+for an extension: `extension.getContexts()`. This will allow extension service
+workers (or other contexts) to identify the active contexts in the extension.
+For example, this can be used to target messages to send using
+<code>[runtime.sendMessage()](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage)</code>,
+etc.). Introducing this API will allow an easier migration from MV2.
+
+## API Proposal
+
+This new API would use a similar parameter signature as
+`extension.getViews()`but instead accept a new `contextProperties` object.
+
+```
+extension.getContexts()(
+  contextProperties?: object,
+)
+```
+
+`contextProperties` will have similar behavior to `extension.getViews()`’s
+<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+object in that it will limit the scope of the query to certain tabs, context
+types, and windows. It would be defined as:
+
+```
+extension.ContextProperties = {
+  tabId? = int,                // Find context by tab id, omitted returns all tabs.
+  contextType? = ContextType,  // Find context by type, omitted returns all types.
+  windowId? = int,             // Find context by window id, omitted returns all
+                               // windows.
+}
+```
+
+`ContextType` would serve a similar purpose of
+<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>’s
+[ViewType](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
+in order to limit the type of extension contexts returned by the API. It would
+be defined as:
+
+```
+extension.ContextType = {
+  TAB: 'TAB',                               // Tabs the extension is running in.
+  POPUP: 'POPUP',                           // Popups created by the extension.
+  SERVICE_WORKER: 'SERVICE_WORKER',         // Service workers for the extension.
+  OFFSCREEN_DOCUMENT: 'OFFSCREEN_DOCUMENT'  // Offscreen documents for the extension.
+};
+```
+
+It would return an array of new `Context` objects for each frame the extension
+is running in:
+
+```
+extension.Context = {
+  tabId: int,                    // ID of the tab for this view.
+  windowId: int,                 // ID of the window this tab is located in.
+  documentId: int,               // ID of the DOM document loaded in the tab.
+  frameId: int,                  // ID of the frame this context is running for.
+  contextType: ContextType,      // Context type -- tab, popup, etc.
+  url: string,                   // The current URL of the tab.
+  origin: string,                // The origin of the context. Mostly this is the
+                                 // extension origin.
+  incognito: boolean             // Whether the context is for an incognito
+                                 // profile.
+}
+```
+
+`Context` objects will contain an `incognito` value indicating if the Context is
+running in an incognito profile. In this area, this brings up the case of an
+extension specifying
+[split incognito mode](https://developer.chrome.com/docs/extensions/mv3/manifest/incognito/#split).
+In this case we would not return the <code>Context</code>s from the related (but
+different) extension process running for the incognito profile.
+<code>spanning</code> mode would return any <code>Context</code>s running in an
+incognito profile (with <code>Context.incognito</code> being <code>true</code>)
+in the same shared process. More comments about this are in the Future Work
+section on this proposal.
+
+## Future Work
+
+### Multi Page Architecture Fields (`DocumentLifeCycle`, `FrameType`, and `parentDocumentId`, and `parentFrameId`)
+
+[Multi Page Architecture](http://doc/1NginQ8k0w3znuwTiJ5qjYmBKgZDekvEPC22q0I4swxQ#heading=h.w1qo2n6sr8wn)
+changed how outermost frames are presented in web navigation in chromium. More
+context on how this can impact extensions is
+[here](https://developer.chrome.com/blog/extension-instantnav/). Considering
+this new architecture it is likely that extensions will want to know what
+navigation state (`DocumentLifeCycle`) a `Context` is in, whether the frame’s
+`FrameType` is an outermost frame, and other fields like `parentDocumentId`, and
+`parentFrameId`. These are fields we would like to include in a future iteration
+of this API.
+
+### Asynchronous and Synchronous Version
+
+A potential avenue is that, since this new API is asynchronous whereas
+`getViews()` is synchronous, we could explore a synchronous version of this new
+API if there’s significant demand for it. As mentioned
+previously<sup>[1](#footnotes)</sup>, there are technical limitations that make
+it infeasible to synchronously fetch all `Context`s. We could provide two
+methods, one synchronous, and one asynchronous API methods:
+
+*   `getContexts()`: as defined above and asynchronous. It fetches all
+    `Context`s available in the process running for the profile.
+*   `getContextsSynchronous()` (name TBD): defined similarly to `getContexts()`,
+    but is synchronous. It returns the data for other `Context`s that are
+    available synchronously (running in the same process and on the same
+    thread). If called from a service worker, it would likely only return the
+    service worker `Context`.
+
+### Content Script Contexts
+
+[Content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/)
+run in a separate
+[Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (and
+process) separate from the extension process. This proposal would not return
+those `Context`s since this adds significant complexity to the design in
+querying all Renderer processes in the browser to find content scripts. However,
+extensions could find it useful to have this `Context` metadata for places where
+the extension is executing javascript in a web page. It could be justified in
+the future if extension developers would have a significant need.
+
+## <a name="footnotes">Footnotes</a>
+
+<sup>1</sup>: Non-main threads in a
+[Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (where
+service workers run in Chromium) cannot access DOM concepts directly (they are
+only accessible from the main browser thread). Service workers run in these
+threads therefore they do not have access to the Javascript
+[HTMLWindow](http://go/mdn/API/Window#instance_properties) objects provided by
+`extension.getViews()`. Supporting this access is likely years of software
+engineer work to implement.
--- a/proposals/extension_get_contexts.md
+++ b/proposals/extension_get_contexts.md
@@ -3,16 +3,16 @@
 ## Background / Summary
 
 Chromium currently has the
-<code>[extension.getViews()](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+[`extension.getViews()`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)
 API method which allows an extension to get information about the views the
 extension is running in. Common reasons for calling this method are: determining
 if a toolbar popup is open, reading and/or modifying a settings page, and etc.
 Chromium's implementation of Manifest V2 (MV2) allows for
 [background scripts (pages)](https://developer.chrome.com/docs/extensions/mv2/background_pages/)
 to call this API. This is possible because these scripts run on our main thread
-which allows them easy access to the Javascript
-[Window objects](http://go/mdn/API/Window#instance_properties) provided by
-<code>extension.getViews()</code>.
+which allows them easy access to the JavaScript
+[`Window`](http://go/mdn/API/Window#instance_properties) objects provided by
+`extension.getViews()`.
 
 ## Problem
 
@@ -21,20 +21,19 @@ However, with Chromium’s migration to Manifest V3 (MV3), background pages
 instead, an extension’s background context is
 [service worker](https://developer.chrome.com/docs/workbox/service-worker-overview/)-based.
 For technical reasons<sup>[1](#footnotes)</sup>, service workers cannot access
-the [HTMLWindow](http://go/mdn/API/Window#instance_properties) objects that
+the [`Window`](http://go/mdn/API/Window#instance_properties) objects that
 `extension.getViews()` provides and it is not feasible to implement that with
-our browser design. Due to this we cannot provide access to the JS context for
-these views, but we can allow an extension to query for them and potentially
-target them for messaging purposes.
+our browser design. Due to this we cannot provide access to the JavaScript
+context for these views, but we can allow an extension to query for them and
+potentially target them for messaging purposes.
 
 ## Solution
 
-Considering the above situation, we’d like to propose a new (Manifest V3 only)
+Considering the above situation, we’d like to propose a new `extension` API
 method to asynchronously provide metadata about the view that is still useful
-for an extension: `extension.getContexts()`. This will allow extension service
-workers (or other contexts) to identify the active contexts in the extension.
+for an extension: `extension.getContexts()`. This will allow extension background scripts to identify the active contexts in the extension.
 For example, this can be used to target messages to send using
-<code>[runtime.sendMessage()](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage)</code>,
+[`runtime.sendMessage()`](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage),
 etc.). Introducing this API will allow an easier migration from MV2.
 
 ## API Proposal
@@ -49,7 +48,7 @@ extension.getContexts()(
 ```
 
 `contextProperties` will have similar behavior to `extension.getViews()`’s
-<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+[`fetchProperties`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)
 object in that it will limit the scope of the query to certain tabs, context
 types, and windows. It would be defined as:
 
@@ -63,8 +62,8 @@ extension.ContextProperties = {
 ```
 
 `ContextType` would serve a similar purpose of
-<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>’s
-[ViewType](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
+[`fetchProperties`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)’s
+[`ViewType`](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
 in order to limit the type of extension contexts returned by the API. It would
 be defined as:
 
@@ -72,7 +71,9 @@ be defined as:
 extension.ContextType = {
   TAB: 'TAB',                               // Tabs the extension is running in.
   POPUP: 'POPUP',                           // Popups created by the extension.
-  SERVICE_WORKER: 'SERVICE_WORKER',         // Service workers for the extension.
+  BACKGROUND: 'BACKGROUND',                 // Background script for the extension.
+                                            // (background pages for MV2, or service
+                                            // workers for MV2 + MV3).
   OFFSCREEN_DOCUMENT: 'OFFSCREEN_DOCUMENT'  // Offscreen documents for the extension.
 };
 ```
@@ -87,7 +88,8 @@ extension.Context = {
   documentId: int,               // ID of the DOM document loaded in the tab.
   frameId: int,                  // ID of the frame this context is running for.
   contextType: ContextType,      // Context type -- tab, popup, etc.
-  url: string,                   // The current URL of the tab.
+  frameUrl: string,              // The current URL of the frame. For background
+                                 // pages and service workers this value is empty.
   origin: string,                // The origin of the context. Mostly this is the
                                  // extension origin.
   incognito: boolean             // Whether the context is for an incognito
@@ -99,55 +101,62 @@ extension.Context = {
 running in an incognito profile. In this area, this brings up the case of an
 extension specifying
 [split incognito mode](https://developer.chrome.com/docs/extensions/mv3/manifest/incognito/#split).
-In this case we would not return the <code>Context</code>s from the related (but
-different) extension process running for the incognito profile.
-<code>spanning</code> mode would return any <code>Context</code>s running in an
-incognito profile (with <code>Context.incognito</code> being <code>true</code>)
-in the same shared process. More comments about this are in the Future Work
-section on this proposal.
+In this case we would not return the `Context`s from the related (but different)
+extension process running for the incognito profile. `spanning`mode would return
+any `Context`s running in an incognito profile (with `Context.incognito` being
+`true`) in the same shared process. More comments about this are in the Future
+Work](#future-work) section on this proposal.
 
 ## Future Work
 
 ### Multi Page Architecture Fields (`DocumentLifeCycle`, `FrameType`, and `parentDocumentId`, and `parentFrameId`)
 
-[Multi Page Architecture](http://doc/1NginQ8k0w3znuwTiJ5qjYmBKgZDekvEPC22q0I4swxQ#heading=h.w1qo2n6sr8wn)
-changed how outermost frames are presented in web navigation in chromium. More
-context on how this can impact extensions is
-[here](https://developer.chrome.com/blog/extension-instantnav/). Considering
-this new architecture it is likely that extensions will want to know what
-navigation state (`DocumentLifeCycle`) a `Context` is in, whether the frame’s
-`FrameType` is an outermost frame, and other fields like `parentDocumentId`, and
-`parentFrameId`. These are fields we would like to include in a future iteration
-of this API.
+[Multi Page Architecture](https://docs.google.com//1NginQ8k0w3znuwTiJ5qjYmBKgZDekvEPC22q0I4swxQ#heading=h.w1qo2n6sr8wn)
+caused
+[multiple changes](https://developer.chrome.com/blog/extension-instantnav/) to
+Chromium's
+[tabs API](https://developer.chrome.com/docs/extensions/reference/tabs/),
+[scripting API](https://developer.chrome.com/docs/extensions/reference/scripting/),
+and
+[web navigation API](https://developer.chrome.com/docs/extensions/reference/webNavigation/).
+For the web navigation API it changed how outermost frames are exposed to
+extensions in Chromium. Considering this new architecture it is likely that
+extensions will want to know what navigation state (`DocumentLifeCycle`) a
+`Context` is in, whether the frame’s `FrameType` is an outermost frame, and
+other fields like `parentDocumentId`, and `parentFrameId`. These are fields we
+would like to include in a future iteration of this API.
 
 ### Asynchronous and Synchronous Version
 
 A potential avenue is that, since this new API is asynchronous whereas
 `getViews()` is synchronous, we could explore a synchronous version of this new
-API if there’s significant demand for it. As mentioned
-previously<sup>[1](#footnotes)</sup>, there are technical limitations that make
-it infeasible to synchronously fetch all `Context`s. We could provide two
-methods, one synchronous, and one asynchronous API methods:
+API if there’s significant demand for it. That demand would need to justify why
+the usage of `await` on the existing asynchronous API isn't sufficient. As
+mentioned previously<sup>[1](#footnotes)</sup>, there are technical limitations
+that make it infeasible to synchronously fetch all `Context`s. We could provide
+two methods, one synchronous, and one asynchronous API methods:
 
 *   `getContexts()`: as defined above and asynchronous. It fetches all
     `Context`s available in the process running for the profile.
 *   `getContextsSynchronous()` (name TBD): defined similarly to `getContexts()`,
     but is synchronous. It returns the data for other `Context`s that are
     available synchronously (running in the same process and on the same
-    thread). If called from a service worker, it would likely only return the
-    service worker `Context`.
+    thread). If called from a background script, it would likely only return the
+    background script `Context`.
 
-### Content Script Contexts <a name="content-scripts-fw">Content Script Contexts</a>
+### Content Script Contexts
 
 [Content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/)
 run in a separate
 [Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (and
-process) separate from the extension process. This proposal would not return
-those `Context`s since this adds significant complexity to the design in
-querying all Renderer processes in the browser to find content scripts. However,
-extensions could find it useful to have this `Context` metadata for places where
-the extension is executing javascript in a web page. It could be justified in
-the future if extension developers would have a significant need.
+process) from the extension process. This proposal would not return those
+`Context`s since this adds significant complexity to the design in querying all
+Renderer processes in the browser to find content scripts. However, extensions
+could find it useful to have this `Context` metadata for places where the
+extension is executing JavaScript in a web page. It could be justified in the
+future if extension developers would have a significant need. If implemented the
+`frameUrl` field would be empty as providing a URL to the content script (within
+the extension) doesn't have a use case.
 
 ### Messaging APIs Support `Context`s as Targets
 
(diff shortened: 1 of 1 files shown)


hanguokai:
> this can be used to target messages to send using runtime.sendMessage()

Yes, this api is very useful in this scenario. Because `sendMessage()` send a message to all contexts but only get one(the first) response. In some use cases, I want to find a target first (usually a tab), then only communicate with it.

Currently, I use some workaround ways. For example, 
```
// This page can be in tab mode or popup mode
chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
  if(request.type == "SomeType") {
    chrome.tabs.getCurrent(tab => {
      if(tab) { // in tab mode
        sendResponse('ok');
      } else { // in popup mode

      }
    });
  }
});
```
Another example, find an extension tab from foreground by `getViews` and check `location.pathname`.
```
let tabWindows = chrome.extension.getViews({type: "tab"});
let win = tabWindows.find(w => w.location.pathname == '/a_specific_page.html');
if(win) {
  win.someMethod(); // call some method on it
} else {

}
```

A typical use case is that implement single tab mode: only open a page in one tab, if it is already open then active it (not open multiple same pages).

Another problem of current API is that `chrome.tabs.query({url: 'a current extension page url'})` doesn't work without the "tabs" permission.

hanguokai:
So I suggest `chrome.runtime.sendMessage()` supports specifying a context as target to send.

justinlulejian:
> So I suggest `chrome.runtime.sendMessage()` supports specifying a context as target to send.

@hanguokai Hi Jackie! Agreed that this could be a useful place to allow `Context`s to be specified and something that should be done. I've committed a section to future work since it seems beyond scope for the initial implementation. PTAL and let me know if the wording accurately reflects your suggestion.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+For technical reasons<sup>[1](#footnotes)</sup>, service workers cannot access
+the [HTMLWindow](http://go/mdn/API/Window#instance_properties) objects that
+`extension.getViews()` provides and it is not feasible to implement that with
+our browser design. Due to this we cannot provide access to the JS context for
+these views, but we can allow an extension to query for them and potentially
+target them for messaging purposes.
+
+## Solution
+
+Considering the above situation, we’d like to propose a new (Manifest V3 only)
```
xeenon: Why is it v3 only? v2 supports service workers in Chrome and Safari.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>’s
+[ViewType](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
+in order to limit the type of extension contexts returned by the API. It would
+be defined as:
+
+```
+extension.ContextType = {
+  TAB: 'TAB',                               // Tabs the extension is running in.
+  POPUP: 'POPUP',                           // Popups created by the extension.
+  SERVICE_WORKER: 'SERVICE_WORKER',         // Service workers for the extension.
```
xeenon: I would call this `BACKGROUND` to allow Safari and Firefox to support this for background pages in v3. As mentioned above this should also not be limited to v3, and a more generic name is warranted, since v2 supports service workers or pages.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+is running in:
+
+```
+extension.Context = {
+  tabId: int,                    // ID of the tab for this view.
+  windowId: int,                 // ID of the window this tab is located in.
+  documentId: int,               // ID of the DOM document loaded in the tab.
+  frameId: int,                  // ID of the frame this context is running for.
+  contextType: ContextType,      // Context type -- tab, popup, etc.
+  url: string,                   // The current URL of the tab.
```
xeenon: Is this the tab URL or the frame's URL? Will is return or non-tab contexts, like popup or background?

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+  origin: string,                // The origin of the context. Mostly this is the
```
xeenon, line 93: What do you mean by origin here? Host name? Security origin?

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+                                 // extension origin.
+  incognito: boolean             // Whether the context is for an incognito
```
xeenon, line 95: Can we get away from browser marketing names in the API? `ephemeral` is a good neutral term for this. Documentation can tie it to the browser's marketing name.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+of this API.
+
+### Asynchronous and Synchronous Version
+
+A potential avenue is that, since this new API is asynchronous whereas
+`getViews()` is synchronous, we could explore a synchronous version of this new
+API if there’s significant demand for it. As mentioned
+previously<sup>[1](#footnotes)</sup>, there are technical limitations that make
+it infeasible to synchronously fetch all `Context`s. We could provide two
+methods, one synchronous, and one asynchronous API methods:
```
xeenon: `await` is good enough for this. Let's not add any more complex sync APIs.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
+# New API: extension.getContexts()
+
+## Background / Summary
+
+Chromium currently has the
+<code>[extension.getViews()](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
```
dotproto:
```suggestion
[`extension.getViews()`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+API method which allows an extension to get information about the views the
+extension is running in. Common reasons for calling this method are: determining
+if a toolbar popup is open, reading and/or modifying a settings page, and etc.
+Chromium's implementation of Manifest V2 (MV2) allows for
+[background scripts (pages)](https://developer.chrome.com/docs/extensions/mv2/background_pages/)
+to call this API. This is possible because these scripts run on our main thread
+which allows them easy access to the Javascript
+[Window objects](http://go/mdn/API/Window#instance_properties) provided by
```
dotproto:
```suggestion
[`Window`](http://go/mdn/API/Window#instance_properties) objects provided by
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+<code>extension.getViews()</code>.
```
dotproto:
```suggestion
`extension.getViews()`.
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+target them for messaging purposes.
+
+## Solution
+
+Considering the above situation, we’d like to propose a new (Manifest V3 only)
+method to asynchronously provide metadata about the view that is still useful
+for an extension: `extension.getContexts()`. This will allow extension service
+workers (or other contexts) to identify the active contexts in the extension.
+For example, this can be used to target messages to send using
+<code>[runtime.sendMessage()](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage)</code>,
```
dotproto:
```suggestion
[`runtime.sendMessage()`](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage),
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+`extension.getViews()`but instead accept a new `contextProperties` object.
+
+```
+extension.getContexts()(
+  contextProperties?: object,
+)
+```
+
+`contextProperties` will have similar behavior to `extension.getViews()`’s
+<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
```
dotproto:
```suggestion
[`fetchProperties`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+extension.ContextProperties = {
+  tabId? = int,                // Find context by tab id, omitted returns all tabs.
+  contextType? = ContextType,  // Find context by type, omitted returns all types.
+  windowId? = int,             // Find context by window id, omitted returns all
+                               // windows.
+}
+```
+
+`ContextType` would serve a similar purpose of
+<code>[fetchProperties](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>’s
```
dotproto:
```suggestion
[`fetchProperties`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)'s
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+[ViewType](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
```
dotproto:
```suggestion
[`ViewType`](https://developer.chrome.com/docs/extensions/reference/extension/#type-ViewType)
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+  incognito: boolean             // Whether the context is for an incognito
+                                 // profile.
+}
+```
+
+`Context` objects will contain an `incognito` value indicating if the Context is
+running in an incognito profile. In this area, this brings up the case of an
+extension specifying
+[split incognito mode](https://developer.chrome.com/docs/extensions/mv3/manifest/incognito/#split).
+In this case we would not return the <code>Context</code>s from the related (but
```
dotproto:
```suggestion
In this case we would not return the `Context`s from the related (but
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+different) extension process running for the incognito profile.
+<code>spanning</code> mode would return any <code>Context</code>s running in an
```
dotproto:
```suggestion
`spanning` mode would return any `Context`s running in an
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+incognito profile (with <code>Context.incognito</code> being <code>true</code>)
```
dotproto:
```suggestion
incognito profile (with `Context.incognito` being `true`)
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+in the same shared process. More comments about this are in the Future Work
```
dotproto: Link directly to that section.

```suggestion
in the same shared process. More comments about this are in the [Future Work](#future-work)
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+    available synchronously (running in the same process and on the same
+    thread). If called from a service worker, it would likely only return the
+    service worker `Context`.
+
+### Content Script Contexts <a name="content-scripts-fw">Content Script Contexts</a>
+
+[Content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/)
+run in a separate
+[Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (and
+process) separate from the extension process. This proposal would not return
```
dotproto: "a separate Renderer separate from" is redundant.

```suggestion
process) from the extension process. This proposal would not return
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+those `Context`s since this adds significant complexity to the design in
+querying all Renderer processes in the browser to find content scripts. However,
+extensions could find it useful to have this `Context` metadata for places where
+the extension is executing javascript in a web page. It could be justified in
```
dotproto:
```suggestion
the extension is executing JavaScript in a web page. It could be justified in
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+`Context`s would require work in retrieving content script `Context`s explained
+[above](#content-scripts-fw).
+
+## <a name="footnotes">Footnotes</a>
+
+<sup>1</sup>: Non-main threads in a
+[Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (where
+service workers run in Chromium) cannot access DOM concepts directly (they are
+only accessible from the main browser thread). Service workers run in these
+threads therefore they do not have access to the Javascript
```
dotproto:
```suggestion
threads therefore they do not have access to the JavaScript
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+
+Chromium currently has the
+<code>[extension.getViews()](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+API method which allows an extension to get information about the views the
+extension is running in. Common reasons for calling this method are: determining
+if a toolbar popup is open, reading and/or modifying a settings page, and etc.
+Chromium's implementation of Manifest V2 (MV2) allows for
+[background scripts (pages)](https://developer.chrome.com/docs/extensions/mv2/background_pages/)
+to call this API. This is possible because these scripts run on our main thread
+which allows them easy access to the Javascript
```
dotproto:
```suggestion
which allows them easy access to the JavaScript
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+
+*   `getContexts()`: as defined above and asynchronous. It fetches all
+    `Context`s available in the process running for the profile.
+*   `getContextsSynchronous()` (name TBD): defined similarly to `getContexts()`,
+    but is synchronous. It returns the data for other `Context`s that are
+    available synchronously (running in the same process and on the same
+    thread). If called from a service worker, it would likely only return the
+    service worker `Context`.
+
+### Content Script Contexts <a name="content-scripts-fw">Content Script Contexts</a>
```
dotproto: GitHub automatically generates an ID for section titles. In this case, the ID will be "content-script-contexts".

```suggestion
### Content Script Contexts
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Messaging APIs Support `Context`s as Targets
+
+This new API now provides the ability to consider specifying targets via
+`Context`s to messaging APIs like `runtime.sendMessage()` or
+`tabs.sendMessage()`. With this support it would avoid
+[various workarounds](https://github.com/w3c/webextensions/pull/334#issuecomment-1343255899)
+needed to target specific pages with messages. `tabs.sendMessage()` accepting
+`Context`s would require work in retrieving content script `Context`s explained
+[above](#content-scripts-fw).
```
dotproto: Update to match the auto-generated ID for this section's title.

```suggestion
[above](#content-script-contexts).
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+
+## <a name="footnotes">Footnotes</a>
```
dotproto: GitHub will automatically generate a `footnotes` anchor for this header.

```suggestion
## Footnotes
```

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+<code>extension.getViews()</code>.
+
+## Problem
+
+However, with Chromium’s migration to Manifest V3 (MV3), background pages
+[no longer exist](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/);
+instead, an extension’s background context is
+[service worker](https://developer.chrome.com/docs/workbox/service-worker-overview/)-based.
+For technical reasons<sup>[1](#footnotes)</sup>, service workers cannot access
+the [HTMLWindow](http://go/mdn/API/Window#instance_properties) objects that
```
dotproto:
```suggestion
the [`HTMLWindow`](http://go/mdn/API/Window#instance_properties) objects that
```

dotproto (commented):
This review contains a number of suggestions to

- normalize the use of HTML in GitHub Flavored Markdown 
- clean up minor grammar issues

More substantive issues will be addressed in separate comments.

proposals/extension_get_contexts.md
dotproto: I read this as meaning the concept as a [URL](https://url.spec.whatwg.org/#url-class) instance's [origin attribute](https://url.spec.whatwg.org/#dom-url-origin).

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
+# New API: extension.getContexts()
+
+## Background / Summary
+
+Chromium currently has the
+<code>[extension.getViews()](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)</code>
+API method which allows an extension to get information about the views the
+extension is running in. Common reasons for calling this method are: determining
+if a toolbar popup is open, reading and/or modifying a settings page, and etc.
```
dotproto, line 9: "Etc." is an abbreviation of "et cetera". This phrase comes from Latin where et means "and" and cetera means "the rest". As such, and should not appear before etc. or the expanded meaning would be "and and the rest".

```suggestion
if a toolbar popup is open, reading and/or modifying a settings page, etc.
```

proposals/extension_get_contexts.md
dotproto: I'm hesitant about using `ephemeral`. While it's technically accurate (private browsing sessions are ephemerial), that label doesn't convey enough about the capability it's referencing. What about `private`? I think that better represents how users think about – and how browser describe – the feature.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,171 @@
...
+incognito profile (with <code>Context.incognito</code> being <code>true</code>)
+in the same shared process. More comments about this are in the Future Work
+section on this proposal.
+
+## Future Work
+
+### Multi Page Architecture Fields (`DocumentLifeCycle`, `FrameType`, and `parentDocumentId`, and `parentFrameId`)
+
+[Multi Page Architecture](http://doc/1NginQ8k0w3znuwTiJ5qjYmBKgZDekvEPC22q0I4swxQ#heading=h.w1qo2n6sr8wn)
+changed how outermost frames are presented in web navigation in chromium. More
```
dotproto: Chromium is a proper noun. The impact of MPArch wasn't limited to the Web Navigation API; it also affected the Tabs API and may have ripple effects on other platform features (e.g. documentId support in the Scripting API).

```suggestion
changed how outermost frames are exposed to extensions in Chromium. More
```

Rob--W:
Here's my feedback, grouped by three themes:

- Use cases?
- Prior art
- Properties and TOCTOU

## Use cases?

What are the use cases depending on this proposal?
From the proposal I can extract 3 motivations:

1. Primarily, an attempt to fill the gap left behind by the absence of `extension.getViews` in workers,
2. A way to query the active views of an extension,
3. A way to target them for messaging purposes.

1 and 2 are vague, and it would be nice to include more background on why such functionality is sufficiently important to be a necessary part of the extension APIs, and not already achievable in other (not super hacky) ways.

3 is useful on its own, and the need for a way to uniquely target a specific destination has been requested before (#294). It is even a prerequisite to improve the performance of extension messaging (elaborated at https://github.com/w3c/webextensions/issues/293#issuecomment-1310203017).

## Prior art

The proposal here is a successor to [`extension.getViews()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getViews), to fill a gap in Service worker-based extensions.
On the web platform, there is already an API designed for that (i.e. [Client](https://developer.mozilla.org/en-US/docs/Web/API/Client)) to enable SW to communicate with web pages and other workers that share the same origin (the opposite direction is via [`navigator.serviceWorker`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/serviceWorker)). Despite the existence of these web platform APIs, I still see value in an extension-specific API to query the views for messaging purposes, for at least two reasons:
1. the API works for extensions based on service workers and event page-based (#134) and
2. allows for extension-specific restrictions/relaxations.
    For example, "incognito" (or private browsing) generally offers isolation between otherwise same-origin contexts. In extensions, this line is more blurry, and extensions may intentionally want to reach other contexts.

"Context" and "ContextType" are quite generic. There is even already a [`contextMenus.ContextType` type](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contextMenus/ContextType). If possible it would be nice to have a more specific name, such as `ExtensionContext` and `extensionContextType`.

## Properties and TOCTOU

The proposed API is async and resolves to an object whose properties are conceptually not immutable. Without clear use cases, this may be a recipe for [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bugs. So let's take a critical look at the fields to determine whether they are really necessary (which is dependent on the use cases).

| Property name | Immutable | Always well-defined | Alternative (in the target context) |
|-|-|-|-|
| tabId | yes | no, e.g. background contexts, potentially ambiguous for devtools/sidebar | extension messaging + [MessageSender](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/MessageSender) or `tabs.getCurrent()` |
| windowId | no, e.g. tab can be moved to different window. | no, e.g. background contexts | extension messaging + [MessageSender](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/MessageSender) or `windows.getCurrent()` |
| documentId | yes | no, e.g. workers | webNavigation API (see [Chrome blog post about Instant Navigation](https://developer.chrome.com/blog/extension-instantnav/)) |
| frameId | yes | no, e.g. workers | extension messaging + [MessageSender](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/MessageSender) or even (synchronous) `runtime.getFrameId()` ([MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getFrameId) |
| contextType | yes | no, e.g. embedded options page, devtools_page, etc. | Generally the (target) context knows where it is running, potentially by duck typing |
| url | no | no; "current URL of the tab." does not work for non-tab contexts. Mostly yes if "url" reflects the context's URL (mostly, because the URL can change at any point in time). | web platform APIs: `document.URL` or `location` |
| origin | yes, provided that the document did not unload, e.g. documentId did not change. | yes | web platform API: `origin` |
| incognito | no | yes | `extension.inIncognitoContext`|

proposals/extension_get_contexts.md
justinlulejian: I'll remove this. It was phrased that way since Chromium doesn't plan to support this for SW in MV2.

proposals/extension_get_contexts.md
justinlulejian: Corrected the typo, this would be the frame's URL so I renamed it to `frameUrl`.  In non-tab context like popup (Chromium as an example) popup  would be "chrome-extension://<id>/popup.html".

For a background page or service worker (since there's only one per extension) it doesn't seem helpful to know URL for it (e.g. "chrome-extension://<id>/background.js" so in those cases this would be empty. However, we could include it if there's a need for it? Applying this same logic would mean the field would be empty for content scripts as well, if supported.

proposals/extension_get_contexts.md
justinlulejian: `private` works for me, does that work for you too @xeenon?

proposals/extension_get_contexts.md
justinlulejian: I added a sentence that required the demand to justify why `await` cannot be used. If no others desire this -- I can replace the whole section before commiting mentioning that there is no significant need for this (just for future reference)?

justinlulejian:
@Rob--W Apologies I think by committing I prevented a directly reply to your comment, responding below

> Here's my feedback, grouped by three themes:
> 
> * Use cases?
> * Prior art
> * Properties and TOCTOU

Use cases: a possible example is to check that a popup is open. Otherwise the hacky way is to have `Context`s send keepalive messages back to the background script which maintains a table of what is open or not at any given time. A popup is a common example I can think of, but any other `Context` an extension wants to often know is open without having to manually keep track of it would benefit from this. Unless there's another way of keeping track of the idea of a `Context`, every extension would implement this differently and doing such state tracking sounds prone to bugs when the tracked state doesn't match what's actually running. If that seems like sufficient use case I can add a section to the proposal to have that for future reference.

For uniquely targeting a specific destination, this brings us closer to that goal. Currently there's just broadcast, but now an extension could check if a `Context` is running before attempting to send a message to it to be a bit more efficient.

I'll change `Context`/ `ContextType` to `ExtensionContext`/ `extensionContextType` -- thank you for the suggestion that seems much less overloaded.

TOCTOU:
Since the API is asynchronous there is the chance for these bugs to occur because of that, but it is something to keep in mind. We'd perhaps have to consider how to handle the case where, for example, the `windowId` would be used in another API call but has since changed and very confusing to extension developers when it happens. Is there prior art of figuring this out in other parts of the extension's ecosystem since these are not new properties?

Something I probably didn't elaborate on when describing these properties is that they might be empty/not defined for the  items where it doesn't make sense -- but altogether would cover the various extension contexts that might be returned. 

`tabId`: for non-tab contexts (e.g. background contexts, and etc.) this would be empty
`windowId`: same as above
`documentId`: empty for workers
`frameId`: same as above
`frameUrl`: empty for background contexts

`origin`: With the web platform API `origin` property, this wouldn't be that useful until/if we support content scripts which would have an origin outside the extension. So this could be dependent on if we want to include them in a (future iteration?) of the proposal?

For `incognito`: can an `ExtensionContext` change it's private status? I thought of it as a static but I'm new to the extension ecosystem. The idea for this property came from the case of [`spanning` mode](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito) where you might get multiple `ExtensionContext`s back in and out of private contexts and want to treat them differently.

Rob--W:
> @Rob--W Apologies I think by committing I prevented a directly reply to your comment

You didn't break anything; I posted a standalone comment because my feedback was not tied to one specific line.
I was asking about use cases because I felt that something was missing from the API. I'll respond to your last comments in line, and elaborate on the gaps in the end.

> Use cases: a possible example is to check that a popup is open. Otherwise the hacky way is to have `Context`s send keepalive messages back to the background script which maintains a table of what is open or not at any given time.

That hacky approach wouldn't even work because it implies some persistent in-memory state in the background, whereas service workers are forcibly shut down at some point (currently 5 minutes in Chrome).

> A popup is a common example I can think of, but any other `Context` an extension wants to often know is open without having to manually keep track of it would benefit from this. Unless there's another way of keeping track of the idea of a `Context`, every extension would implement this differently and doing such state tracking sounds prone to bugs when the tracked state doesn't match what's actually running. If that seems like sufficient use case I can add a section to the proposal to have that for future reference.

Even without querying the state of the popup, the existing extension messaging APIs (`runtime.sendMessage` / `runtime.connect`) can already be used to communicate with the popup if desired (and even track whether the context disappears - with `port.onDisconnect`). It is not strictly necessary to call `extension.getContexts()` first.

A common use of `extension.getViews()` (and `extension.getBackgroundPage()`) is to get a direct reference to the `window` object of the background or other views, in order to directly manipulate the DOM or global (e.g. caching values, rendering DOM). This use case cannot be supported by the proposed `extension.getContexts()` method because the functionality is incompatible with service workers.


> For uniquely targeting a specific destination, this brings us closer to that goal. Currently there's just broadcast, but now an extension could check if a `Context` is running before attempting to send a message to it to be a bit more efficient.

Non-broadcast semantics is desirable and requested at #294. It is currently missing a way to unambiguously describe any extension context (whether a tab, worker, content script, popup, sidebar/sidepanel, devtools panel, ...). Solving that would be valuable.

> For `incognito`: can an `ExtensionContext` change it's private status? I thought of it as a static but I'm new to the extension ecosystem. The idea for this property came from the case of [`spanning` mode](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito) where you might get multiple `ExtensionContext`s back in and out of private contexts and want to treat them differently.

A context cannot change its private status. It is fixed at the start of the context (dependent on where the context is loaded) and immutable afterwards. Note that the table from my previous comment (https://github.com/w3c/webextensions/pull/334#issuecomment-1358475431) lists the APIs that the recipient of the message can use to get the information to respond to the sender.


> Something I probably didn't elaborate on when describing these properties is that they might be empty/not defined for the items where it doesn't make sense -- but altogether would cover the various extension contexts that might be returned.

I asked for use cases because the value of the `getContexts()` method and returned properties is dependent on their use cases. The PR mentions the next use cases:

1. (from PR) "reading and/or modifying a settings page" - this is is not supported by `getContexts()`, extensions would have to use extension messaging anyway because they cannot directly access the DOM of UI pages such as the popup, options_ui, tab, etc.
2. (from PR) "determining if a toolbar popup is open" - requested at https://crbug.com/1322432.
3. (from PR) `OFFSCREEN_DOCUMENTS` type - referenced at https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/offscreen.idl;l=63-70;drc=bcffcd423659a28620ece7e155d05c6b82ed13db
   > Determines whether the extension has an active document. (...) Instead of this, we should integrate offscreen documents into a service worker-compatible getViews() alternative.
4. (from PR) "avoid [various workarounds](https://github.com/w3c/webextensions/pull/334#issuecomment-1343255899) needed to target specific pages with messages."

Use case 2 and 3 do not need any properties.
The two examples for use case 4 do not involve service workers: example 1 can continue to use `getViews()`. Example 2 is about direct access to methods of `window` and not supported by `getContexts()`. The broader interpretation of use case 4 (the ability to target a specific context for messaging) is #294.

Other use cases that I have encountered in practice is determining whether there is already an instance of an extension tab, and if not open it (and if there is, focus it). There are already various UI-specific APIs for this scenario (designed and implemented after `extension.getViews()`): `action.openPopup()`, `runtime.openOptionsPage()` and `tabs.query()`.

### Gaps

If I rewind, and start from the premise of "What gaps does the absence of `extension.getViews()` leave", then there are mainly two gaps:
1. `extension.getViews()` offered the ability to find all active extension contexts that can be interacted with.
2. "interacted with" is broadly defined as invoking any DOM API or extension API in the target context.
  - with the additional note that the target context is unambiguously identifiable due to the (synchronous) access to all globals/DOM in that context.
  - with the small note that we cannot support synchronous access due to the desire to support the method in service workers.

The first one is easy to satisfy by merely introducing the `extension.getContexts()` method proposed here.
The second one is very broad, but the most generic way to cover that is to somehow exchange messages between the current context (e.g. background) and the target context (e.g. popup), so that the target context can do something useful on demand. And for that, we would need an unambiguous way to describe the target context (for use with #294). The proposal currently returns a bag of properties, but the connection between these properties and a mechanism to reach the target context is unclear. It is currently not directly possible to determine which of the contexts is the current context (where the script is executing; for comparison: `getViews()` can be: `chrome.extension.getViews().filter(w => w !== window))`.

An easy answer could be to include "contextId" in the property set, and then support `contextId` in `runtime.sendMessage` and/or `runtime.connect`. The advantage of this is that this can finally fill the gap in extension APIs of uniquely describing a specific extension context, as tabId+frameId / documentId cannot describe non-tab contexts such as workers, devtools pages, sidebar, etc. A (minor?) disadvantage is that we'd be introducing yet another way to describe a context (after the recent introduction of "documentId"), and the mapping between contextId and the actual context may not be very obvious. But if such a unique ID were to be introduced, the following use cases could easily be covered as a bonus:
- https://github.com/w3c/webextensions/issues/8
- https://github.com/w3c/webextensions/issues/91
- https://github.com/w3c/webextensions/issues/294

A more complicated answer would be to extend APIs that target a context (`runtime.sendMessage`, etc.) with a `targetContext` option that takes a dictionary whose keys are a subset of the `ExtensionContext` (Context) type described here. This is more complex, because we would have to define the properties that can uniquely identify a single context. The main advantage of this mechanism is that it would also be usable outside of privileged extension contexts (i.e. content scripts).

I think that it's a must-have to define how extension messaging is intended to be used with this `getContexts()` method, in order to reach a state where the API serves a useful alternative to `extension.getViews()`.

hanguokai:
I think the ultimate goal should be two parts: 
1. Provide a way to find the target.
2. Communicate with this specific target.

### Provide a way to find the target
My first example in https://github.com/w3c/webextensions/pull/334#issuecomment-1343255899 is that the service worker communicate with the popup page by sending a message(note: this popup page can also be opened as a tab). This is a workaround for the problem 1) can't use `extension.getViews()` in service worker. 2) `runtime.sendMessage()`'s broadcast behavior.

My second example is finding a specific extension page first, then communicate with it by calling a method on it's `window` object. Because if you get a `window`, it is unnecessary to send a message(and sendMessage is broadcast, not only to the target).

The point of these two example is on how to find a specific target, not how to communicate. Here communication is the means to find the target.

**Current problem of finding the target**
- `extension.getViews()` can't be used in service worker
- `sendMessage()` is not an effective means of finding the target.
- `runtime.openOptionsPage()` only for the options page, can't be used for other extension pages.
- `browser.tabs.query({url})` by url can't work, if you don't have the "tabs" permission, but this permission should be unnecessary for finding an extension page itself.

I know `Client` API is another workaround for both finding the target and communicate with the target.


### Communicate with this specific target
@Rob--W think this is unclear at present. He thinks this should be clarified with this proposal and proposes #294. I think the main purpose of finding a target is to communicate with it, so it should be clear how to do that in this proposal.

proposals/extension_get_contexts.md
mukul-p: private lgtm.

xeenon: Yes, `private` is good.

rdcronin:
Thank you all for the thoughtful feedback.

I agree with most of what's been said here.  In my mind, I think there are two main benefits to this API:
1) It provides a _canonical_ way to determine if a given context is active.  At the outset, a "context" here would be limited to contexts that commit to the extension origin (tabs, popups, background contexts, offscreen docs, etc), but we'd like to expand this in the future to include contexts like content scripts, as well.
2) It _paves the way_ (but does not itself directly implement) for different APIs to leverage a way to uniquely identify a context.  The most obvious here is message passing (via sendMessage or connect), but there could be others in the future.

As has been mentioned, for 1), there are workarounds to this that exist today: predominantly, message passing itself (send a message from a created context and track them in the interested context; or, send out a message from the interested context and see if there's a reply).  This, to me, seems hacky and undesirable — it pollutes the message channels (which are currently broadcast everywhere), it's a bad fit for ephemeral contexts like service workers and event pages (since a message port can extend the lifetime of that context, and any incoming messages could result in the background context being spun up, even if it's not needed), and just feels generally "clunky".  `getViews()` can be used in some cases, but a) is only available in non-worker contexts and b) is fundamentally incompatible with any contexts that may run in different processes (content scripts, incognito).

There's also existing requests for a way to tell if an <x> context is open, such as [determining if an action popup is open](https://crbug.com/1322432) and determining if an offscreen document is present.  Rather than requiring either the use of message sending to determine if these contexts are active or introducing different methods for each individual context type we might be interested in (`action.hasOpenPopup()`, `offscreen.hasDocument()`, etc), it would be cleaner to introduce a canonical way to determine this.

This API is targeted at achieving 1).

After we have this API established, we'd like to use it to solve 2) — targeted messaging.  I'd like to avoid blocking this proposal on fully spec'ing what that change looks like, because I think it has its own set of considerations (do we want to more fully differentiate between "broadcast" methods and non-"broadcast" methods?  Are there other implications in the existing messaging APIs that we'd need to reconsider?).

Proposal: We introduce `contextId` to the bag-of-properties returned.  I was originally thinking this would be done as part of 2) (since it's not necessary for 1)), but I can understand the desire for it to be added here.  We agree that the messaging changes will, in essence, be that "we provide a way for an extension to specify a `contextId` to which a message should be sent", without firm commitments on the method signature or related aspects.

Given that, does this proposal seem reasonable _at a high level_?  I.e., providing an asynchronous method to determine which contexts are active for an extension by providing a bag-of-properties descriptor for each, which can later be used to target specific messages.

If so, I agree there are other aspects we may want to discuss more.  E.g.:
* Rob called out (and Justin and I were also discussing) that URL is ambiguous.  I think we'll probably want to change this to be `documentUrl` or similar, and have it be undefined for service worker contexts.  (We could potentially introduce a `scriptUrl` for worker contexts, which could also be set for context script contexts in the future, if we felt that was valuable.)  We'd probably want to do something similar for `origin`.
* We may re-brand context type, if we feel it's ambiguous.
We can definitely iterate on these, but I'd like to get consensus on the general shape of the API first.

Regarding TOCTOU:

In short, yes, there are TOCTOU risks.  Given extension contexts can run on different threads (worker vs main) and in different processes (content scripts, incognito contexts), I think asynchronous behavior is largely unavoidable.  (And, even if the extension were able to synchronously determine if a context were active, subsequent actions — like sending a message — are also async, and so have the same issue.)  Rather than trying to fully avoid TOCTOU issues, I think it's better to ensure we provide developers with the proper tools to cope with asynchronous APIs and events.  For instance, `documentId` and `contextId` can be used to help avoid issues that may arise.

proposals/extension_get_contexts.md
Rob--W: For new concepts onwards, I advocate for neutral names.
But "incognito" is already widespread in extension APIs.
@xeenon @mukul-p Is the confusion from different names for exactly the same concept worth the introduction of a new name? E.g.

- `extension.inIncognitoContext`
- `tabs.Tab.incognito`
- `windows.Window.incognito`
- `cookie.cookies.incognito`
- `downloads.DownloadItem.incognito`

Secondly, the relevance of `incognito` is that it reflects the extent to which DOM APIs share state across contexts, such as network state (e.g. cookies, cache), to help extensions with respecting the state isolation between different contexts. The "incognito" boolean is not enough for that. In Firefox, `cookieStoreId` is used as a generic way to distinguish contexts with different cookie stores. This name originates from Chrome's cookies extension API, and values are browser-specific opaque strings: Chrome uses "0" and "1", Firefox uses "firefox-default", "firefox-private", "firefox-container-(digits)" (the latter is used for container tabs - https://wiki.mozilla.org/Security/Contextual_Identity_Project/Containers ).

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+`contextProperties` will have similar behavior to `extension.getViews()`’s
+[`fetchProperties`](https://developer.chrome.com/docs/extensions/reference/extension/#method-getViews)
+object in that it will limit the scope of the query to certain tabs, context
+types, and windows. It would be defined as:
+
+```
+extension.ContextProperties = {
+  tabId? = int,                // Find context by tab id, omitted returns all tabs.
```
Rob--W, line 57: I suggest to rephrase "returns all tabs" - there are contexts that are not tabs. Similarly for windows".

Rob--W:
> In my mind, I think there are two main benefits to this API:
> 
>    1. It provides a _canonical_ way to determine if a given context is active.  At the outset, a "context" here would be limited to contexts that commit to the extension origin (tabs, popups, background contexts, offscreen docs, etc), but we'd like to expand this in the future to include contexts like content scripts, as well.
> 
>    2. It _paves the way_ (but does not itself directly implement) for different APIs to leverage a way to uniquely identify a context.  The most obvious here is message passing (via sendMessage or connect), but there could be others in the future.
>
> (...)
> This API is targeted at achieving 1).
> (...)
> 
> Given that, does this proposal seem reasonable _at a high level_? I.e., providing an asynchronous method to determine which contexts are active for an extension by providing a bag-of-properties descriptor for each, which can later be used to target specific messages.

This looks reasonable at a high level.

> If so, I agree there are other aspects we may want to discuss more. E.g.:
> 
>     * Rob called out (and Justin and I were also discussing) that URL is ambiguous.  I think we'll probably want to change this to be `documentUrl` or similar, and have it be undefined for service worker contexts.

Agreed. "url" / "frameUrl" should be renamed to "documentUrl".

>  (We could potentially introduce a `scriptUrl` for worker contexts, which could also be set for context script contexts in the future, if we felt that was valuable.)

There may be multiple scripts running in a worker or content script world. The only stable choice would be the first script that is running in it. 

>  We'd probably want to do something similar for `origin`.

I think that `origin` should reflect the `self.origin` value of that execution environment. For documents and workers that's already well-defined in the web platform. In content scripts it's the origin of the document.

So far we've mostly explored the returned properties. I'd also like to examine the input to `getContexts`. The current specified behavior is to "return all contexts, optionally with some filters".
When expanded to include content scripts, the API's default behavior would be very expensive to run. That issue could be solved by adding a parameter in `getContexts()` to opt in to including cross-origin results, and default to only return contexts that are same-origin with the extension.

With `extension.getViews()`, there was an unambiguous way to identify the current context. The current proposal does not have that, but the result can be approximated by the context properties (contextType, documentUrl, etc.). Do we consider that enough, or do we want to introduce an extra parameter to `getContexts()` (e.g. `currentContext: true`) or even a `runtime.getContextId()` method (also available to content scripts?).

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,189 @@
...
+`Context`s would require work in retrieving content script `Context`s explained
+[above](#content-script-contexts).
+
+## Open Questions
+
+### Include `sandbox` Pages and Iframes as `Context`s?
+
+`sandbox` documents and iframes have an
+[opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque),
+should we return them? What if they come from within the extension (or outside
```
Rob--W, line 177: IMO: no. Sandboxed documents have a `"null"` origin and don't have access to extension APIs. They're hardly different from regular sandboxed iframes from the web.

proposals/extension_get_contexts.md
```diff
@@ -0,0 +1,189 @@
...
+    thread). If called from a background script, it would likely only return the
+    background script `Context`.
+
+### Content Script Contexts
+
+[Content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/)
+run in a separate
+[Renderer](https://developer.chrome.com/blog/inside-browser-part3/) (and
+process) from the extension process. This proposal would not return those
+`Context`s since this adds significant complexity to the design in querying all
```
Rob--W, line 153: s/design/implementation/.

proposals/extension_get_contexts.md
xeenon: That makes sense @Rob--W, I am fine with sticking with `incognito` to keep the API consistent.

mukul-p: Agree with Rob that we should be consistent, hence incognito is good.

Compared to _ephemeral_ inPrivate looks better.

rdcronin:
Hey folks — just a quick status update here.  This has been on the back burner for the last couple weeks, but we still plan on making more progress on it this quarter.

rdcronin:
I'm going to be taking over this work for now.  Since Github doesn't (I think?) support taking over someone else's PR, there's a new PR for this here:

https://github.com/w3c/webextensions/pull/358

oliverdunk:
Closing this per Devlin's comment.

PR #345, opened 2023-01-20
Don't deploy for changes to _minutes/

Rob--W:
The content of _minutes/ never affects the publication on the gh-pages branch. To avoid unnecessary deploy builds when we publish meeting notes, add `_minutes/**` to the paths to ignore in the deploy workflow.

`paths-ignore` is documented at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths


dotproto (approved):
LGTM

Issue #113, opened 2021-10-20
Proposal: Make extension APIs browser neutral


carlosjeurissen:
Currently several aspects of the webExtension platform reference specific browsers. This should be streamlined across all browsers.

Examples are:
- the main API namespace. Google Chrome uses `chrome` while Mozilla Firefox, the old Microsoft Edge and Safari also support the browser-neutral `browser`. A request to also support `browser` in Google Chrome has been filed here: [CrBug 798169](https://bugs.chromium.org/p/chromium/issues/detail?id=798169)
- The `chrome_settings_overrides` manifest property could be replaced with the neutral `browser_settings_overrides`
- The `chrome_url_overrides` manifest property could be replaced with the neutral `browser_url_overrides`
- The `omnibox` API, mentioned here https://github.com/w3c/webextensions/issues/113#issuecomment-951104552
- `incognito` manifest property

Let me know if I'm missing any of them.

xeenon:
The `omnibox` API is another example that should have a browser-neutral name.

carlosjeurissen:
Added both `omnibox` and `incognito`. Seems `private` is more common in the browser market.

What about `devtools_page`?

tophf:
FWIW, `chrome` was a term for UI internals aka "browser chrome" ([coined by Mozilla, probably](https://firefox-source-docs.mozilla.org/devtools/files/adding-files.html#chrome-content)) and Google used this term as a [reverse joke](https://www.quora.com/Why-is-Google-Chrome-browser-named-as-Chrome/answer/Glen-Murphy), so I never considered `chrome` to mean the name of a particular browser.

dotproto:
Anecdotally, the oldest concrete references I've been able to find for "chrome" in the GUI sense are [Jargon File v2.1.1](http://magic-cookie.co.uk/jargon/jarg211/jargon.htm) (1990) and Netscape Navigator's source ([Mozilla FTP](http://ftp.mozilla.org/pub/mozilla/source/), [GH mirror](https://github.com/zii/netscape)) (1995-1998?).

js-choi:
Whether chrome is a browser-neutral term, it would be good for interoperability to harmonize the naming of the `chrome`/`browser` object between implementations.

xeenon:
Safari already supports `browser_url_overrides` as a synonym along with a `browser_specific_settings` in the manifest. Same for `browser` as the namespace JS object.

abitrolly:
[...]

PR #358, opened 2023-03-03
Add `runtime.getContexts()` proposal

rdcronin:
This adds a proposal for a new `runtime.getContexts()` API.

This work is based off the original pull request from justinlulejian [here](https://github.com/w3c/webextensions/pull/334).


dotproto:
FYI for reviewers: you can view a rendered version of the proposal Markdown by navigating to the Files Changed tab, opening the options menu for the file, and selecting "View file". 

Screen Shot 2023-03-03 at 11 31 00 AM

dotproto:
@rdcronin, the W3C's IPR (Intellectual Property Rights) bot was not able to verify your membership in the group. To fix this, you'll need to create a W3C account (if you don't already have one), make sure your [GitHub account is linked to your W3C account](https://lists.w3.org/Archives/Public/public-publ-wg/2017Sep/0021.html), and [join the WebExtensions Community Group](https://www.w3.org/community/webextensions/join).

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
+# New API: runtime.getContexts()
+
+## Background / Summary
```
dotproto: I'd suggest renaming this to just "Background" as this section does not currently make reference to `runtime.getContexts()` or the concept of a context.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Solution
+
+Considering the above situation, we'd like to propose a new extension API
+method, `runtime.getContexts()`, to asynchronously provide metadata about
+associated contexts that is still useful for an extension. This will allow
+extension background scripts to identify the active contexts in the extension.
+For example, this can be used to target messages to send using
+[`runtime.sendMessage()`](https://developer.chrome.com/docs/extensions/reference/runtime/#method-sendMessage),
+etc.). Introducing this API will allow an easier migration from MV2.
```
dotproto: The term "context" as used in `runtime.getContexts()` is not defined. This contrasts with the Background section which introduces `extension.getViews()` and describes what the concept of a "view" encompasses.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+
+This also obviates the need for introducing multiple one-off APIs, such as
+separate APIs to query for the extension popup, offscreen documents, etc.
+
+## API Proposal
+
+This method will return an array of matching contexts, represented by a new
+`ExtensionContext` type. This will be defined as:
+
+```
```
dotproto: Use JS code blocks to help improve legibility.

````suggestion
```js
````

File: proposals/runtime_get_contexts.md [hunk 4]
```diff
@@ -0,0 +1,233 @@
...
+  documentOrigin?: string,
+  // Whether the context is for an incognito profile.
+  incognito: boolean,
+}
+```
+
+`ContextType` will indicate the type of context retrieved. It is an enum
+defined as:
+
+```
```
dotproto: Use JS code blocks to help improve legibility.

````suggestion
```js
````

**Comment by rdcronin** [comment]
Request to join sent.  (I had linked Github + W3C, but hadn't yet joined)

File: proposals/runtime_get_contexts.md [hunk 5]
```diff
@@ -0,0 +1,233 @@
...
+This also obviates the need for introducing multiple one-off APIs, such as
+separate APIs to query for the extension popup, offscreen documents, etc.
+
+## API Proposal
+
+This method will return an array of matching contexts, represented by a new
+`ExtensionContext` type. This will be defined as:
+
+```
+runtime.Context = {
```
dotproto: This object appears to be referenced elsewhere (including the proceeding paragraph) as "ExtensionContext."

```suggestion
runtime.ExtensionContext = {
```

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+
+```
+runtime.getContexts(
+  filter?: ContextFilter
+): Promise<ExtensionContext[]>;
+```
+
+The `filter` argument will be used to filter down to a particular type of
+context. It will share the same properties as `ExtensionContext`, but will
+have all fields be optional. Any omitted field matches all available contexts.
```
dotproto: If I'm reading this correctly, a filter can either match against a single value for a given property or any value, but it's not possible to match against a _set_ of values. For example, if the extension wanted to retrieve a set of user visible contexts they would have to execute two `getContexts()` calls:

```js
async function getUserVisibleContexts() {
  const contexts = await Promise.all([
    browser.runtime.getContexts({filter: {contextType: browser.runtime.ExtensionContext.TAB}}),
    browser.runtime.getContexts({filter: {contextType: browser.runtime.ExtensionContext.POPUP}}),
  ]);
  return contexts.flat();
}
```

It may be preferable to allow developers to specify a single value or an array of values for properties.

```js
async function getUserVisibleContexts() {
  return browser.runtime.getContexts({
    filter: {
      contextType: [
        browser.runtime.ExtensionContext.TAB,
        browser.runtime.ExtensionContext.POPUP
      ]
    }
  });
}
```

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+}
+```
+
+`ContextType` will indicate the type of context retrieved. It is an enum
+defined as:
+
+```
+extension.ContextType = {
+  // Tabs the extension is running in.
+  TAB: 'TAB',
```
dotproto, line 95: Does `ContextType TAB` cover both normal extension pages and sandboxed pages?

Does `ContextType.TAB` cover only extension pages loaded in the main frame of a tab or does it include sub-frames? 

Does `ContextType.TAB` cover other documents in a tab such as those in BFCache and preloaded pages? If so, does it also cover sub-frames of those documents?

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+  documentId?: string,
+  // ID of the frame for this context, or -1 if this context is not hosted in a
+  // frame.
+  frameId: int,
+  // The current URL of the document, or undefined if this context is not
+  // hosted in a document.
+  documentUrl?: string,
+  // The current origin of the document, or undefined if this context is not
+  // hosted in a document.
+  documentOrigin?: string,
```
dotproto, line 83: Do `documentUrl` and `documentOrigin` refer to the document that the extension context is inside (e.g. top level main frame of a tab), the extension context itself, or something else? Depending on the answer, I may have other questions.

The "Sandboxed Pages" section notes that this method will not return sandboxed pages as they have "a separate origin (`"null"`) and do not have access to extension APIs." I find this somewhat surprising as I had assumed that the `documentOrigin` property to the origin of the extension context and therefore this property would either match the extension's origin (i.e. the value returned by evaluating `new URL(browser.runtime.getURL("")).origin`) for normal extension pages or `null` for sandboxed pages.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+  windowId: int,
+  // ID of the DOM document for this context, or undefined if this context is
+  // not associated with a document.
+  documentId?: string,
+  // ID of the frame for this context, or -1 if this context is not hosted in a
+  // frame.
+  frameId: int,
+  // The current URL of the document, or undefined if this context is not
+  // hosted in a document.
+  documentUrl?: string,
```
dotproto, line 80: When used as a filter parameter, is it possible to use a match pattern for the value of `ContextFilter.documentUrl`?

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+  // The current origin of the document, or undefined if this context is not
+  // hosted in a document.
+  documentOrigin?: string,
+  // Whether the context is for an incognito profile.
+  incognito: boolean,
```
dotproto, line 85: In other issues we've discussed changing the terminology for private browsing mode. At the moment the only reference I can find is #113, but I recall there was another comment thread where @xeenon and I discussed naming in more detail.

```suggestion
  private: boolean,
```

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+#### Context ID
+
+Each extension context will have a unique context ID, represented by a string.
+This is necessary to uniquely identify a context, since other fields may be
+non-unique (such as URL) or absent (such as documentId).
+
+Like `documentId`s, the extension `contextId` will update on (non-same-page)
+navigation.
+
+#### Incognito mode
```
dotproto, line 184: This section currently describes split mode but does not directly address spanning mode. It seems reasonable to assume that in spanning mode `getContexts()` will return both private and non-private contexts, but I'd prefer to be explicit about that.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+
+[Split-mode](https://developer.chrome.com/docs/extensions/mv3/manifest/incognito/#split)
+extensions will _not_ have access to the contexts from their corresponding
+profile. That is, the incognito extension process will not be able to access
+contexts from the non-incognito extension process, and vice versa.
+
+#### Sandboxed Pages
+
+Sandboxed pages will not be included in the returned contexts. They are
```
dotproto: Grammar fix.

```suggestion
Sandboxed pages will not be included in the returned contexts. They are on
```

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+### Additional Considerations
+
+#### Context ID
+
+Each extension context will have a unique context ID, represented by a string.
+This is necessary to uniquely identify a context, since other fields may be
+non-unique (such as URL) or absent (such as documentId).
+
+Like `documentId`s, the extension `contextId` will update on (non-same-page)
+navigation.
```
dotproto: I don't quite follow this paragraph. As I understand it, a document's `documentId` would persist as it moves between pretender, active document, and BFCache states. Is the intent for `contextId` to change during these transitions?

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+document). Some of these values -- `tabId`, `windowId`, and `frameId` -- use
+constant values to indicate "no state", such as `-1` for `tabId`. Others --
+`documentId`, `documentUrl`, and `documentOrigin` -- use undefined to indicate
+this.
+
+This is an artifact of existing APIs and precedence. Since many existing APIs
+use the constant integer values, we want to be consistent with those. However,
+for newly-introduced fields, we use the more intuitive `undefined` state.
+
+## Future Work
```
dotproto, line 239: Just wanted to note that I think it's super useful to have a section like this. IMO it's very helpful to communicate to readers that the current proposal does not aim to solve all of the issues a developer might have with targeting and communicating across contexts in their extension. Additionally, this explicitly signals other areas that have been identified, considered, and deemed out of scope for the work we're currently target.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Messaging APIs Support `ContextId`s as Target
+
+In practice, many extension messages are meant for a single target, but are
+broadcast to all extension contexts. With the ability to uniquely identify a
+single extension context, we will modify messaging APIs (such as
+`runtime.sendMessage()` and `runtime.connect()`) to allow specifying specific
+targets that should receive a message.
+
+### Multi Page Architecture Fields (`DocumentLifeCycle`, `FrameType`, and `parentDocumentId`, and `parentFrameId`)
```
dotproto, line 249: Note that there does not appear to be good, cross-browser terminology to refer to having multiple "documents" in a single "window." The [Web Platform Design Principles](https://www.w3.org/TR/design-principles/#support-non-fully-active) Group Note uses the term "BFCached documents" and references "non-[fully active](https://html.spec.whatwg.org/multipage/document-sequences.html#fully-active) BFCached documents". I'm somewhat hesitate to use a similar pattern here as both BFCached and preloaded documents may cause a single window to contain multiple documents.

dotproto (approved):
Overall this proposal looks good to me and I'm very supportive of the direction. I have a several questions about the current draft, but I don't think any of them are blocking for experimental implementation purposes.

proposals/runtime_get_contexts.md
rdcronin: Done

proposals/runtime_get_contexts.md
rdcronin: I added a bit about this.  Since the context is a bit fuzzy (it's something that's running extension code, but we're not starting off with including everything), it's a little vague, but now it's explicitly vague : )

proposals/runtime_get_contexts.md
rdcronin: Done

proposals/runtime_get_contexts.md
rdcronin: Thanks for catching!  Done

proposals/runtime_get_contexts.md
rdcronin: Currently, no.  I've thought about if we'd want to support a different filter object (so you could specify e.g. array of types, match patterns, etc), but I wasn't sure if it was worth it.  It seems like you'd normally either a) be able to specify an explicit URL or b) identify the contexts in different ways (say, by context type).

If folks feel strongly that we should support a richer filter from the beginning, I'm amenable to introducing a new ContextFilter type.  Otherwise, we could potentially do this in the future (by supporting `choices` for the appropriate fields).  Lemme know what y'all think.

proposals/runtime_get_contexts.md
rdcronin: > Do documentUrl and documentOrigin refer to the document that the extension context is inside (e.g. top level main frame of a tab), the extension context itself, or something else? Depending on the answer, I may have other questions.

They refer to the document the extension context is executing in, so it depends on the context:
* Anything frame-based: correspond to the extension (e.g. chrome-extension://<id>/popup.html)
* Service workers: undefined
* Content scripts: document the script is injected within.

I've added a new section on this in "Additional Considerations"

> The "Sandboxed Pages" section notes that this method will not return sandboxed pages as they have "a separate origin ("null") and do not have access to extension APIs." I find this somewhat surprising as I had assumed that the documentOrigin property to the origin of the extension context and therefore this property would either match the extension's origin (i.e. the value returned by evaluating new URL(browser.runtime.getURL("")).origin) for normal extension pages or null for sandboxed pages.

This was inline with Rob's suggestion on the old PR to not include sandboxed pages as contexts here, which I agree with (they are reasonably separate from the rest of the extension).  We could always change that in the future by adding a new context type.

proposals/runtime_get_contexts.md
rdcronin: This was discussed in the old pr [here](https://github.com/w3c/webextensions/pull/334/files#r1051164878), where @xeenon and @Rob--W agreed that "incognito" made sense here.

proposals/runtime_get_contexts.md
rdcronin: Done

proposals/runtime_get_contexts.md
rdcronin: > Does ContextType TAB cover both normal extension pages and sandboxed pages?

Sandboxed pages are not returned in this API.  This is described below already.

> Does ContextType.TAB cover only extension pages loaded in the main frame of a tab or does it include sub-frames?
> Does ContextType.TAB cover other documents in a tab such as those in BFCache and preloaded pages? If so, does it also cover sub-frames of those documents?

`ContextType` refers to the overall embedder for a context; for instance,
`ContextType.TAB` will be used for any context associated with a tab. This
includes subframes and cached frames. Developers can determine the finer
grained state of these via other properties, such as `frameId`.

I've added this to the doc.

proposals/runtime_get_contexts.md
rdcronin: See above.

proposals/runtime_get_contexts.md
rdcronin: The `contextId` is unique and persistent for the lifetime of a context.  For
contexts associated with a document, this means `contextId` behaves like
`documentId`. For service worker contexts, the `contextId` is unique for the
duration of the service worker context (but will be different across service
worker restarts).


I've added this to the doc.

proposals/runtime_get_contexts.md
rdcronin: spanning mode is... weird.

Try embedding an iframe in an incognito window in a spanning mode extension: you get a pseudo-split-mode-but-not-really extension.

I didn't really want to spec out what happens here.  I've taken a page from the fetch API and said "it's left as an exercise to the implementor."

proposals/runtime_get_contexts.md
rdcronin: Done

proposals/runtime_get_contexts.md
rdcronin: Ack; thanks!

proposals/runtime_get_contexts.md
rdcronin: Since there isn't really cross-browser lingo for this, I'd prefer to not stress about it in this proposal.  If/when there is, I'm happy to update public documentation to use it.

rdcronin (commented):
Thanks, Simeon!  Comments addressed.

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,259 @@
...
+  // Whether the context is for an incognito profile.
+  incognito: boolean,
+}
+```
+
+`ContextType` will indicate the type of context retrieved. It is an enum
+defined as:
+
+```js
+extension.ContextType = {
```
bershanskiy, line 93: Is this enum meant to be exhaustive? Can other browsers extend it? For example, a number of browsers support [sidebars](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Sidebars). Some extensions for Firefox and Vivaldi currently struggle with messaging with these contexts (since sidebars do not have a meaningful `tabId` and `frameId` for messaging. Developers can create persistent ports, but it is ugly.

rdcronin: As with all APIs discussed in this community group, it's non-binding, and browser vendors can deviate when it makes sense.  If a browser supports a different type of context, they can add it in here, and if they don't support a given context, they can remove the enum entry.

FWIW, Chrome is also going to support side panels, so we will add a SIDE_PANEL context type when that happens.  (I was tempted to include it here, but didn't want to confuse the proposal by having unshipped concepts).

proposals/runtime_get_contexts.md
```diff
@@ -0,0 +1,259 @@
...
+Some common reasons for calling this method are: determining if a toolbar
+popup, tab, or options page is open; directly interacting with those pages by
+reaching into their `HTMLWindow`; etc.
+
+Chromium's implementation of Manifest V2 (MV2) allows for
+[background pages](https://developer.chrome.com/docs/extensions/mv2/background_pages/)
+to call this API. This is possible because these pages are themselves an
+extension frame (albeit one that isn't visibly rendered) and run on the main
+thread of the renderer; this allows them easy access to the JavaScript
+[`Window`](http://go/mdn/API/Window#instance_properties) objects provided by
```
Rob--W:
```suggestion
[`Window`](https://developer.mozilla.org/en-US/docs/Web/API/Window#instance_properties) objects provided by
```

go/ redirector does not resolve publicly.

Rob--W (approved):

proposals/runtime_get_contexts.md
Rob--W: @rdcronin Could you list all types that you're thinking of for Chrome? If you'd like, prefaced by a comment `// (Chrome-specific)`, so that we can see where we could converge towards a common type if desired.

E.g. I would expect at least devtools and devtools panel to have their distinct type here (as these do not fit in any of the other existing categories).

rdcronin: Thanks for the call-out!  Devtools are interesting -- in Chromium, they commit to a different origin (even though from the extension's POV, they are just the extension page).  They are also not currently included at all in extension.getViews().  I'd like to tackle these separately, but I've added an explicit section for them in future work.

I also added in a type for SIDE_PANEL (though it's still under development in Chrome)

proposals/runtime_get_contexts.md
rdcronin: Updated

rdcronin:
Thanks for the feedback, all!  I think all comments have been addressed.  @Rob--W , if this looks good, please feel free to merge it in.

PR #359, opened 2023-03-03
Add interface definitions for Chromium

dotproto:
This pull request introduces an initial set of WebExtensions definitions for the Chromium project. Additional details on how the data was gathered can be found in `interfaces/chromium/README.md`.


interfaces/chromium/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. Download the `.tar.giz` archive for each directory.
+2. Extract each `.tar.gz` file to a directory.
+3. Remove all files except those with a `json` or `idl` extension.
+    * Use the following Bash command or appropriate equivalent
+
+      ```bash
+      find -E . -type f ! -iregex ".*\.(json|idl)" -exec rm {} \;
+      ```
+
```
Rob--W, line 21: Would it make sense tot explain the meaning of these files, how to interpret them? Perhaps by linking to https://source.chromium.org/chromium/chromium/src/+/main:chrome/common/extensions/api/_features.md?

Rob--W (commented):
This patch includes files that are specific to legacy Chrome Apps, as well as private APIs.

The Chrome Apps APIs can certainly be omitted, as they (chrome/common/apps/) cannot be used in any extension.

If we do that, then we can reduce the number of results from 203 to:

- Chromium code search: `(file:^chrome/common/extensions/api/ OR file:^extensions/common/api/) (file:.json$ OR file:.idl)`
  - 198 results
  - URL: https://source.chromium.org/search?q=(file:%5Echrome%2Fcommon%2Fextensions%2Fapi%2F%20OR%20file:%5Eextensions%2Fcommon%2Fapi%2F)%20(file:.json$%20OR%20file:.idl)&sq=&ss=chromium

If I omit private and clear internal implementation details, then we have:

- Chromium code search: `(file:^chrome/common/extensions/api/ OR file:^extensions/common/api/) (file:.json$ OR file:.idl) -file:_private -file:_internal`
  - 132 results
  - URL: https://source.chromium.org/search?q=(file:%5Echrome%2Fcommon%2Fextensions%2Fapi%2F%20OR%20file:%5Eextensions%2Fcommon%2Fapi%2F)%20(file:.json$%20OR%20file:.idl)%20-file:_private%20-file:_internal&ss=chromium

There are still many non-public APIs. If you'd like to be extra precise, you could look at entries with "extension_types" containing "extension" WITHOUT "whitelist" at [chrome/common/extensions/api/_permission_features.json](https://source.chromium.org/chromium/chromium/src/+/main:chrome/common/extensions/api/_permission_features.json) and [extensions/common/api/_permission_features.json](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/_permission_features.json). This will give a list of public APIs (filter by `"channel": "stable"` to only get stable APIs, but for the purpose of this spec we probably accept pre-release API definitions too). Next, filter by APIs that depend on non-public permissions (listed at [chrome/common/extensions/api/_api_features.jso](https://source.chromium.org/chromium/chromium/src/+/main:chrome/common/extensions/api/_api_features.json) and [extensions/common/api/_api_features.json](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/_api_features.json)).

xeenon (approved):

interfaces/chromium/chrome-common-extensions-api/autofill_private.idl
```diff
@@ -0,0 +1,356 @@
+// Copyright 2015 The Chromium Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Use the <code>chrome.autofillPrivate</code> API to add, remove, or update
+// autofill data from the settings UI.
+namespace autofillPrivate {
```
mukul-p: Since namespace*Private are not really publicand are usable by the Extension developers, so adding to this repo would not be correct

mukul-p (commented):
There is a mix of idl and json files. This is a good start, but i feel that having a common format (for  one platform) would be consistent. Comments?

interfaces/chromium/README.md
dotproto: I just added a note to the end of the README that links out to this documentation.

dotproto:
> There is a mix of idl and json files. This is a good start, but i feel that having a common format (for one platform) would be consistent. Comments?

IMO that would be ideal, but for better or worse Chromium currently uses a mix of formats. Normalizing them here is out of scope. This PR only aims to provide a snapshot of Chromium's interface definitions and provide guidance on how to manually update them in the future.

interfaces/chromium/chrome-common-extensions-api/autofill_private.idl
dotproto: In response to this comment and [Rob's feedback](https://github.com/w3c/webextensions/pull/359#pullrequestreview-1329504658), I've stripped all `_private` and `_internal` files from this PR.

Issue #361, opened 2023-03-05
Add an API to integrate with the Credential Management Web API


lukewarlow:
This proposal is to provide a way for password manager extensions to integrate with website authentication flows more directly. This would allow site users the UX benefits of the [credential management API](https://developer.mozilla.org/en-US/docs/Web/API/Credential_Management_API) while also being able to use a password manager of their choice not just the default browser one.

Currently password manager extensions have to rely on heuristics and / or overriding the `navigator.credentials` JS API to effectively work for their users. I've experienced first hand issues with password managers not always being able to effectively determine when a successful registration or login occurs especially on single page apps where there's no conventional page reload. If sites could reliably make use of `navigator.credentials` for Password credentials and the browser allow the user to choose a provider rather than having only the choice of (currently only Chromium) the browsers built manager it would massively improve the experience, for end users, website developers and password manager developers.

An example of an extension being forced to override the JS API itself is Dashlane's Passkey saving [functionality](https://support.dashlane.com/hc/en-us/articles/7888558064274-Passkeys-in-Dashlane). (Not directly mentioned in their article but I've tested and that's what the extension is doing).

I imagine this API could work similarly to Android 14s new credential management API where apps use a standard API (which the web already has) and then there's a corresponding API that apps can implement to handle password and Passkey create and get requests.

lukewarlow:
Some very quick thoughts on what this could look like.

```ts
// Implementation of a new ExtensionCredentialsContainer which could probably be the same as (or similar to) the CredentialsContainer type (type of navigator.credentials)
const implementation: ExtensionCredentialsContainer = { }

// Extension API to register the container. Perhaps supported credentials could be declared in the manifest instead of programmatically?
chrome.credentials.registerContainer(implementation, {supportedCredentials: ['password', 'public-key']});
```

```js
// New function inside of navigator.credentials to return the supported credential types.
// This could replace searching for Credential types on the window object to allow extensions to support types that browsers don't natively support. Such as password inside of Firefox and Safari?
navigator.credentials.getSupportedTypes()
```

Preferably the UA could act as a middleware that delegates to multiple places at once, listing credentials from multiple providers? Not sure how this would work with custom credentials?

Like this crude mockup:

![image](https://user-images.githubusercontent.com/32498324/223438447-70e0f633-f4e7-4ef7-89ea-663a2bf0876e.png)
 
If the custom credential stuff opens up too many issues it'd be much better to not support that case rather than not doing this API at all imo.

[...]

PR #368, opened 2023-03-18
Add `browser.credentials` proposal

lukewarlow:
Rough draft of a proposal related to https://github.com/w3c/webextensions/issues/361

To clarify I've got no affiliation with any password manager companies this is just based on my experience as a web developer and web user.

### DIFF
--- a/proposals/credential-management-mock-pagescript.js
+++ b/proposals/credential-management-mock-pagescript.js
@@ -1,0 +1,74 @@
+const code = document.currentScript.getAttribute("credentialsContainer");
+const supportedTypes = JSON.parse(document.currentScript.getAttribute("supportedTypes"));
+document.currentScript.remove();
+
+const extensionCredentialsContainer = JSON.parse(code);
+
+extensionCredentialsContainer.create = eval(`(function ${extensionCredentialsContainer.create})`);
+extensionCredentialsContainer.store = eval(`(function ${extensionCredentialsContainer.store})`);
+extensionCredentialsContainer.get = eval(`(function ${extensionCredentialsContainer.get})`);
+
+function getTypeFromOptions(options) {
+    // Perhaps there's a better way to determine this type
+    let type = "";
+    if ("password" in options) {
+        type = "password";
+    } else if ("publicKey" in options) {
+        type = "publicKey";
+    } else {
+        throw new Error("Unsupported credential type");
+    }
+
+    return type;
+}
+
+const credentialsContainer = {
+    _internal: navigator.credentials,
+    create(options) {
+        // Browser would validate options to make sure it exists and is an expected shape (how could custom types work with this?)
+        if (!options) {
+            return Promise.reject(new DOMException());
+        }
+
+        const type = getTypeFromOptions(options);
+
+        if (supportedTypes.includes(type)) {
+            return extensionCredentialsContainer.get(options);
+        }
+        return this._internal.get(options);
+    },
+    get(options) {
+        // Browser would validate options to make sure it exists and is an expected shape (how does custom work with this?)
+        if (!options) {
+            return Promise.reject(new DOMException());
+        }
+
+        const type = getTypeFromOptions(options);
+
+        if (supportedTypes.includes(type)) {
+            return extensionCredentialsContainer.get(options);
+        }
+        return this._internal.get(options);
+    },
+    store(credential) {
+        // Browser would validate credential to make sure it exists and is an expected shape (how does custom work with this?)
+        if (!credential) {
+            return Promise.reject(new DOMException());
+        }
+
+        if (supportedTypes.includes(credential.type)) {
+            return extensionCredentialsContainer.store(credential);
+        }
+
+        return this._internal.store(credential);
+    },
+    preventSilentAccess() {
+        // This would be something the browser handles and the extension wouldn't need to worry about this
+        return this._internal.preventSilentAccess();
+    },
+}
+
+Object.defineProperty(navigator, 'credentials', {
+    value: credentialsContainer,
+});
+
--- a/proposals/credential-management.md
+++ b/proposals/credential-management.md
@@ -1,0 +1,128 @@
+# browser.credentials
+
+WebExtension proposal to allow interfacing with the [Credential Management API](https://w3c.github.io/webappsec-credential-management)
+
+## Motivation
+
+Password manager extensions currently rely on a mixture of heuristics and API overriding to detect and store credentials.
+To fill credentials they also rely on DOM APIs to detect relevant input fields to display UI (or autofill).
+
+Browsers have APIs for interaction with various credential types, most notably PublicKey and Password (Chromium only). However, these aren't exposed to browser extensions.
+
+## Proposal
+
+We propose a new browser.credentials API that would allow extensions to integrate with these existing web APIs.
+
+A mock for this proposal is available [here](credential-management-mock.js) and the corresponding web_accessible_resource that's required is [here](credential-management-mock-pagescript.js).
+
+## Types
+
+```ts
+interface ExtensionCredentialsContainer {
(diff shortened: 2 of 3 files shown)


lukewarlow:
See https://github.com/lukewarlow/demo-credential-manager-extension for a rough example of a usage of this API (well the mock for it).

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Proposal
+
+We propose a new browser.credentials API that would allow extensions to integrate with these existing web APIs.
+
+A mock for this proposal is available [here](credential-management-mock.js) and the corresponding web_accessible_resource that's required is [here](credential-management-mock-pagescript.js).
+
+## Types
+
+```ts
```
lukewarlow, line 20: Not sure how we feel about TypeScript types would probably be better as WebIDL

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+
+browser.credentials.registerContainer(credentialsContainer);
+```
+
+**browser.credentials.isActiveContainer**
+
+This retrieves a boolean indicating if this extension is the active credential manager.
+
+```js
+browser.credentials.isActiveContainer();
```
lukewarlow, line 66: Potentially this should be per credential type depending on whether browsers let users pick their manager per credential type?

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+## Manifest
+
+To make use of the browser.credentials namespace a new `credentials` key would be required inside of the extension manifest.
```
lukewarlow, line 71: Not sure if this or a programmatic approach with an options object passed to browser.credentials.registerContainer would be better?

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+
+Today password manager extensions use a signals such as heuristics for detecting login forms as well as signals such as a page navigation to determine successful login. Along with overriding of navigator.credentials to intercept requests for Passkeys.
+
+The first one is susceptible to false results, single page applications lack the conventional navigation on success. Some sites also navigate on error which extensions could misinterpret as a success.
+
+**Are there any extensions that would use this?**
+
+Any conventional password manager extension would benefit from this API. However, those looking to support Passkeys would particularly benefit.
+
+The Dashlane browser extension, a password manager with over 4M weekly active users in the Chrome Web Store, currently provides functionality to create and retrieve Passkey credentials.
```
lukewarlow, line 100: Based on looking at the extensions source code I've determined that Dashlane (and I imagine soon 1Password) make use of code very similar to the mock where a content script injects a "page" script that overrides the navigator.credentials implementation. However, they don't rely on evil evals because they can just put their implementation inside the page script.

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+The 1Password password manager also looks set to support Passkeys soon.
+
+**Would this API encourage use of passwords which are bad for security?**
+
+While it could be seen to be encouraging password authentication, this API (and its Web API sibling) on its own wouldn't change the fact that the vast majority of websites use passwords.
+This API would help to encourage best practices such as strong, unique per site passwords by making them more ergonomic to use.
+
+This API would also help extensions and in turn websites embrace the passwordless future.
+
+**Only Chromium implements the password credential type. Would this limit the usefulness of this API?**
```
lukewarlow, line 111: This to me is this biggest blocker to the usefulness of this API without some sort of "custom" credentials support.

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+While it's a product decision whether to support specific credential types, this API could potentially allow extensions to support types not natively supported by the browser. Feature detection would need to be considered with this.
+
+## Custom credential types
+
+Custom credential types could be a useful extension to this API.
+Allowing more niche credential types to be supported without needing to be fully specified and implemented by browser vendors.
+
+They could also address the lack of password credential support.
+
+These would probably require a change to the [Credential Management API](https://w3c.github.io/webappsec-credential-management) to provide a way to check supported credential types.
```
lukewarlow, line 124: Depending on how we handle the password type this may be required regardless.

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+## Motivation
+
+Password manager extensions currently rely on a mixture of heuristics and API overriding to detect and store credentials.
+To fill credentials they also rely on DOM APIs to detect relevant input fields to display UI (or autofill).
+
+Browsers have APIs for interaction with various credential types, most notably PublicKey and Password (Chromium only). However, these aren't exposed to browser extensions.
+
+## Proposal
+
+We propose a new browser.credentials API that would allow extensions to integrate with these existing web APIs.
```
lukewarlow, line 14: Need more detail on what the expected integration looks like from a UI /UX perspective. Potentially a UI mock could be useful but the exact details rely on unknowns such as whether to have 1 credential manager or multiple based on credential type.

proposals/credential-management-mock-pagescript.js
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        if (supportedTypes.includes(credential.type)) {
+            return extensionCredentialsContainer.store(credential);
+        }
+
+        return this._internal.store(credential);
+    },
+    preventSilentAccess() {
+        // This would be something the browser handles and the extension wouldn't need to worry about this
```
lukewarlow, line 66: I imagine the browser would handle the logic for this. That would rely on Firefox and WebKit implementing this "properly" though?

proposals/credential-management.md
```diff
@@ -0,0 +1,128 @@
...
+## Custom credential types
+
+Custom credential types could be a useful extension to this API.
+Allowing more niche credential types to be supported without needing to be fully specified and implemented by browser vendors.
+
+They could also address the lack of password credential support.
+
+These would probably require a change to the [Credential Management API](https://w3c.github.io/webappsec-credential-management) to provide a way to check supported credential types.
+
+## Remaining Work
```
lukewarlow, line 126: Main questions (beyond the obvious "is this something we want?"

Exact behaviour of choosing whether to use an in-built vs specific extension credential manager:

- Do we allow multiple providers with the browser delegating between them all? Or should it just be one manager either built in or a single extension?

How does an extension determine if a site is using the credential management API with their login forms?:

- Potentially this is something that just accepted as not possible? 
- Without knowing either way you'd still have to use the old approach for password credentials and could still run into false results (positives being worse than negatives because if the site DOES use this API you will always know, whereas with positives there's no API to say DON'T save this)

Should all types be allowed to be handled by extensions?:

- Not sure if federated or the new Identity (which I think is Federated V2) make sense to be handled by extensions?

Types not natively supported by all browsers:
- Passwords
- OTPs (only SMS is supported by any browser)

Custom types not supported by any browser:

- e.g. https://github.com/w3c/webappsec-credential-management/issues/152 (not sure what this is about but it's a proposal for a new type that potentially could be implemented by an extension)

lukewarlow:
Another use case beyond Passkeys and Passwords could be the OTPCredential type currently supported by Chromium. Currently chromium only supports the SMS transport but credentials could support others. Bitwarden (and I'm sure many other extensions) let you store and retrieve time based one time passwords. This API (along with probable required additions to the Credential Management API) could allow them to be programatically managed.

proposals/credential-management.md
lukewarlow, line 66: Does this function even make sense to include? Only really makes sense if we allow only one active credential manager.

lukewarlow:
An extra thought, the credential and options objects passed to the extension credential container implementation would also have to include the credentials internal origin property so the extension can correctly bind it to that origin.

Password managers currently have the ability to associate multiple origins to a password credential. So this might have to be considered in the API design.

lukewarlow:
Going to close this for now until there's a bigger consensus on the issue.

PR #370, opened 2023-03-30
Initial Firefox schemas, 2023Q1

zombie:
Still confirming with legal if the licensing section is correct, so this won't merge immediately.


dotproto:
@zombie to update the PR to reflect current schemas.

dotproto (approved):

dotproto:
@zombie, are we still blocked on landing this? If not, I've approved it so we can proceed with the merge.

PR #420, opened 2023-07-11
Add an API proposal for browser.readingList()

dlbjames:
## Description
Adds an API proposal for the ReadingList feature in the Chromium browser. Covers the background of the problem, and the initial solution / schema for the API.

Note: The Chromium ReadingList uses the URL of the entries as the key, meaning no duplicate URLs are allowed. This may need to be addressed browser to browser depending on how the feature is implemented.

## Links
Tracking Issue: [1421058](https://bugs.chromium.org/p/chromium/issues/detail?id=1421058)
Feature Request: [1238372](https://bugs.chromium.org/p/chromium/issues/detail?id=1238372)
[Chromium API Proposal Doc](https://docs.google.com/document/d/1f1wW2955nY5Rp3vfS0JGuo1kACOAmsjZupvLYknWKII/edit)

### DIFF
--- a/proposals/reading_list.md
+++ b/proposals/reading_list.md
@@ -1,4 +1,5 @@
 # New API: browser.readingList()
+Contributors: dljames@chromium.org
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


dotproto:
@dlbjames For clarity, what is the status of this API in Chromium? Is Chromium open to discussion and/or iteration on this API design?

It looks like the relevant Chromium tracking issue is [1421058](https://crbug.com/1421058) and the relevant feature request is [1238372](https://crbug.com/1238372. This is the relevant [Chromium API proposal doc](https://docs.google.com/document/d/1f1wW2955nY5Rp3vfS0JGuo1kACOAmsjZupvLYknWKII/edit).

dlbjames:
> @dlbjames For clarity, what is the status of this API in Chromium? Is Chromium open to discussion and/or iteration on this API design?

We are very early in the implementation of this API. As such, we are open to some discussion / iteration on the design during implementation. However, after initial implementation, we will be more open to discussion of the design provided the Chromium reading list can support the changes / the discussions do not fundamentally change the feature.

I hope the initial version will suffice for the time being (after some comments / light modifications).

> It looks like the relevant Chromium tracking issue is [1421058](https://crbug.com/1421058) and the relevant feature request is [1238372](https://crbug.com/1238372). This is the relevant [Chromium API proposal doc](https://docs.google.com/document/d/1f1wW2955nY5Rp3vfS0JGuo1kACOAmsjZupvLYknWKII/edit).

Yes, this is all correct! I will add these to the initial comment.

Thank you!

proposals/reading_list.md
```diff
@@ -0,0 +1,178 @@
...
+# API Proposal
+
+The `browser.readingList` will introduce a few new types and functions. Below is the schema in the IDL format:
+
+## Types
+A ReadingListEntry is the extensions representation of a Reading List object. Where every URL stored in the Reading List must be unique.
+```js
+ dictionary ReadingListEntry {
+	// The url of the entry.
+	DOMString url;
```
oliverdunk, line 33: Could we say a bit more about what we mean by URL here? Can it have query parameters? Can you have two entries with the same hostname and path but a different hash (for example)?

bershanskiy:
1. Are reading list entries keyed by the URL internally?
     What happens if extension actually wants to update the URL itself? For example, `browser.bookmarks` API provides a unique bookmark `id` for this exact purpose. [Source](https://developer.chrome.com/docs/extensions/reference/bookmarks/#type-BookmarkTreeNode)
 2. Can permission `readingList` be optional?
     The [Chromium API Proposal Doc](https://docs.google.com/document/d/1f1wW2955nY5Rp3vfS0JGuo1kACOAmsjZupvLYknWKII/edit) page 5 mentions permission `readingList` but does not mention whether or not it can be an optional permission.
 3. Is there a way to restrict extension's access to only a subset of reading list (e.g., to have less elevated permissions and hence fewer permission warnings)?
      For example, is there a way to restrict extensions only to reading list entries which extension created? (E.g., "weekly tech news reading list" extension might not need access to all reading list entries if it only wants to add and delete its own entries and might achieve a greater install rate if it was deemed non-intrusive by the privacy-conscious "tech" crowd).

dpenning:
> Are reading list entries keyed by the URL internally?

Yes they are keyed by URL, and a unique URL can only be part of the reading list once, it can't be added multiple times.

> Can permission readingList be optional?

Yes this permission will be optional.

> Is there a way to restrict extension's access to only a subset of reading list (e.g., to have less elevated permissions and hence fewer permission warnings)?

There is currently no way for the reading list entries to be marked as added by a specific extension / action.

proposals/reading_list.md
dlbjames: Oh I see now! 

Yes, the URL can have query parameters. Two entries with the same hostname and path but different hashes are also allowed. As long as the string of the url is different even by a single letter. I will update the comment to reflect this. 

Another thing is that the URL must be "valid". For chromium, this means `chrome://` paths cannot be added in addition to any URLs that do not have protocols or hostnames.

dlbjames: Thank you!

yankovichv:
Hooray, we at Manganum have been waiting for this API for a long time and will implement it as soon as possible.

The document and the API itself looks simple and clear. We wait.

dotproto (approved):
Looks good to merge

xeenon (approved):

yankovichv:
Colleagues, is there any news about when this API may get into the beta and stable channels?

dlbjames:
Oops, apologies for the very delayed response, yankovichv! 

Short answer: Included in beta some time next week!

Since we are targeting chrome version 120, beta promotion (the version that should have the reading list api implemented) will release November 1st. Stable will release on December 5th. 

More information about the release schedules can be found here: https://chromiumdash.appspot.com/schedule

Additionally, in the future for shorter wait times between what we implement and when you can use it, the Canary and Dev versions of chrome receive these updates much faster (I believe Canary is updated 2x a day, and Dev is once a week). 

👍

Issue #433, opened 2023-08-02
Inconsistency: Alarms API time slipping


bershanskiy:
## TL;DR

Alarms API alarm timing is not guaranteed when devices go to sleep.

## Details

The documentation on Alarms API implies/suggests that the alarms internally uses absolute timestamps, fire nearly exactly at specified timestamps, and are independent from one another. In reality, the situation is a bit different:
 - internally Chrome and Firefox convert absolute timestamps to time deltas (from the time internal alarm object is created) and schedule alarms to fire
 - internally Chrome and Firefox do not track time contiguously, instead using something like "ticks" and the timers stop counting when device is inactive. The alarms are "slipping" by exact amount the device is slipping
 - internally Chrome (and, I believe, Firefox) use a single internal alarm object for all scheduled alarms and this internal alarm object tracks the earliest scheduled alarm (at the time when this alarm fires, a new closest alarm is found and scheduled)

This gives a rise to a few curious behaviors:
 - if device goes to sleep, the current internal "next" alarm gets delayed; when internal alarm fires, it is possible that a few other alarms  already past their time and they fire immediately
 - all extensions share a single internal alarm, so extensions can trigger each other's slipped alarms unexpectedly
 - creating an alarm scheduled in the past triggers the recalculation of the "next" internal alarm and thus resets the slippage

patrickkettner:
Hi @bershanskiy!
Are you intending to file a bug? Or are you wanting looking to standardize certain behaviors?

bershanskiy:
> Are you intending to file a bug?

There is a bug filed (I need to find it).

> Or are you wanting looking to standardize certain behaviors?

Ideally, yes, there should be a standard reasonably simple behavior for all browsers. Unfortunately, extensions may have adopted to the current (in my opinion slightly odd) behavior and changing current implementation details might introduce some incompatibilities.

xeenon:
Related to #406 and #422.

xeenon:
Safari calculate alarms by always using wall clock no matter what type of alarm. And repeating alarms are off that original first fire wall clock, not chained from when it fires.

bershanskiy:
Some more Chromium-specific context is available here: https://github.com/darkreader/darkreader/pull/11307#issuecomment-1601488291

[...]

Issue #432, opened 2023-08-01
Agenda discussion for public meeting on 2023-08-03 


dotproto:
Please comment on this issue to add topics for our August 3, 2023 meeting. The agenda will be finalized approximately 12-24 hours before the actual meeting and migrated to our [meeting notes doc](https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit).

## New topics

- [Issue #431](https://github.com/w3c/webextensions/issues/431): Proposal: Default to Active Tab in scripting.executeScript() [@erosman]

## Follow up

- [Issue #362](https://github.com/w3c/webextensions/issues/362): Proposal: Declarative Cosmetic Rules [@ameshkov]
- [Issue #403](https://github.com/w3c/webextensions/issues/403): Proposal: Ability to insert CSS from content script context [@ameshkov]
- [Issue #394](https://github.com/w3c/webextensions/issues/394): [DNR] Add support for wildcards in initiatorDomains and excludedInitiatorDomains fields [@ameshkov]
- [Issue #344](https://github.com/w3c/webextensions/issues/344): Regular expressions in DNR API (regexFilter) [@ameshkov]

ameshkov:
Hi Simeon,

I'd like to touch upon a few existing issue and see if there're anything new there.

1. First of all #362 (declarative cosmetic rules), we have not yet received much feedback on the proposal. I'd like to discuss what needs to be done to move it forward.
2. I think it makes sense to discuss the @gorhill's proposal #403 (insertCSS from a content script) as there's new feedback from Chrome.
3. #394 (wildcards in initiatorDomains). It seems that at the proposal causes some confusion and I'd like to discuss the alternative approach: https://github.com/w3c/webextensions/issues/394#issuecomment-1661719140 
4. Then there's #344 (regexFilter inconsistencies and lack of documentation).

dotproto:
Closing in prep for the meeting.

Issue #431, opened 2023-07-25
Proposal: Default to Active Tab in scripting.executeScript()


erosman:
## Preamble

> **Note:** When using Manifest V3 or higher, use [scripting.executeScript()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript) to execute scripts.
>
> — [tabs.executeScript()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/executeScript)

`scripting.executeScript()` is introduced as the replacement for `tabs.executeScript()`, however:

- TabId in [tabs.executeScript()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/executeScript) is optional and defaults to the active tab of the current window
- TabId in [scripting.executeScript()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript) is mandatory with no default

## Proposal

Make TabId optional and default to the active tab in `scripting.executeScript()` like `tabs.executeScript()`

## Purpose

Many extensions only require `"activeTab"` permission and if necessary, `tabs.executeScript()` can be executed without any additional process or permission in MV2. 

Aforementioned extensions in MV3 `scripting.executeScript()` would be required to include the active tab ID, e.g. via [tabs.query()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/query) ~which would additionally require `"tabs"` permission~.

A common example is [onCommand](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/commands/onCommand) API which does not include any tab information.

tophf:
Also suggested in https://crbug.com/1194519.

P.S. Note that there's no need for the `tabs` permission to get the tab id. The problem with the current design is only the lack of consistency with MV2 and the need to write boilerplate code.

dotproto:
Nit: "active tab" is a little overloaded due to it's close association with the `activeTab` permission. I'd suggest using "current tab" in order to better reflect the intended meaning.

@oliverdunk, can you speak to the rationale behind making tab selection explicit when calling `scripting.executeScript()` in Chrome?

rdcronin:
Chrome's opposed to this.

Determining the currently-active tab is fraught with complexity:
[...]

Issue #403, opened 2023-06-04
Proposal: Ability to insert CSS from content script context


gorhill:
## Background

I have been working on a MV3-compatible _lite_ version of uBlock Origin, named uBlock Origin Lite, or uBOL. The goal with the extension is to be fully declarative, such that no service worker is needed for the filtering to occur. The service worker is required only when the user interacts with the extension settings (either from the popup panel or options page.

The fully declarative nature of the extension is fulfilled, but there is currently a negative side effect as a result: the inability to reliably enforce cosmetic filtering on web pages.

This issue arises because the CSS styles to inject in a web page do not have precedence over the webpage's own CSS styles. This CSS styles precedence issue is normally resolved using the `scripting.insertCSS()`, but since uBOL is entirely declarative, it can't use this approach, as this would require the content script to send a message to the service worker, causing the service worker to wake up as a result. This would defeat the goal of being entirely declarative to avoid the cost of having to constantly wake up the service worker each time a webpage navigation occurs.

## Solution

This issue would be solved if it was possible to call `scripting.insertCSS()` (or equivalent) from the content script context. The API in the context of a content script could be made simpler, as its purpose would be always to inject CSS targeting the current document (so no need for tab id, frame id, etc.)

    chrome.scripting.insertCSS(
      injection: CSSContentScriptInjection
    )

`injection`: The details of the styles to insert.

#### `CSSContentScriptInjection`

`css`: string: A string containing the CSS to inject.

`origin?`: [`StyleOrigin`](https://developer.chrome.com/docs/extensions/reference/scripting/#type-StyleOrigin). The style origin for the injection. Defaults to `AUTHOR`.

## Notes

There are already a set of API methods available from content script contexts: <https://developer.chrome.com/docs/extensions/mv3/content_scripts/#capabilities>, so this is not a new approach.

Using the browser dev tools, I notice that there is already an instance of `chrome.scripting` object in the context of the content script (possibly as a result of the content script being injected through [scripting.registerContentScripts](https://developer.chrome.com/docs/extensions/reference/scripting/#method-registerContentScripts)), so it would be a matter of adding `insertCSS` method to it.

Essentially the API method would accomplish in a more efficient way what can be already accomplished by sending a message to the service worker, and security-wise, I cannot see a difference.

[...]

PR #437, opened 2023-08-03
Publish minutes of 2023-08-03 meeting

Rob--W:

During this meeting we discussed or mentioned issues #432, #431, #433, #362, #403, #394.

### DIFF
--- a/_minutes/2023-08-03-wecg.md
+++ b/_minutes/2023-08-03-wecg.md
@@ -1,0 +1,123 @@
+# WECG Meetings 2023, Public Notes, Aug 3
+
+ * Chair: Timothy Hatcher
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=64caee00,384
+Call-in details: [WebExtensions CG, 3rd August 2023](https://www.w3.org/events/meetings/51a7d2ff-e00d-462b-9071-73b93e78b053/20230803T080000/)
+Zoom issues? Ping @zombie (Tomislav Jovanovic) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+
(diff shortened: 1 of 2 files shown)


_minutes/2023-08-03-wecg.md
```diff
@@ -0,0 +1,123 @@
...
+ * [oliver] I seem to remember similar changes in other APIs, to make the target of script execution explicit.
+ * [timothy] I see the reasoning there; at the same time it makes the API usage cumbersome for common uses. Perhaps we can achieve the explicitness by supporting a new property instead of tabId?
+ * [simeon] This brings up a broader question around API design philosophy. The change to require a tab ID feels in like with the Extensible Web Manifesto, where lower level and more specific capabilities are provided and the expectation is that the community will produce libraries that provide nicer to use APIs. In our case, though, we don't have a demonstrable history of extension-specific libs being widely adopted. A question I still don't have resolved in my own head is “How batteries-included should extension APIs be?”
+ * [timothy] I'm of the mindset of that we should build the APIs the way they are intended to be used.
+ * [oliver] If we were to add a property such as “useCurrentTab”, do we expect people to understand the risks?
+ * [simeon] We see it in action.onClicked, where people use tabs.query even though the tab is provided in the callback.
+ * [oliver]
```
xeenon: @oliverdunk Do you remember what you said here?

oliverdunk: Not for definite, but use this 😄 

```suggestion
 * [oliver] I think that may be further motivation to make this explicit, so developers eventually figure out the callback has what they need. As a side note - some APIs already have a [current window](https://developer.chrome.com/docs/extensions/reference/windows/#the-current-window) concept which tries to use the executing context and may be applicable here.
```

xeenon (approved):

Rob--W:
That was a quick turn-around today, thanks :)

Issue #440, opened 2023-08-16
Proposal: add dictionary operations to DNR modifyHeaders


carlosjeurissen:
### Problem
Currently when attempting to change headers with a dictionary structure one can only set or remove the header. For many use cases like https://github.com/w3c/webextensions/issues/169 and https://crbug.com/1254637, this is problematic. This proposal is there to make sure `dnr` can replace as many `webRequest` usecases.

In addition, if multiple extensions attempt to modify headers conflicts are likely to occur.

### Proposal
In addition to the 'append', 'set' and 'remove' operations, we can introduce a dictionary operation with a set of key-value transformations. The `value` property will be an Array with headerDictionaryAction as such:

```json
"responseHeaders": [{
  "header": "Set-Cookie",
  "operation": "dictionary",
  "value": [{
    "key": "SameSite",
    "operation": "set",
    "value": "Strict"
    }, {
    "key": "Secure",
    "operation": "set",
    "value": true
  }]
}]
```

The operations available for headerDictionaryAction are 'set', 'append', 'remove' and 'removeValue'.
```json
"responseHeaders": [{
  "header": "Permissions-Policy",
  "operation": "dictionary",
  "value": [{
    "key": "geolocation",
    "operation": "removeValue",
    "value": "self"
  }, {
    "key": "geolocation",
    "operation": "append",
    "value": "self"
  }, {
    "key": "geolocation",
    "operation": "remove"
  }]
}]
```

Since headers using dictionaries have different syntaxes (for example, `Content-Security-Policy`, `Permissions-Policy`, `Set-Cookie`), this is something we have to consider.

The Content-Security-Policy header can have multiple dictionaries. Which makes this especially interesting. My proposal would be to have the operations be applied to each dictionary individually. Keep in mind CSP values with multiple dictionaries are very common, but common enough we need to handle them.

Say we have a `Content-Security-Policy` header with the following value:
`default-src 'none'; img-src 'self'; script-src 'self'; frame-src 'self', default-src 'self'; media-src 'none', default-src 'none'; connect-src 'self'`

The dictionaries are separated by commas. So we deal with three CSPs here, being:
`default-src 'none'; img-src 'self'; script-src 'self'; frame-src 'none'`
`default-src 'self'`
`default-src 'none'; connect-src 'self'`

With below operations:
```json
"responseHeaders": [{
  "header": "Content-Security-Policy",
  "operation": "dictionary",
  "value": [{
    "key": "img-src",
    "operation": "removeValue",
[...]

Issue #439, opened 2023-08-16
Proposal: Add filter for DNR `modifyHeaders`


carlosjeurissen:
### Problem
Currently conditionally modifying headers is generally only possible with `webRequest`. This proposal is there to make sure `dnr` can replace as many `webRequest` usecases.

The proposal would fix: https://crbug.com/1254637 and address some of the use cases described in https://github.com/w3c/webextensions/issues/169

### Proposal
The proposal is to add a `regexFilter` property to `ModifyHeaderInfo` which only apply the operation if and only if the regex matches.

```json
"action": {
  "type": "modifyHeaders",
  "responseHeaders": [{
    "operation": "set",
    "header": "set-cookie",
    "regexFilter": ";\\s*SameSite=(Lax|Strict)|$",
    "value": "; SameSite=None; Secure"
  }]
},
```

In addition, a `valueFilter` could be introduced to do a simple string match for performance reasons. As proposed by @tophf, both `valueFliter` and `regexFilter` could be specified, in which case `valueFilter` is tested first. If it fails, no regexMatch is required thus improving performance.

```json
"action": {
  "type": "modifyHeaders",
  "responseHeaders": [{
    "operation": "set",
    "header": "Set-Cookie",
    "valueFilter": "SameSite=Lax",
    "value": "; SameSite=None; Secure"
  }]
},
```

Since some operations require a global match (see use case 2). So we need a way to specify them. Something like this could work:
```json
"action": {
  "type": "modifyHeaders",
  "responseHeaders": [{
    "operation": "set",
    "header": "Content-Security-Policy",
    "regexFilter": "(^|,\\s?)",
    "regexFlags": "g",
    "value": "$1img-src 'self'; "
  }]
},
```

### Use case 1, `Set-Cookie`  SameSite flag
See https://bugs.chromium.org/p/chromium/issues/detail?id=1254637
Set SameSite flag of `Set-Cookie` header to 'None' if it originally was 'Lax' or 'Strict'.

#### Blocking WebRequest
```js
function responseListener (details) {
  const responseHeaders = details.responseHeaders;
  responseHeaders.forEach((header) => {=
    if (header.name === 'Set-Cookie'){
        header.value = header.value.replace(/;\s*SameSite=(?:Lax|Strict)|$/, '; SameSite=None; Secure');
    }
  });
  return { responseHeaders: responseHeaders };
}

chrome.webRequest.onHeadersReceived.addListener(
  responseListener,
[...]

Issue #436, opened 2023-08-03
Handling web pages in a freeform environment


carlosjeurissen:
### Introduction
Browser extensions leverage existing web features for their functionality. This can sometimes be challenging when it is used for something which was not initially designed for this. Take serviceworkers for example, which had to adapt a bit to accommodate browser extensions as well.

The same issue we are having with webpage sizing. In browser extensions, the size of certain pages is mostly defined by the extension page itself. This happens namely in the extension popup and option pages. Since there is no clear way of how this should be handled, browsers handle it differently.

### List of inconsistencies
After exploring some situations in which browsers handle them differently, I have created a few demo extensions showcasing the differences in different browsers. Examples are different handling to a lot of content, to overflow:hidden; with scrollbars, and to how body and html tags work together. See: https://github.com/carlosjeurissen/webext-tech-demo-extensions/tree/main/issues/crossbrowser-freeform

This is only a very small collection. There is much more which is different. Think about `min-width`, `max-width` and lets not talk about media queries.

There are also situations in which a page has to handle multiple situations. For example, the options page could in some browsers be opened in a normal tab. Meaning the extension needs to handle both freeform and classic page situations.

### Main goal of extension developers
Since it is good to know why this is even an issue. Currently it is very hard for extension developers to create something which works and is handles consistently in different situations (browsers, options dialog, full option page). Not going into detail about how this should be approached, ideally an extension developer would be able to figure out if it is in a freeform situation (either horizontally or vertically or both), and sets its preferred size.

We could potentially also collaborate with other w3c communities on this one. Also a potential issue to talk about during [TPAC](https://github.com/w3c/webextensions/issues/385).

carlosjeurissen:
This repository: https://github.com/carlosjeurissen/webext-tech-demo-extensions/tree/main/issues/crossbrowser-freeform contains many examples of inconsistencies. However, to make it a bit easier, here is a 1:1 comparison of some of the inconsistencies in Firefox, Chrome and Safari Desktop. Additional screenshots are available in the specific repositories.

### [1. Inconsistency with much content](https://github.com/carlosjeurissen/webext-tech-demo-extensions/blob/main/issues/crossbrowser-freeform/crossbrowser-much-content/)

![image](https://github.com/carlosjeurissen/webext-tech-demo-extensions/blob/main/issues/crossbrowser-freeform/crossbrowser-much-content/screens/comparison.png?raw=true)

[...]

Issue #435, opened 2023-08-03
Agenda discussion for public meeting on 2023-08-17


dotproto:
Please comment on this issue to add topics for our August 17, 2023 meeting. The agenda will be finalized approximately 12 hours before the actual meeting and migrated to our [meeting notes doc](https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit).

## Carry over

The following topics were on the agenda for our 2023-07-03 meeting, but we did not have enough time to discuss them.

- [Issue 362](https://github.com/w3c/webextensions/issues/362): Declarative cosmetic rules
- [Issue 344](https://github.com/w3c/webextensions/issues/344): Regular expressions in DNR API (regexFilter)
- [Issue 154](https://github.com/w3c/webextensions/issues/154): Proposal: browser.secureStorage API

## New issues

- [Issue 436](https://github.com/w3c/webextensions/issues/436): Freeform inconsistencies [@carlosjeurissen]
- [PR 434](https://github.com/w3c/webextensions/pull/434): Create persistence-of-states.md (part of [issue 424](https://github.com/w3c/webextensions/issues/424))

## Ongoing issues

* [PR 420](https://github.com/w3c/webextensions/pull/420): browser.readingList
* [Issue 319](https://github.com/w3c/webextensions/issues/319): Increase the maximum number of dynamic rules to 30,000
* [Issue 433](https://github.com/w3c/webextensions/issues/433): Inconsistency: Alarms API time slipping [@bershanskiy]
* [Issue 169](https://github.com/w3c/webextensions/issues/169): Blocking webRequest usecase - add, modify and remove CSP directives [@carlosjeurissen]
* [Issue 439](https://github.com/w3c/webextensions/issues/439): Proposal: Add filter for DNR modifyHeaders [@carlosjeurissen]
* [Issue 440](https://github.com/w3c/webextensions/issues/440): Proposal: add dictionary operations to DNR modifyHeaders [@carlosjeurissen]

carlosjeurissen:
@dotproto Thanks for creating this. Would like to nominate https://github.com/w3c/webextensions/issues/436 It is something brought up in the past yet described in more detail. This is something which we could collaborate with different w3c groups and thus could be interesting bringing up during TPAC.

bershanskiy:
I'll add the relevant details to #433 and record a demo of the issue so that we can discuss it.

[...]

PR #442, opened 2023-08-17
Publish minutes of 2023-08-17 meeting

Rob--W:

During this meeting we discussed or mentioned issues #435, #154, #436, #424, #319, #433, #169, #439, #440, and PRs #434, #421, #420.

### DIFF
--- a/_minutes/2023-08-17-wecg.md
+++ b/_minutes/2023-08-17-wecg.md
@@ -1,0 +1,161 @@
+# WECG Meetings 2023, Public Notes, Aug 17
+
+ * Chair: Simeon Vincent
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=64dd6300,384
+Call-in details: [WebExtensions CG, 17th August 2023](https://www.w3.org/events/meetings/51a7d2ff-e00d-462b-9071-73b93e78b053/20230817T080000/)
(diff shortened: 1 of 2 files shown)


_minutes/2023-08-17-wecg.md
```diff
@@ -0,0 +1,161 @@
...
+
+ * [carlos] (see issue) Goal is to reduce inconsistencies of freeform pages.
+ * [rob] Web pages in general or extension UI pages?
+ * [carlos] Extension UI pages (e.g. options UI pages); It could be helpful to define explicitly how these should be handled.
+ * [simeon] A comparison table describing the differences could make it easier to identify actionable parts for the browser.
+ * [carlos] The issue links to a repo with screenshots.
+ * [rob] An explicit comparison with suggestion of desired behavior would make it easier to discuss.
+ * [timothy] Think that this is a good area to focus on. E.g. on WebKit we trigger a special layout mode for autosizing.
+ * [rob] I'm supportive of converging towards something common when it's clear that we should be heading towards one behavior.
+ * [oliver] ...
```
Rob--W: @oliverdunk If you recall what you said, could you add it? I recall you expressing some support while also raising concerns, but not the exact message.

oliverdunk: Here you go:
```suggestion
 * [oliver] In places where the browsers already have similar behaviour, I think it makes sense to align. The one caveat I would call out is that part of our WECG charter allows browsers to differentiate - so we can never rule out a browser deciding to have significantly different rendering.
```

PR #484, opened 2023-11-08
Update table with persistence of APIs

oliverdunk:
Adds some additional data to the table based on experiments I had run a while back.


hanguokai:
@oliverdunk Thanks for updating. I only tested `action.setBadgeText()`. In Chrome, it is B+D in MV2, and A in MV3.

I'm not too familiar with Firefox and Safari, especially Safari. It would be better to let @Rob--W and @xeenon review it.

oliverdunk:
> I only tested `action.setBadgeText()`. In Chrome, it is B+D in MV2, and A in MV3.

Thanks for checking. I just tested again and you're right, pushed an update.

dotproto (approved):

xeenon (approved):

PR #499, opened 2023-11-28
action.onUserSettingsChanged proposal

EmiliaPaz:
Proposal based on w3c#346: New API: UserSettings.isOnToolbar change event

### DIFF
--- a/proposals/action-on-user-settings-changed-api.md
+++ b/proposals/action-on-user-settings-changed-api.md
@@ -31,8 +31,13 @@ dictionary UserSettings {
 #### Events
 
 ```javascript
-// Fire when user-specified settings relating to an extension's action changed.
-<browser>.action.onUserSettingsChanged(
+// Fires when user-specified settings relating to an extension's action changed.
+<browser>.action.onUserSettingsChanged.addListener(
+  callback: function, // where callback looks like: (userSettings: UserSettings) => void
+);
+
+// Deregisters the event listener.
+<browser>.action.onUserSettingsChanged.removeListener(
   callback: function, // where callback looks like: (userSettings: UserSettings) => void
 );
 ```
--- a/proposals/action-on-user-settings-changed-api.md
+++ b/proposals/action-on-user-settings-changed-api.md
@@ -1,0 +1,63 @@
+# Extension API Proposal: action.onUserSettingsChanged
+
+## Background
+
+Proposal based on [issue #346](https://github.com/w3c/webextensions/issues/346): New API: UserSettings.isOnToolbar change event
+
+## API Overview
+
+### Objective
+
+Fire an event when user-specified settings relating to an extension's action changed.
+
+### Use Cases
+
+#### Pinning / unpinning extension
+
+Extensions can know if its action is pinned by calling `chrome.action.getUserSettings.` If the extension wants to know when this value changes, it has to poll the API method. Instead, it could register for an event.
+
+## Implementation
+
+### API Schema
+
+#### Types
+
+```javascript
+dictionary UserSettings {
+  isOnToolbar: boolean
+}
+```
+
+#### Events
+
+```javascript
+// Fire when user-specified settings relating to an extension's action changed.
+<browser>.action.onUserSettingsChanged(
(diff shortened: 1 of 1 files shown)


bershanskiy:
This API permits addition of new observers, but not removal of old ones: `action.onUserSettingsChanged()`.

Instead, it could match the established pattern of `<eventName>.addListener()` and `<eventName>.removeListener()` and become `action.onUserSettingsChanged.addListener()` and `action.onUserSettingsChanged.removeListener()`.

oliverdunk:
> Instead, it could match the established pattern of `<eventName>.addListener()` and `<eventName>.removeListener()`

Thanks for the callout. I'll make sure @emilia-paz sees this but I suspect it is just a mistake in the proposal. We would almost certainly want to follow the normal event listener pattern here :)

oliverdunk:
@zombie / @dotproto / @xeenon - could you take a look at this?

proposals/action-on-user-settings-changed-api.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Objective
+
+Fire an event when user-specified settings relating to an extension's action changed.
+
+### Use Cases
+
+#### Pinning / unpinning extension
+
+Extensions can know if its action is pinned by calling `chrome.action.getUserSettings.` If the extension wants to know when this value changes, it has to poll the API method. Instead, it could register for an event.
```
Rob--W: Nit: move `.` (end of sentence) outside of the code block (delimited by backticks).

```suggestion
Extensions can know if its action is pinned by calling `chrome.action.getUserSettings`. If the extension wants to know when this value changes, it has to poll the API method. Instead, it could register for an event.
```

proposals/action-on-user-settings-changed-api.md
```diff
@@ -0,0 +1,68 @@
...
+  isOnToolbar: boolean
+}
+```
+
+#### Events
+
+```javascript
+// Fires when user-specified settings relating to an extension's action changed.
+<browser>.action.onUserSettingsChanged.addListener(
+  callback: function, // where callback looks like: (userSettings: UserSettings) => void
```
Rob--W: The type of the event parameter is identical to the return value of `getUserSettings`. Is that an intentional design?

With the current proposal consisting of one property only, it does not make that much of a difference. Once multiple events get added, the answer depends on the design decision:

- If we want to always include the latest value of all settings, then we can use the same type as the event parameter to `onUserSettingsChanged` listeners and as the return value of `getUserSettings`.
- If we want to only include modified values, then we need a new type that looks like `onUserSettingsChanged`, but with all fields optional.

proposals/action-on-user-settings-changed-api.md
```diff
@@ -0,0 +1,68 @@
...
+
+### New Permissions
+
+| Permission Added | Proposed Warning |
+| ---------------- | ------------------------------------------------------- |
+| N/A              | This is an addition to the action API, which is exposed if the extension includes the "action" key in the manifest. |
+
+### Design and Implementation
+
+Currently, `UserSettings` only has one property. This makes things easy, as we just need to trigger the event when such property changes. If new properties are added, we will trigger the event every time any of the properties change.
```
Rob--W: Could you explicitly call out whether the event will receive an object with the modified values, or an object with the same value as `getUserSettings`? See my other comment at the API signature for more context.

Rob--W (commented):
This looks good. Just one question, which is mainly relevant for the future implementation (and current documentation).

proposals/action-on-user-settings-changed-api.md
EmiliaPaz: Done

proposals/action-on-user-settings-changed-api.md
EmiliaPaz: Great catch! We think it's more valuable to the user to only return the modified values. Updated the proposal

proposals/action-on-user-settings-changed-api.md
EmiliaPaz: Done.

Rob--W (approved):

PR #508, opened 2023-12-14
create specification for self.browser

patrickkettner:
Create specification for window.browser/self.browser.

preview available [here](https://patrickkettner.github.io/window.browser/)

### DIFF
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -48,13 +48,33 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       <a attribute lt="browser"><code>window.browser</code></a> API
     </h3>
 
-  {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
-  it MUST be used only for WebExtension related functionality.
+    {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
+    it MUST be used only for WebExtension related functionality.
   </section>
 
 
-  <pre class="idl">
+<pre class="idl">
   partial interface Window {
    attribute object browser;
+  };
+
+</pre>
+
+  <section>
+    <h3 id="worker-interface">
+      Worker API
+    </h3>
+
+    When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
+    It MUST be used exclusivly for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
+  </section>
+
+<pre class="idl">
+  partial interface DedicatedWorkerGlobalScope {
+    attribute object browser;
+  };
+
+  partial interface ServiceWorkerGlobalScope {
+    attribute object browser;
   };
 </pre>
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -65,7 +65,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       Worker API
     </h3>
 
-    When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
+    When {{browser}} is defined on {{window}}, it SHOULD also be exposed on {{ServiceWorkerGlobalScope}}
+    of origins associated with WebExtensions.
     It MUST be used exclusively for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
   </section>
 
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -33,7 +33,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   Chromium, other vendors used the more neutral <code>browser</code> namespace.
   That <code>browser</code> is now considered the canonically correct namespace
   for all WebExtension APIs.  Some interactions with WebExtensions (such as
-  external messaging from content scripts to background processes) necessitate
+  external messaging from websites to background processes) necessitate
   the exposure of this namespace to the web.
 
   Given that, this document defines <code>window.browser</code> as reserved for
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -24,6 +24,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   ## Introduction {#introduction}
   <em>This section is non-normative.</em>
 
+  Browser extensions are programs to extend the capabilities of web browsers.
+  Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
+  
   Modern WebExtensions are an evolution of a system originally introduced in
   Chrome version 3. In that system, any API that was added was put under the
   <code>chrome</code> namespace. When WebExtensions were adopted outside of
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -1,0 +1,57 @@
+<pre class="metadata">
+Title: window.browser
+Shortname: wecg-browser
+Level: 1
+Group: wecg
+Status: UD
+URL: https://patrickkettner.github.io/window.browser
+Editor: Patrick Kettner, Google, patrickkettner@google.com
+Abstract: This specification reserves the <code>window.browser</code> namespace for use by WebExtensions.
+Repository: patrickkettner/window.browser
+Complain About: accidental-2119 on, missing-example-ids on
+</pre>
+
+<pre class=link-defaults>
+spec:html; type:attribute; text:window
+</pre>
+
+<pre class="anchors">
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
+    type: dfn; text: valid non-empty URL; url: valid-non-empty-url-potentially-surrounded-by-spaces
+</pre>
+
+<section class="non-normative">
+  ## Introduction {#introduction}
+  <em>This section is non-normative.</em>
+
+  Modern WebExtensions are an evolution of a system originally introduced in
+  Chrome version 3. In that system, any API that was added was put under the
+  <code>chrome</code> namespace. When WebExtensions were adopted outside of
+  Chromium, other vendors used the more neutral <code>browser</code> namespace.
+  That <code>browser</code> is now considered to canonically correct namespace
+  for all WebExtension APIs.  Some interactions with WebExtensions (such as
+  external messaging from content scripts to background processes) necessitate
+  the exposure of this namespace to the web.
+
+  Given that, this document defines <code>window.browser</code> as reserved for
+  the use of APIs designed to interact with WebExtensions. The specifics of what
+  APIs within <code>window.browser</code> are at present intentionally undefined,
+  and are left up to User Agent implementation.
+</section>
+
+
+  <section>
+    <h3 id="window-interface">
+      <a attribute lt="browser"><code>window.browser</code></a> API
+    </h3>
+
+  {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
+  it MUST be used only for WebExtension related functionality.
+  </section>
+
+
+  <pre class="idl">
+  partial interface Window {
+    readonly attribute object browser;
+  };
+</pre>
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -28,7 +28,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
 
   Modern WebExtensions are an evolution of a system originally introduced in
-  Chrome version 3. In that system, any API that was added was put under the
+  Chromium. In that system, any API that was added was put under the
   <code>chrome</code> namespace. When WebExtensions were adopted outside of
   Chromium, other vendors used the more neutral <code>browser</code> namespace.
   That <code>browser</code> is now considered the canonically correct namespace
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -52,6 +52,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 
   <pre class="idl">
   partial interface Window {
-    readonly attribute object browser;
+   attribute object browser;
   };
 </pre>
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -66,7 +66,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     </h3>
 
     When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
-    It MUST be used exclusivly for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
+    It MUST be used exclusively for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
   </section>
 
 <pre class="idl">
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -70,10 +70,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   </section>
 
 <pre class="idl">
-  partial interface DedicatedWorkerGlobalScope {
-    attribute object browser;
-  };
-
   partial interface ServiceWorkerGlobalScope {
     attribute object browser;
   };
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -21,7 +21,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 </pre>
 
 <section class="non-normative">
-  ## Introduction {#introduction}
+  ## Introduction
   <em>This section is non-normative.</em>
 
   Browser extensions are programs to extend the capabilities of web browsers.
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -4,10 +4,10 @@ Shortname: wecg-browser
 Level: 1
 Group: wecg
 Status: UD
-URL: https://patrickkettner.github.io/window.browser
+URL: https://w3c.github.io/webextensions/specification/window.browser.html
 Editor: Patrick Kettner, Google, patrickkettner@google.com
 Abstract: This specification reserves the <code>window.browser</code> namespace for use by WebExtensions.
-Repository: patrickkettner/window.browser
+Repository: w3c/webextensions
 Complain About: accidental-2119 on, missing-example-ids on
 </pre>
 
@@ -26,7 +26,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
 
   Browser extensions are programs to extend the capabilities of web browsers.
   Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
-  
+
   Modern WebExtensions are an evolution of a system originally introduced in
   Chrome version 3. In that system, any API that was added was put under the
   <code>chrome</code> namespace. When WebExtensions were adopted outside of
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -31,7 +31,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   Chrome version 3. In that system, any API that was added was put under the
   <code>chrome</code> namespace. When WebExtensions were adopted outside of
   Chromium, other vendors used the more neutral <code>browser</code> namespace.
-  That <code>browser</code> is now considered to canonically correct namespace
+  That <code>browser</code> is now considered the canonically correct namespace
   for all WebExtension APIs.  Some interactions with WebExtensions (such as
   external messaging from content scripts to background processes) necessitate
   the exposure of this namespace to the web.
--- a/specification/window.browser.html
+++ b/specification/window.browser.html
@@ -6,8 +6,8 @@
   <meta content="UD" name="w3c-status">
   <link href="https://www.w3.org/StyleSheets/TR/2021/W3C-UD" rel="stylesheet">
   <meta content="Bikeshed version fe5461827, updated Mon Dec 18 10:46:21 2023 -0800" name="generator">
-  <link href="https://patrickkettner.github.io/window.browser" rel="canonical">
-  <meta content="8e082534c5a28f252dae33ac7618b7145b1b02fb" name="revision">
+  <link href="https://w3c.github.io/webextensions/specification/window.browser.html" rel="canonical">
+  <meta content="0c9789208a9c689ccc89ab3a2e6e1d055ddde021" name="revision">
 <style>/* Boilerplate: style-autolinks */
 .css.css, .property.property, .descriptor.descriptor {
     color: var(--a-normal-text);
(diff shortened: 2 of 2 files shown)


patrickkettner:
@Rob--W @xeenon would yall be able to review when time permits?

specification/window.browser.html
```diff
@@ -0,0 +1,1673 @@
...
+    <p><code class="idl"><a data-link-type="idl" href="#dom-window-browser" id="ref-for-dom-window-browser">browser</a></code> is UA defined attribute exposed on <code class="idl"><a data-link-type="idl" href="https://fetch.spec.whatwg.org/#dom-requestinit-window" id="ref-for-dom-requestinit-window">window</a></code>. When implemented,
+  it MUST be used only for WebExtension related functionality.</p>
+   </section>
+<pre class="idl highlight def"><c- b>partial</c-> <c- b>interface</c-> <a class="idl-code" data-link-type="interface" href="https://html.spec.whatwg.org/multipage/nav-history-apis.html#window" id="ref-for-window"><c- g>Window</c-></a> {
+  <c- b>readonly</c-> <c- b>attribute</c-> <a class="idl-code" data-link-type="interface" href="https://webidl.spec.whatwg.org/#idl-short" id="ref-for-idl-short"><c- b>short</c-></a> <dfn class="dfn-paneled idl-code" data-dfn-for="Window" data-dfn-type="attribute" data-export data-readonly data-type="short" id="dom-window-browser"><code><c- g>browser</c-></code></dfn>;
```
xeenon: `short` is the wrong type for this. `object` would be the best type to use here until we have a `Namespace` object type spec.

specification/window.browser.html
```diff
@@ -0,0 +1,1673 @@
...
+  Some interactions with WebExtensions (such as external messaging from content
+  scripts to background processes) necessitate the exposure of this namespace
+  to the web.
+  Given that, this document defines `window.browser` as reserved for the use of
+  APIs designed to interact with WebExtensions. The specifics of what
+  APIs within `window.browser` are at present intentionally undefined, and are
+  left up to User Agent implementation.</p>
+   </section>
+   <section>
+    <h3 class="heading settled" id="window-interface"><span class="content"> <a class="idl-code" data-link-type="attribute" href="https://compat.spec.whatwg.org/#dom-window-orientation" id="ref-for-dom-window-orientation"><code>window.browser</code></a> API </span></h3>
```
xeenon: Not sure why `orientation` is referenced here.

specification/window.browser.html
```diff
@@ -0,0 +1,1673 @@
...
+    <p><code class="idl"><a data-link-type="idl" href="#dom-window-browser" id="ref-for-dom-window-browser">browser</a></code> is UA defined attribute exposed on <code class="idl"><a data-link-type="idl" href="https://fetch.spec.whatwg.org/#dom-requestinit-window" id="ref-for-dom-requestinit-window">window</a></code>. When implemented,
```
xeenon: `window` should should likely reference `https://html.spec.whatwg.org/#window` instead of the Fetch spec.

xeenon (changes requested):
Looks good! Some references need corrected to point in the right places.

Also, do you have the original bike shed file? We have Bikeshed transformation setup in our repo, so you should be able to just post that file, like [`index.bs`](https://github.com/w3c/webextensions/blob/main/index.bs) which is more human friendly. That gets published [here](https://w3c.github.io/webextensions/specification).

patrickkettner:
thanks @xeenon!
I added the bikeshed file. The current build system is using [pr-preview](https://github.com/tobie/pr-preview), and [configured](https://github.com/w3c/webextensions/blob/ff053a84ac68eb6e6d8f8df3704eeb7f6631f23c/.pr-preview.json#L2C1-L2C1) to build the [existing spec](https://w3c.github.io/webextensions/specification/). pr-preview has a [long running request](https://github.com/tobie/pr-preview/issues/18) to support multiple files, but the existing PR has been [untouched for years](https://github.com/tobie/pr-preview/pull/95). @dotproto and I discussed replacing the build system with something that supports multiple files built directly on bikeshed, rather than using pr-preview. I plan on digging into it soon. In the meantime, given devlin's desire to give a few months notice to developer before any change landing, I didn't want to block on that.

specification/window.browser.html
patrickkettner: removed. It was vestigial code from the compat spec, thanks for pointing it out

specification/window.browser.html
patrickkettner: updated

specification/window.browser.bs
```diff
@@ -0,0 +1,57 @@
+<pre class="metadata">
+Title: window.browser
+Shortname: wecg-browser
+Level: 1
+Group: wecg
+Status: UD
+URL: https://patrickkettner.github.io/window.browser
```
Rob--W: I suppose that the personal Github repositories in the metadata is only temporarily, for preview purposes, right? That is, before merging, this URL and other references will point to the WECG repo.

specification/window.browser.bs
```diff
@@ -0,0 +1,57 @@
...
+    </h3>
+
+  {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
+  it MUST be used only for WebExtension related functionality.
+  </section>
+
+
+  <pre class="idl">
+  partial interface Window {
+    readonly attribute object browser;
```
Rob--W:
```suggestion
   attribute object browser;
```

Suggestion to remove `readonly`, for the following reasons:

- These properties are currently not read-only, at least not in Chrome, nor in Firefox.
- Making it writable reduces the webcompat risk. Using `readonly` breaks websites that assign to `browser`.
- Making it writable also makes it easier to replace the `browser` namespace in case one wants to replace it altogether to support unit tests.

A counterpoint, in favor of adding `readonly`, is that it enables web pages to rely on the presence of an unmodified `browser`, unaffected by other browser extensions or third-party libraries, for example. I am not attributing much value to this feature, because one cannot expect any protections once third-party code has executed in the context.

specification/window.browser.bs
```diff
@@ -0,0 +1,57 @@
...
+  and are left up to User Agent implementation.
+</section>
+
+
+  <section>
+    <h3 id="window-interface">
+      <a attribute lt="browser"><code>window.browser</code></a> API
+    </h3>
+
+  {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
```
Rob--W: How about extension service workers?

Do we want to reserve `browser` in regular workers as well?

specification/window.browser.bs
```diff
@@ -0,0 +1,57 @@
...
+<pre class="anchors">
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
+    type: dfn; text: valid non-empty URL; url: valid-non-empty-url-potentially-surrounded-by-spaces
+</pre>
+
+<section class="non-normative">
+  ## Introduction {#introduction}
+  <em>This section is non-normative.</em>
+
+  Modern WebExtensions are an evolution of a system originally introduced in
```
Rob--W, line 30: Question: Is this sufficient, or do we need a very brief introduction to what "WebExtensions" is, to people unfamiliar with WebExtensions?

Originally, it was just "Chrome Extensions". When Mozilla adopted the same model, the name "WebExtensions" was coined, described at https://wiki.mozilla.org/Add-ons/Terminology. Although that page asserts that "WebExtensions" ought to refer to the technology, and not the concrete extensions built with it, in practice the term is being used to refer to both.

Here is a small suggestion to add some context, and to also call out why there is a need for a "browser" namespace at all.

```suggestion
  Browser extensions are programs to extend the capabilities of web browsers.
  Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
  
  Modern WebExtensions are an evolution of a system originally introduced in
```

specification/window.browser.bs
patrickkettner: yes, that is correct

specification/window.browser.bs
Rob--W: Although my suggested edit has been approved, I'd like to temporarily re-open this thread to allow other vendors to chime in. In particular, I'd like to hear @xeenon 's perspective on Safari's behavior. I see that Safari's interface of the `browser` namespace content itself is defined at https://github.com/WebKit/WebKit/blob/ea5e30e46342b0e755281a1ffb2f23c37a2b9bcd/Source/WebKit/WebProcess/Extensions/Interfaces/WebExtensionAPINamespace.idl , with widespread use of "readonly", but cannot tell from the WebKit source code whether the `browser` global is supposed to be `readonly` or writable.

xeenon: window.browser is not readonly in WebKit / Safari.

specification/window.browser.bs
patrickkettner: added

specification/window.browser.bs
patrickkettner: it is also not readonly in Chrome. I would be surprised if there was interest in changing it.

specification/window.browser.bs
patrickkettner: updated to the presumed future location

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+  <em>This section is non-normative.</em>
+
+  Browser extensions are programs to extend the capabilities of web browsers.
+  Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
+
+  Modern WebExtensions are an evolution of a system originally introduced in
+  Chrome version 3. In that system, any API that was added was put under the
+  <code>chrome</code> namespace. When WebExtensions were adopted outside of
+  Chromium, other vendors used the more neutral <code>browser</code> namespace.
+  That <code>browser</code> is now considered to canonically correct namespace
```
oliverdunk:
```suggestion
  That <code>browser</code> is now considered the canonically correct namespace
```

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+  for all WebExtension APIs.  Some interactions with WebExtensions (such as
+  external messaging from content scripts to background processes) necessitate
```
oliverdunk: This can happen from a content script or just a website itself so I think avoiding the use of an extension-specific term makes this clearer to outside readers.

```suggestion
  external messaging from websites to background processes) necessitate
```

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+    </h3>
+
+    {{browser}} is UA defined attribute exposed on {{window}}. When implemented,
+    it MUST be used only for WebExtension related functionality.
+  </section>
+
+
+<pre class="idl">
+  partial interface Window {
+   attribute object browser;
```
oliverdunk, line 58: Should we make it clear that this is sometimes undefined, such as if a site is not included in the `externally_connectable` array in any extension's manifest? I'm guessing the right syntax based on [this](https://hg.mozilla.org/integration/autoland/file/d65760500d063d80f6cc3c1fa8c61bfe1ac81412/dom/bindings/test/TestExampleGen.webidl#l582) but it seems to be a new feature that doesn't appear in many real specs so not sure if there's a better way. cc/ @Rob--W @xeenon 

```suggestion
   attribute (undefined or object) browser;
```

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+
+</pre>
+
+  <section>
+    <h3 id="worker-interface">
+      Worker API
+    </h3>
+
+    When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
+    It MUST be used exclusivly for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
```
oliverdunk:
```suggestion
    It MUST be used exclusively for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
```

oliverdunk (changes requested):
Thanks so much for putting this together, I'm excited! Left a couple of thoughts but this generally looks good.

specification/window.browser.bs
patrickkettner: it feels weird to set the only attribute being specified to undefined in the spec. Certainly can add it, but I am not sure I see the value if it is user agent defined. Would love to hear Rob's and Timothy's thoughts

Jack-Works:
are you considering renaming this PR (or part of the non-normative sections) to `self.browser`? that covers service workers.

specification/window.browser.bs
dotproto: Since WebIDL treats `object` and `undefined` as distinct types, it seems like this would more accurately represent the possible values a user agent may assign the global we're reserving. 

Would it be more appropriate (or conventional) to use a union type as @oliverdunk suggested, a nullable type, or possibly even to declare a (partial) namespace?

```webidl
partial interface Window {
 attribute object? browser;
};
```

```webidl
[Exposed=(Window,Worker)]
partial namespace browser {}
```

I tried looking around at other specs to see if/when/how `namespace` was used and was surprised at how rarely it appeared. The only notable place I saw it used was the [console spec](https://console.spec.whatwg.org/#console-namespace).

oliverdunk: As a note, my (uneducated) reading of the spec was that nullable types were equivalent to `null | something` so I'm not sure they'd be appropriate here. Still very unsure what the right path is though.

xeenon (approved):

dotproto (approved):
LGTM

specification/window.browser.bs
oliverdunk: We've agreed to revisit this after landing.

patrickkettner: @oliverdunk can you LGTM?

oliverdunk (approved):
LGTM 👍

patrickkettner:
@Rob--W any thoughts?

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+spec:html; type:attribute; text:window
+</pre>
+
+<pre class="anchors">
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
+    type: dfn; text: valid non-empty URL; url: valid-non-empty-url-potentially-surrounded-by-spaces
+</pre>
+
+<section class="non-normative">
+  ## Introduction {#introduction}
```
Rob--W: This renders as is, with `{#introduction}`. What was the intended result?

specification/window.browser.html
```diff
@@ -0,0 +1,1704 @@
...
+};
+
+</pre>
+   <section>
+    <h3 class="heading settled" id="worker-interface"><span class="content"> Worker API </span></h3>
+    <p>When <code class="idl"><a data-link-type="idl" href="#dom-window-browser" id="ref-for-dom-window-browser">browser</a></code> is defined on <code class="idl"><a data-link-type="idl" href="https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-window" id="ref-for-dom-window①">window</a></code>, it SHOULD also be implemented on <code class="idl"><a data-link-type="idl" href="https://w3c.github.io/ServiceWorker/#serviceworkerglobalscope" id="ref-for-serviceworkerglobalscope">ServiceWorkerGlobalScope</a></code> and <code class="idl"><a data-link-type="idl" href="https://html.spec.whatwg.org/multipage/workers.html#dedicatedworkerglobalscope" id="ref-for-dedicatedworkerglobalscope">DedicatedWorkerGlobalScope</a></code>.
+    It MUST be used exclusivly for WebExtension purposes, however the contents of each instance of <code class="idl"><a data-link-type="idl" href="#dom-dedicatedworkerglobalscope-browser" id="ref-for-dom-dedicatedworkerglobalscope-browser①">browser</a></code> is UA defined.</p>
```
Rob--W: I looked at the preview, which features this "exclusivly" typo, which is not present in the .bs file. Can you re-generate the html file to make sure that it matches the .bs file?

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+
+    When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
+    It MUST be used exclusively for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
+  </section>
+
+<pre class="idl">
+  partial interface DedicatedWorkerGlobalScope {
+    attribute object browser;
+  };
+
```
Rob--W:
```suggestion
```

AFAIK there are no browser implementations that expose `chrome`/`browser` to the global scope of `Worker`s, not even those created from an extension origin. Therefore I suggest to remove the mention of `DedicatedWorkerGlobalScope`.

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+</pre>
+
+  <section>
+    <h3 id="worker-interface">
+      Worker API
+    </h3>
+
+    When {{browser}} is defined on {{window}}, it SHOULD also be implemented on {{ServiceWorkerGlobalScope}} and {{DedicatedWorkerGlobalScope}}.
```
Rob--W:
```suggestion
    When {{browser}} is defined on {{window}}, it SHOULD also be exposed on {{ServiceWorkerGlobalScope}}
    of origins associated with WebExtensions.
```

Suggested edit (dropped the dedicated worker as explained in another comment), and emphasized that the origin of the scope must be a WebExtension origin. Also changed "implemented" to "exposed", because "implemented" seems more vague.

Currently we don't have a separate definition for extension workers, but if we ever do (as suggested in #11), then that would be the type to extend instead.

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+    It MUST be used exclusively for WebExtension purposes, however the contents of each instance of {{browser}} is UA defined.
```
Rob--W: Question: The mention of Service Workers here may raise questions, because Service workers are defined for http(s) URLs only: https://w3c.github.io/ServiceWorker/#start-register-algorithm
(in practice mainly https, http is only supported for edge cases such as localhost).

Chrome (and Firefox's experimental implementation) does not permit extension code to register arbitrary service workers. Instead, the extension framework takes care of registering workers despite the http(s) restrictions.

Should we clarify in a note that extensions cannot register Service workers as usual (i.e. the https restriction), but that browsers may spawn a Service worker associated with a WebExtension origin?

Should we drop the mention of workers entirely from this "window.browser" document, and define the worker behavior in an entirely separate spec/PR?

specification/window.browser.bs
```diff
@@ -0,0 +1,80 @@
...
+
+<section class="non-normative">
+  ## Introduction {#introduction}
+  <em>This section is non-normative.</em>
+
+  Browser extensions are programs to extend the capabilities of web browsers.
+  Although not standardized at first, web browsers have independently implemented a common interoperable set of extension APIs, called WebExtensions.
+
+  Modern WebExtensions are an evolution of a system originally introduced in
+  Chrome version 3. In that system, any API that was added was put under the
```
Rob--W: Is it that relevant to mention "version 3" after "Chrome"? I suggest to drop it so that there is no confusion with "Manifest version 3".

(also rename Chrome to Chromium because the rest of the paragraph refers back to it as Chromium.)

```suggestion
  Chromium. In that system, any API that was added was put under the
```

Rob--W:
I know that we wanted us to spec the existence of `browser` because Chrome considers a spec as a prerequisite to exposing this. The specific behavior of "browser" is left up to the UA, so my question here is not a blocking part of the review. Still, I'm curious about the ideas behind it.

Chrome (and Safari) both have the concept of "isolated world" for content scripts. An isolated world shares the same view of the DOM (DOM instances from native code are shared), with unique JavaScript wrappers (isolated per world). If `window.browser` were to be defined in the idl as part of `window`, what does that say about the value read by the main world vs isolated world? Is it the same (native) object, but with different wrappers that somehow filters the available extension API methods? Or is the `browser` namespace a different object (and therefore implicitly also a separate wrapper if any)?

Here is some background on isolated world and a description of bindings: https://chromium.googlesource.com/chromium/src.git/+/123.0.6266.0/third_party/blink/renderer/bindings/core/v8/V8BindingDesign.md

In Firefox, externally_connectable has not been implemented yet ([bug 1319168](https://bugzilla.mozilla.org/show_bug.cgi?id=1319168)), which means that there is no `browser` namespace in the main world, only in that of content scripts. The current implementation of the sandbox in Firefox is that `globalThis` is a sandbox inheriting from `window`. `globalThis.browser` is the extension API namespace, `window.browser` is currently undefined. I imagine that if we were to implement `window.browser`, that the main world version would be generated through WebIDL, and that the content script continues to have its own implementation-defined extension API namespace separate from `window.browser`.

xeenon:
The Safari and WebKit implementations of web extensions currently dynamically add the `browser` object as needed to the DOM global object, per world. So it is always a different wrapper _and_ a different internal object.

It needed to be different since the things contained in the `browser` namespace for the main world in web pages is drastically reduced and different from the extension content script world or main world in extensions pages.

We will likely continue to dynamically add this property to the global scope, instead of adding it in our DOM window IDL.

I don't think that fundamentally blocks anything here. Browser implementations can expose the APIs however their implementation requires, as long as it matches how the IDL in the spec defines it.

specification/window.browser.bs
patrickkettner: ah, thanks for pointing that out. That is the syntax for declaring a custom ID in markdown. But bikeshed doens't support it. The correct thing to do would be to define an ID on the section on the line above. But since I added it to be able to link to it later in the doc, and that link was factored out, I will just remove this.

specification/window.browser.bs
patrickkettner: Thanks for the feedback @Rob--W!

> because Service workers are defined for http(s) URLs only: https://w3c.github.io/ServiceWorker/#start-register-algorithm (in practice mainly https, http is only supported for edge cases such as localhost).

I am not sure I follow. The spec in this PR says that if `window.browser` is defined, then `self.browser` should be defined on any `ServiceWorkerGlobalScope` whose `origin` is a webextension. While the service worker spec is only http(s), the thing that is registered as a background script in Chrome for manifest v3 _is_ 
 `ServiceWorkerGlobalScope`. 

> Chrome (and Firefox's experimental implementation) does not permit extension code to register arbitrary service workers. Instead, the extension framework takes care of registering workers despite the http(s) restrictions.

If they are not permitted, then I don't see the problem? ServiceWorkerGlobalScope is what is created when you have an extension in Chrome/Firefox's-experimental-implementation. Can you explain what you are wanting changed?

> Should we clarify in a note that extensions cannot register Service workers as usual (i.e. the https restriction)

While I see the use for that in a broader web extensions specification, as there is no other portion of this document is about registering a service worker, or when it should be registered on `window`. That seems outside of this document's scope.

> Should we drop the mention of workers entirely from this "window.browser" document, and define the worker behavior in an entirely separate spec/PR?

Given the current level of web extension specification, having two essentially identical documents to accomplish the same thing ("`globalThis.browser` should only be used for web extensions") feels unnecessary.

specification/window.browser.html
patrickkettner: done

specification/window.browser.bs
Rob--W: > Thanks for the feedback @Rob--W!
> 
> > because Service workers are defined for http(s) URLs only: https://w3c.github.io/ServiceWorker/#start-register-algorithm (in practice mainly https, http is only supported for edge cases such as localhost).
> 
> I am not sure I follow. The spec in this PR says that if `window.browser` is defined, then `self.browser` should be defined on any `ServiceWorkerGlobalScope` whose `origin` is a webextension. While the service worker spec is only http(s), the thing that is registered as a background script in Chrome for manifest v3 _is_ 
>  `ServiceWorkerGlobalScope`. 

In the realm of specifications, Service Workers (and thus ServiceWorkerGlobalScope instances) only exist on https schemes (and localhost). Here we specify that ServiceWorkerGlobalScope should have a browser object, for WebExtension origins only. I initially found this inconsistent, because the SW spec doesn't leave room for SW on other schemes, while our spec here implies the existence of an extension origin.

I have now reconciled this inconsistency by observing that "extension origin" could theoretically also involve one of the accepted schemes (even though that doesn't happen in practice - as an example from a far past, Chrome had the concept of [Hosted apps](https://web.archive.org/web/20191221070723/https://developer.chrome.com/webstore/hosted_apps) where regular websites received extra privileges if a crx file defining the "hosted app" was installed).
We (browser vendors) understand the intent and practical meaning here, and I am fine with not mentioning anything more about SW for now.

Rob--W (approved):
LGTM! I can't easily verify the preview, but the source looks good!

Issue #454, opened 2023-09-19
Add a new manifest key to include a origin trial token


patrickkettner:
Chrome and Firefox both have the concept of an [Origin Trial](https://developer.chrome.com/en/docs/web-platform/origin-trials/). Normally, these are registered via either an HTTP header, or a meta tag. For background pages, this can be done via a meta tag. But for background service workers, or the different worlds of content scripts, this is not currently able to be done. 

The Chrome team is supportive of this, but we are still figuring out prioritization of the work. [Here is the implementation tracking bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1484767)

Is anyone _against_ the concept?

bershanskiy:
A new manifest key with origin token(s) for Origin/Deprecation Trials does sound very reasonable, especially since it would allow introduction of new APIs without fear of ossification and deprecation of old behaviors (e.g., edge cases) without fear of breakage. Do you happen to have a concrete syntax in mind yet?

hanguokai:
> But for background service workers, or **the different worlds of content scripts**, this is not currently able to be done.

1. Do you mean that extension's origin trial token can be applied to content scripts? That is out of extension's origin.

2. Other Chromium-based browsers also support OT except Edge which has it's own OT and Extension Store?

3. OT usually lasts for a few months. Once it expires, will the extension be removed from Extension Store if it doesn't release a new version?

patrickkettner:
@bershanskiy 

_> Do you happen to have a concrete syntax in mind yet?_

nothing written yet, however I assume it would be something like 

```json
  "trial_token": "AxujKG9INjsZ8/gUq8+dTruNvk7RjZQ1IFhhgQbcTJKDnZfbzSTE81wvC2Hzaf3TW4avA76LTZEMdiedF1vIbA4AAABueyJvcmlnaW4iOiJodHRwczovL2ltYXNkay5nb29nbGVhcGlzLmNvbTo0NDMiLCJmZWF0dXJlIjoid293WW91QXJlRGVkaWNhdGVkVG9UaGlzIiwiZXhwaXJ5IjoxNjk1NjU4MjQ5LCJpc1RoaXJkUGFydHkiOnRydWV9"
```

inside of the manifest.json.


@hanguokai 
_> Do you mean that extension's origin trial token can be applied to content scripts? That is out of extension's origin._

To be investigated - if the MAIN world has an active origin trial, I was not able to confirm that isolated worlds inherit it.



[...]

PR #529, opened 2024-01-24
Add permissions.requestSiteAccess() API proposal

EmiliaPaz:
Proposal for a new API method to allow the extension to show site access requests at runtime without any user gesture in a less-obtrusive way than with `permissions.request()`


proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+// optional_host_permissions, a content script match pattern, or an applicable
+// activeTab site).
+// Only one of tabId or documentId must be declared. Request will be
+// automatically reset on navigation.
+<browser>.permissions.requestSiteAccess(
+  // The id of the tab where site access requests can be shown.
+  number?: tabId,
+  // The id of a document where site access requests can be shown.
+  string?: documentId,
+  callback?: function
```
xeenon: Does this also return a `Promise`? Does Chrome allow callbacks in v3? (Safari allows either promises or callbacks.)

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Abuse Scenarios
+
+Extensions can enable site access requests on every tab, which could be annoying to the user.
+
+#### Mitigations
+Attacks of annoyance are not in our threat model. Extensions can have annoying behaviors, but they risk uninstallation.
+On the browser side, we can place some restrictions to limit noisiness:
+- User can decide whether they want to see site access requests for an specific extension
+- We can explore setting a max amount of times the request is shown when the user ignores it.
```
xeenon, line 141: I think leaving all this up to the browser is fine, and does not need any explicit mitigations in the API. As long as the `Promise` can be rejected to signal the request was denied, or a `bool` result in the Promise.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Additional Security Considerations
+None
+
+## Alternatives
+
+### Existing Workarounds
+
+Currently, extensions have site access by default. Users can change the site access to be when clicked. The only way for extensions to signal the user they want access to the site is through permissions.request() flow which is noisy and requires a user gesture. Otherwise, they rely on the user realizing through the browser UI (e.g extensions menu).
```
xeenon: > Currently, extensions have site access by default. Users can change the site access to be when clicked.

This is browser specific, and not true for all browsers. In Safari extension don't have any site access by default.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+Extensions have no way to signal the user that they want access to the site, and rely on the user realizing through the browser UI (e.g extensions menu)
+
+### Other Alternatives Considered
+
+Specifying URL patterns instead of tabId or documentId in `permissions.requestSiteAccess`. We decided against that because:
+- This is designed to be a highly-contextual signal -- the extension should do it only if they have strong believe they will provide value to the user on the given page
+- We do not want extensions to simply show a request on every page, and specifying a list of patterns would lend itself to that behavior
```
xeenon: Patterns vs no patterns does not prevent the extension from calling this API for every tab / page. So I'm not sure this is true.

But I agree, taking a `tabId` / `documentId` is best (other than my caveat about how does the extension know what tab it needs without access.)

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Applies to a specific tab or document id
+- Doesn’t need to be made inside the handler for a user action
+- Shows the request in the UI, handled differently by each browser. See more in [UI Elements and User-Visible Effects](#ui-elements-and-user-visible\-effects) section
+- When accepted, grants always access to the site’s origin
+- Resets the request on cross-origin navigation
+
+#### Use Cases
+
+An extension requested access to a site but the user withheld its access and forgot about it. Extension wants to signal it needs site access to execute. For example, “shopping” extension wants to show site access when user navigates to “amazon.com” and access was withheld.
```
xeenon: If an extension does not have access, how does it know what `tabId` is needed to pass to this API? They can't see the `url` or `title` on the tab object if they don't have access (at least in Safari, where those are empty unless you have host permissions. That might be different in Chrome where only the `tabs` permission is needed?).

xeenon (commented):
Thanks for doing this!

proposals/permissions-requestSiteAccess-api.md
xeenon: Should the `Promise` / `callback` be passed a `bool`? What happens if the request is denied?

fregante:
At what point would an extension want to use this? As @xeenon pointed out, how do I know which tab to target if I don't know their URL? 

Is this API only useful when the user is already interacting with the extension via sidePanel? That's the only scenario I can think of where the user doesn't have activeTab but the extension would probably want access.

Or maybe is this API meant to preserve access to a tab with a temporary activeTab permission?

I worked on a related UI before ([webext-permission-toggle](https://github.com/fregante/webext-permission-toggle)) so I wonder how I could use this.

proposals/permissions-requestSiteAccess-api.md
oliverdunk: Oh, that's an interesting question. In Chrome the `tabs` permission is sufficient to access the URL. In Safari, I wonder if there are other situations where the extension realises it might be useful on a page - based on a shortcut running perhaps? - and the extension could request access. I know Safari has a lot of behaviour to automatically prompt the user though so perhaps not.

oliverdunk: Out of interest, does the `tabs` permission do anything in Safari? Despite the name it isn't needed to call (I think all) tabs APIs in Chrome.

xeenon: @oliverdunk Yes, `tabs` does signal to us that the extension implicitly wants access to all websites, so they don't need `*://*/*` in the `host_permissions` to get our implicit permission request dialogs when doing `tabs` actions like `query()`, `executeScript()`, etc.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+Allow the extension to show site access requests at runtime without any user gesture in a less-obtrusive way than with permissions.request(). This can be done with a new API that:
+
+- Applies to a specific tab or document id
+- Doesn’t need to be made inside the handler for a user action
+- Shows the request in the UI, handled differently by each browser. See more in [UI Elements and User-Visible Effects](#ui-elements-and-user-visible\-effects) section
+- When accepted, grants always access to the site’s origin
+- Resets the request on cross-origin navigation
+
+#### Use Cases
```
hanguokai, line 27: Overall, this is a confusing API to developers, and it would be better if there were more concrete use cases. What value can this API bring to extensions? Is there any relevant data for browsers' "site access" function? Are users satisfied with "site access" function, or are users encountering any problems? The above questions are just to avoid developing an API that no one uses.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Existing Workarounds
+
+Currently, extensions have site access by default. Users can change the site access to be when clicked. The only way for extensions to signal the user they want access to the site is through permissions.request() flow which is noisy and requires a user gesture. Otherwise, they rely on the user realizing through the browser UI (e.g extensions menu).
+
+Extensions have no way to signal the user that they want access to the site, and rely on the user realizing through the browser UI (e.g extensions menu)
+
+### Other Alternatives Considered
+
+Specifying URL patterns instead of tabId or documentId in `permissions.requestSiteAccess`. We decided against that because:
```
hanguokai: URL pattern is a reasonable way to tell the browser which websites an extension can work for. And extensions don't need to get Tab's URL. We should not deny a reasonable approach. The browser can decide how to prompt the user with this hint(URL pattern). BTW, a specific host permission (not too broad pattern like <all_urls>) is already a hint.

proposals/permissions-requestSiteAccess-api.md
oliverdunk: I can share a bit of extra context here. We are still looking at possible permission UI changes like the ones mentioned at [I/O](https://youtu.be/QYd2XiUYNlE?t=608). This doesn't include changing any defaults, but we do still expect more users to experiment with different access settings. It feels like putting APIs in place first is better than adding them retrospectively, which would imply a gap between when the UI launches and when APIs to adapt to them are available.

fregante: This feels like it belongs in a manifest rather than as an API. I think extensions fall into these categories:

- on demand: just take an action when invoked (`activeTab`)
- for specific websites: think "Reddit Enhancement Suite" or "Refined GitHub", that have a pre-set list of domains to run on
- for all websites: think content blockers, Grammarly

Between these there are a couple more cases:

- A: pre-set list, but more can be added
- B: all websites, but some can be disabled

"A" can be achieved via `.request()` when the user desires; "B" is covered by Chrome's current UI.

I'm not sure where `requestSiteAccess` falls on this spectrum.

Once I specify `optional_host_permissions: all_urls`, the browser should already:

- Show "This extension is available on this website"
- Allow toggling the permission

In your proposed new UI, it looks like this could be covered already: just show the extension on the list, but disabled.

---

As a user, I wish that I had a way to enable the type A on more websites in a consistent way.

As a developer, I wish I didn't have to manually handle [content script registration](https://github.com/fregante/webext-dynamic-content-scripts) and [permission requests](https://github.com/fregante/webext-permission-toggle).

oliverdunk: I definitely agree that not having to manually handle this would be nice and think that'd be interesting to experiment with. For some of your cases I agree that might work.

I do think there is another case to the ones you mentioned. Consider a shopping extension that can only offer coupons on certain sites, and it downloads this list remotely. It specifies `<all_urls>`, so can run everywhere, but a user explicitly revokes its access from all pages. It would be nice if the extension had a way to less intrusively flag "I checked my list and I can run here, you might want to add me back on this domain".

fregante: >  It would be nice if the extension had a way to less intrusively flag "I checked my list and I can run here, you might want to add me back on this domain".

Why isn't the list in the manifest? 

But still from a manifest standpoint, what is the point of `host_permissions` if they can be disabled? It sounds like we need:

- compatible host list: all_urls
- suggested host list: eBay, amazon, etc

To fit into this proposal:

- compatible host list: `host_permissions`
- suggested host lists: a new `preferred_hosts`

The drawback is that extension developers have to submit a new extension to update the list, but the advantages are that the browser has total control on how/when to show it, and the developer doesn't have to deal with issues like "when and how often can I `requestSiteAccess`?"

xeenon: As-is the current proposed API would not be useful in Safari for shopping extensions, since the URLs are hidden from extensions without host permissions. A declarative model might be the better approach.

With declarative approach, shopping like extensions to register patterns and reasons / descriptions to say "I'm interested in shopping.example/* because I have '2 coupons for 10%-20% off'". This would give a signal the browser can bubble up in the UI to the user in a privacy preserving way, that has a call to action for why the user would want to grant access.

hanguokai: In addition to what others have already mentioned, host permissions are not always bound to Tabs. For example, an extension needs a host permission, just to bypass CORS limitation in the background page. User never opens a tab for that host.

And host permissions are not always 1-to-1 relationship to Tab's URL. It may also be a 1-to-many or many-to-many relationship. For example, an extension needs two host permissions, only granting one by tab's host still can't work.

In short, host permissions are more complex and can't be reduced to the Tab's host the user is opening. If users want to manage host permissions fine-grained manually, existing functionality (in the chrome://extensions/?id=extension-id) already supports this.

xeenon: Users often don't want to manage the host permissions though, and will just grant access to all sites to avoid any hassle. We should strive to make the defaults be privacy preserving and not require host permission micro-management. So new APIs should be considered that improves things for the user and developers.

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: > Does this also return a Promise? 

Yes, this should return a promise. In Chrome, promises are supported in Manifest V3 and later, but callbacks are provided for backward compatibility. You cannot use both on the same function call. The promise resolves with the same type that is passed to the callback.

> Should the Promise / callback be passed a bool? 

As for the return, there are a couple options:
  a. Returns a boolean whether the request was valid (e.g request will have no effect if extension doesn't have host permissions)
  b. Returns a boolean whether the user accepted the request

I think b. makes sense. One detail is that on cross-origin navigation, the request is effectively not accepted so it would need to return false. Haven't looked at the code, but there should be a way to do so.

Will update the doc once we decide on return value

> What happens if the request is denied?

We should probably add some limitation on cannot re-request on a denied request. Otherwise, it would be very intrusive to the user

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Thanks for the callout, I reworded it to be more browser inclusive

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Thanks everyone for your comments. I'll try to summarize this.

There are two main **issues**:
1. URLs are hidden from extensions without host permissions
2. Host permissions are not always bound to tab

There is one main **suggestion**:  declarative approach by listing the sites in the manifest

For 1., this is a browser design constraint and something we should talk about in more details.. which will be easier to do in the next WECG meeting

For 2., if I understand this correctly the multiple host permissions can happen when there are multiple iframes with different host permissions. The request would only be for the top origin, same as how `activeTab` works

For the declarative approach suggestion, the main reason we don't want to do that is that we want there to be a deliberate, strong signal that the extension can do something useful on the site and really wants to run.  This should not be a passive, "hey, I think I can do something here", it should be a "hey, you, the user, probably want me to do something here".  

When we talked internally at Chrome about having this be declarative and take in URL patterns (either via a runtime method or a manifest field), it became far too close to host permissions themselves. We suspect that the vast majority of extensions would just have the same field match as in their host permissions, since there is no more knowledge at manifest time about why the extension would run on a specific site.

For example, for the shopping extension it can request <all_urls> host permission (or even a large set of specific site) but only show the request when there is a coupon (e.g it doesn't need to show on store.com but yes on store.com/item)

Yes, this is more work for developer since they need to call it every time. However, we hope this to be better for the extension, it can send requests when it really wants to run, and the user, more likely to accept the requests.

EmiliaPaz: Now answering some specific comments

@fregante 
> In your proposed new UI, it looks like this could be covered already: just show the extension on the list, but disabled.

This is covered in the UI, but wasn't clear. Extensions with withheld host permissions, or extensions with not-yet granted optional host permissions are listed in the extensions menu and display they are requesting access.
The proposal here is giving the extension a way to request site access on a site that doesn't require user action

> Why isn't the list in the manifest?

I believe this was answered in previous comment :)

> The drawback is that extension developers have to submit a new extension to update the list, but the advantages are that the browser has total control on how/when to show it, and the developer doesn't have to deal with issues like "when and how often can I requestSiteAccess?"

With both approaches, the browser would have control on how/when to show it. Similarly, with both approached developer can only request site access if it has host permission for the site. As of how often, it is deliberately left as open ended in case browsers need to mitigate abuse scenarios (which would be properly communicated if implemented)

@xeenon, we can discuss further in WECG about the URL constraint in Safari

@hanguokai, I believe I answered the multiple host permissions on a tab on previous comment

EmiliaPaz: These are good point to bring up. Happy to discuss them further at the next public WECG meeting, which is tomorrow February 1st :)

hanguokai: As oliver said before, my understanding is that the background for the use of this API is that when the extension is installed, it is not granted any host permissions although it declares (non-optional) host permissions in manifest.json.

In addition, I feel you are implying that this API is only for content scripts, not for extension contexts (service worker, extension pages, etc). And the function only runs when the page is visually to the user. If so, this is obviously only some use cases of extensions, and it is not enough to address other cases where extensions do not have permissions after installation. 

> For 2., if I understand this correctly the multiple host permissions can happen when there are multiple iframes with different host permissions. The request would only be for the top origin, same as how activeTab works

Not exactly that case, for example, 
1. The network requests of many large websites are complex. After the page is loaded, it will request data from other hosts. E.g. when "www.site-A.com" is loaded, then it fetches data from "api.site-A.com". In this case, the extension may need "api.site-A.com" permission instead of "www.site-A.com" permission in the service worker for it's function. The user never open "api.site-A.com" in a tab, because this domain is not designed for users.
2. If some core host permissions (not related to a specific shopping website) are not granted, the extension may can't work (i.e. `host_permissions` in manifest.json).

oliverdunk: > In addition, I feel you are implying that this API is only for content scripts, not for extension contexts (service worker, extension pages, etc). And the function only runs when the page is visually to the user.

This is definitely the intent. I understand it doesn't solve background use cases - since this is a very high visibility prompt we want it to only show when linked to an open tab that the user has context for. Improving the background use case is a non-goal but that isn't to say we couldn't look into new APIs in the future if needed.

fregante: I can't reconcile these two statements:

> It specifies `<all_urls>`, so can run everywhere, but a user explicitly revokes its access from all pages

> this API is only for content scripts

Content scripts cannot run if the user revoked a permission. It seems that this API exclusively fits this situation:

1. the extension has permission to execute scripts in a tab (temporarily or permanently)
2. it asks for further host permissions or it asks for permanent permission to the current context

Seems niche, why not just have the extension deal with this? This way they can clearly communicate the intent and reason behind the request.

As a user, if a the browser asks for a permission for a third domain, I don’t think I'd accept without further explanation, especially if I explicitly blocked the extension on that domain before.

oliverdunk: Ah, the wording may have been slightly confusing there. It would often be used to _grant access_ to content scripts. The API itself would definitely be available in background contexts.

An example use case in Chrome would be something like this:

- An extension requests <all_urls> in host_permissions
- The user changes the site access to "on click" or "on specific sites" (note that they can do this even if the host permissions weren't specified as optional)
- In the background, the extension realises it can inject a useful content script (for example, it notices a tabs.onUpdated event and the tab has a URL it has special logic for)
- The extension would then prompt for access to that tab ID

Does that make sense?

greghuc:
I want to make a slightly off-topic comment expanding on what @hanguokai [said earlier about host permissions not being bound to Tabs](https://github.com/w3c/webextensions/pull/529#discussion_r1467950140). Sorry for being off-topic, but I think it's worth increasing awareness on this, and whether it's worth a fix. @oliverdunk or @dotproto - I'd be happy for this to be spun off into a separate issue, but I can only describe the problem and not the solution.

There's a confusing situation where an extension would ideally have only "activeTab" permissions ('When You Click the Extension' in the Chrome UI), but needs broad <all-urls> `host_permissions` at install time ('Can Read / Write Data on All Sites')

The problem happens for an extension that runs on the active-tab when the user clicks the extension button. The extension gains `host_permissions` to the origin in the top frame (the main web page), but _not_ to any other origins used in rendering the page (like the origin of a remote image hosting service). If the content script then tries to fetch and access a resource from one of these other origins (like a remote-hosted image), then it runs into CORS errors. The current solution used is to request broad <all_urls> `host_permissions`, and fetch and access these resources from a background script, which is unaffected by CORS.

I ran into this issue when writing an extension that fetched and processed images on the current webpage. @dotproto and I had a [Chromium Extensions chat about this back in December 2021](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/Xz2ghqB9cwY/). I submitted a Chromium feature request back in July 2022 to [allow CORS-free access to dom image elements](https://bugs.chromium.org/p/chromium/issues/detail?id=1341943). However, I'm now thinking this issue should be handled at the CORS level when fetching a resource.

Maybe "activeTab" permission could be relaxed slightly to allow extension access to _all_ origins used in rendering a page. Or at least GET read access. In the [previous chat](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/Xz2ghqB9cwY/m/zEkJTkzCDQAJ), @dotproto said this wasn't done, as "if a user invokes the extension on a page with, say, social media sharing widgets, it's not clear that the user also wants to grant the extension access to all of those social media sites as well". However, if you're balancing this against the current approach of broad `host_permissions` to resolve this issue, then a relaxed "activeTab" permission doesn't seem so bad.

Sorry for being off-topic @EmiliaPaz - I thought it was worth adding this comment, given we're covering `host_permissions`

oliverdunk:
> I want to make a slightly off-topic comment expanding on what @hanguokai [said earlier about host permissions not being bound to Tabs](https://github.com/w3c/webextensions/pull/529#discussion_r1467950140). Sorry for being off-topic, but I think it's worth increasing awareness on this, and whether it's worth a fix.

Thanks for raising this. I think there is some overlap since this API will also only grant access to the top-level site's origin. The reasoning for this is very similar to everything you mentioned with the behavior of `activeTab` today. There is a high level of risk, especially with "tracking pixels" and social media embeds, that users may be granting more access than they realise.

There is actually some precedence for granting more than the current origin with `activeTab` - I believe Firefox granted access to all frames on a page in MV2. They [aligned](https://bugzilla.mozilla.org/show_bug.cgi?id=1781801) with Chrome and Safari in MV3 to be consistent, though my understanding at least as of a year ago was that they were still interested in supporting the capability.

All of that said - the general opinion last time I discussed this within Chrome (which I am inclined to agree with) is that we should keep looking for a good solution, but automatically granting access to all origins associated with a tab isn't it. It may reduce the use of widely scoped host permissions, but at the cost of making it harder for users to understand what an extension can do. It doesn't feel like it meaningfully improves the situation unfortunately and it would be better to keep thinking about this.

greghuc:
@oliverdunk many thanks for the considered response.

I do wonder if the CORS issue I described is narrower that it looks, and predominately applies to images. If so, the solution to this might be image-specific (like my [chrome.dom.createDomImageBitmap API proposal](https://bugs.chromium.org/p/chromium/issues/detail?id=1341943)), rather than a 'grant wider origin access' solution, and so fix 90+% of issues.

My detailed knowledge of CORS has faded, but if I recollect correctly:
- Images/canvas are in a strange place with CORS, where an image from a foreign origin can be rendered in a webpage, and so seen by the user. But if code accesses the image data, then a CORS error is raised
- But other resource types just fail on cross-origin load?

In which case, if extensions are primarily focussed on processing a rendered webpage and associated loaded resources, then the issue becomes image-specific.

oliverdunk:
> the solution to this might be image-specific (like my [chrome.dom.createDomImageBitmap API proposal](https://bugs.chromium.org/p/chromium/issues/detail?id=1341943)), rather than a 'grant wider origin access' solution, and so fix 90+% of issues.

For your use case it definitely sounds like something narrower in scope might work, though I do share the concerns about doing something cross-origin in nature within a content script. Feel free to ping on the issue and we can keep discussing there 😁

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Let's take a step back and see current behavior (updated proposal `Background` section too)
_Please, feel free to correct me on other browsers behavior since I'm not fully familiar!_

- An extension can request host permissions. At installation, they are granted (Chrome, Firefox) or withheld (Safari) by default
- An extension can request optional host permissions, that can be granted after a user action
- In Chrome and Safari, extension site access can be withheld
- In Chrome, withholding an extension site access means the extension is set to run "on click". When the extension is clicked (either the pinned action or through the extensions menu) it gets *access to run on the origin top frame until cross-origin navigation*.
- `permissions.request()` allows an extension to request site access after a user action

Let's have an example now (similar to Oliver's on previous comment). In Chrome:
1. An extension requests `<all_urls>` in host_permissions
2. User navigates to site.com
3. The user changes the `site.com` site access to "on click"
4. User navigates to `site.com/useful`
5. In the background, the extension realizes it can inject a useful content script on `site.com/useful` 

User can realize the extension wants to run by opening the extensions menu. However, there is no way for the extension to say "I really want to run on this specific page" without user action. Yes, it could work with a declarative list. However, we have listed our reason why we prefer not to (updated proposal's `Other Alternatives Considered` section)

@hanguokai 
As for the issue of extension requesting host_permissions that are not in the tab top frame origin. At least on Chrome, this is the actual behavior of granting site access to an extension. Granting site access to a host (either by activeTab or allowing site access "on site" / "on all sites") grants only to that host, meaning the tab's top frame origin. This new API is following the same, so it wouldn't fix that issue.

@fregante 
> it asks for further host permissions or it asks for permanent permission to the current context

Extension cannot ask further host permissions. It would ask for host_permissions that were withheld by the user

> As a user, if a the browser asks for a permission for a third domain, I don’t think I'd accept without further explanation, especially if I explicitly blocked the extension on that domain before.

Browser cannot ask permission for a third domain

hanguokai: Thanks for your further explanation. Now I have a clearer understanding of the intent of this API. But in my opinion, this is still a controversial API.

## How to use this API

> 5. In the background, the extension realizes it can inject a useful content script on site.com/useful

Can you further explain what developers are supposed to do here? 

Because this API is bound to browser tabs, I think developers need to use `tabs.onCreated/onUpdated/onReplaced` events to detect tab's top frame changes, and must need the "tabs" permission to get `tab.url` and/or `tab.title`. In these events, developers are facing two possible situations: 1. already has the host permission and already inject content scripts 2. don't has the host permission and no injected content scripts. Therefore, developers first need to determine whether has the host permission. If no host permission, at this stage, extensions only know tab's url and title, and **don't know the page's content** since it has not inject a content script. At this point, developers can only rely on these two pieces of information (url and title) to determine whether to apply for the site permission, i.e. call `permissions.requestSiteAccess()`.

Here, without knowing the content of the page, it may be difficult for developers to determine whether it strongly requires the site access. In the case of shopping, I only know the URL and title, I may can't know what product is being sold on the current page, so I can't tell if there is a coupon for that product.

## The model itself

> In Chrome, withholding an extension site access means the extension is set to run "on click". When the extension is clicked (either the pinned action or through the extensions menu) it gets access to run on the origin top frame until cross-origin navigation.

I rarely use this feature. I also rarely hear users use this feature. So I want to repeat the question I asked at the very beginning: Is there any relevant data about this function? Are users satisfied with this function, or are users encountering any problems at present?

I think the model itself is flawed, it only fits a small subset of real use cases, and a lot of use cases (perhaps 50% or more) don't fit the model (We have already given examples before).

## Experiment with this API without standardization

In my opinion, there are still doubts about the actual value of this API for developers and users. I think Chrome can first experiment with this API (for a year) to see the actual effect. Tell developers explicitly that this is an experimental API and that it may be modified incompatibly or even removed in the future. This way, an experimental API does not require the agreement from other browsers. For example, let developers apply an Origin Trial token to experiment this API (see #454).

EmiliaPaz: ### How to use this API

Correct, those are the two scenarios although I would phrase them slightly differently (to match terminology):
a. Extension request host permission and it's granted by the user. Extension has injected scripts
b. Extension request host permission and it's withheld by the user. Extension ha no injected  scripts

On scenario a., you are right. Extension does not know the page content.  Extension has access to:
- `permissions.getAll()` returns the active permissions. Extension can know if a host permission is not in the returned permissions origin
- `tabs.get()` returns tab URL, title, ...
- other APIs like `windows.getCurrent()` to get more information 

> Here, without knowing the content of the page, it may be difficult for developers to determine whether it strongly requires the site access. In the case of shopping, I only know the URL and title, I may can't know what product is being sold on the current page, so I can't tell if there is a coupon for that product.

Currently, the only way for developers to determine whether it requires site access is to check if it has active permission for the page. This provides a way to show a visible request when this happens. 

Yes, it's probably not sufficient to know given a tab id / document id that the extension "strongly needs access to the site". However, what would be another way to determine that? (if host permissions are withheld, extension doesn't have access to the page). By looking at document id / tab id, extension can get more information than if it just gave a declarative list of URLs (as explained in previous point).  

### The model itself

> I rarely use this feature. I also rarely hear users use this feature. 

User withholding host permissions is not a common practice today. However, we expect that to change in the future because:

- Browsers are exploring withholding permissions at installation. Safari already does that, and I believe Firefox is exploring that in MV3. Chrome is exploring that too
- At least in Chrome, it will be more accessible for the user to withhold host permission in the extensions menu (that means, setting the extension to run "on click")

> Is there any relevant data about this function? Are users satisfied with this function, or are users encountering any problems at present?

We did user research on showing a request in the Chrome toolbar every time a extension wants site access to a withhold host permission. Users found it to be too nosy, and not always helpful. However, they mentioned that the request is relevant then it provides value.
Thus, we drafted this proposal to show the request only when the extension requests to show it. Yes, it can be always but at least it gives the option to the extension to have it be more relevant and the browser to better control it. 

> I think the model itself is flawed, it only fits a small subset of real use cases, and a lot of use cases (perhaps 50% or more) don't fit the model (We have already given examples before).

Agree, the real use cases subset is small now. However, we expect this to increase as mentioned on the first point on this section.

### Experiment with this API without standardization

Do other browsers see a potential path forward for this API? Otherwise, Chrome can continue without standardization

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Taking a second look, I think it's better to return a. whether the request is valid. My reasoning is better explained in the proposal (see other alternatives considered). Basically, to have a more clear difference vs showing a permission request and granting a permission (which can be done in various ways).

However, happy to discuss this one further

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Proposal was updated to include url so it can fit other browsers. Browser can add requirements on which parameters must be provided. For example, Chrome will require `tabId` or `documentId` to be provided and `url` can be added on top of that.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+### API Schema
+
+```
+// Shows a site access request. Request will only be signaled to the user if
+// extension can be granted access to site in the request (i.e., one specified
+// in host_permissions, optional_host_permissions, a content script match
+// pattern, or an applicable activeTab site).
+// Resolves whether the request is valid and is signaled to the user.
+Promise<bool> <browser>.permissions.showSiteAccessRequest(
```
xeenon: As mentioned in the meeting this morning, I think `permissions.addSiteAccessRequest` would be a better name for this. At least in Safari, there is no guarantee this will show anything to the user and the name can be misleading in that regard.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+## Implementation
+
+### API Schema
+
+```
+// Shows a site access request. Request will only be signaled to the user if
+// extension can be granted access to site in the request (i.e., one specified
+// in host_permissions, optional_host_permissions, a content script match
+// pattern, or an applicable activeTab site).
+// Resolves whether the request is valid and is signaled to the user.
```
xeenon: I think it should only resolve if the request is valid, since for Safari's case, there is no guarantee this will be shown soon (if the user is not on a site with a matching URL).

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+  // The id of a document where site access requests can be shown.  Must be
+  // the top-level document within a tab.  If provided, the request is shown on
+  // the tab of the specified document and is removed when the document
+  // navigates to a new origin.
+  // Chrome requires either this or `tabId` to be specified.
+  string?: documentId,
+  // The URL pattern where site access requests can be shown. If provided,
+  // site access requests will only be shown on URLs that match this pattern.
+  // Browsers may require different levels of specificity.
+  string?: url
```
xeenon: I wouldn't mention "pattern", this is a full URL, not a match pattern. Right?

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+// Resolves whther the request was removed.
+Promise<bool> <browser>.permissions.removeSiteAccessRequest(
+  // The id of the tab where site access request will be removed.
+  // Chrome requires either this or `documentId` to be specified.
+  number?: tabId,
+  // The id of a document where site access request will be removed.
+  // Chrome requires either this or `tabId` to be specified.
+  string?: documentId,
+  // The URL pattern where site access request will be removed.
+  string?: url
```
xeenon: Ditto on "pattern".

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+### Consumers
+
+Extensions that want to signal the user when they need site access.
+
+## Implementation
+
+### API Schema
+
+```
+// Shows a site access request. Request will only be signaled to the user if
```
xeenon: "Adds a site access request. …"

xeenon (changes requested):

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: As discussed on _minutes/2024-02-29-wecg.md, returning whether request is valid is preferred. Resolving this comment

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Agree. Fixed

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Done

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Having it as a match pattern, and browser having different levels of specificity means a browser could require full paths. Left it open ended, because wasn't aware what was preferred by each browser. Any thoughts?

EmiliaPaz:
In the public WECG meeting (see _minutes/2024-02-29-wecg.md), we decided to move forward with `url` parameter, and each browser could determine which parameters it requires (e.g Chrome requires `tabId` or `documentId` and Safari requires `url`).

The only decision left is whether the parameters `tabId`, `documentId` and `url` are AND or OR:
a) All parameters can be provided and are all matched.
b) Only one of the parameters can be provided.

We (Chrome) think option a) is better because it maintains the same signature for all browsers. For example:
```
permissions.addSiteAccessRequest({tabId: 123, url:"example.com"})
```
would be a valid API call and behave the same in every browser.
They key part here is that *every parameter can be provided*. While we have discussed `tabId` by itself doesn't add value to Safari, adding it on top doesn't hurt or could even be helpful. Same with `url` for Chrome. It leaves the decision to the extension to add more matching granularity.

Option b. means the same signature could not be used across browsers and there would need to be some try/catch or parameter handling. 

We believe a. is better for the extensions ecosystem. However, if other browsers prefer b. we will also accept that (since it supports the API proposal goal)

proposals/permissions-requestSiteAccess-api.md
xeenon: I would name the property `matchPattern` or `pattern` then.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+### API Schema
+
+```
+// Adds a site access request. Request will only be signaled to the user if
+// extension can be granted access to site in the request (i.e., one specified
+// in host_permissions, optional_host_permissions, a content script match
+// pattern, or an applicable activeTab site).
+// Resolves whether the request is valid.
+Promise<bool> <browser>.permissions.addSiteAccessRequest(
```
Rob--W: I don't see any `{` and `}` here. Without these, you are specifying positional parameters instead of properties of an object. An object is generally preferable because it supports more options in the future without ambiguity.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+  callback?: function,
+);
+```
+
+Note: We don’t support iframes since they are not included in the runtime host permissions UI.
+
+### New Permissions
+| Permission Added | Proposed Warning                                         |
+| ---------------- | -------------------------------------------------------- |
+| N/A              | Permission’s API is used to “request declared optional permissions at run time rather than install time, so users understand why the permissions are needed and grant only those that are necessary” (according to documentation). The goal of this new method is for the extension to request site access, which is effectively a permission. We can adjust the  description to not be restricted just to optional permissions. |
```
Rob--W: "permission's API" -> "permissions API"? (There is no extension API called "permission")

Do we want this API to be unconditionally available to all extensions? I think that it would be good to put this behind a permission or manifest key that doesn't require a warning, because that makes it easier to statically determine whether extensions use this feature.

Otherwise it is rather unfeasible to enforce the policy of responsible use, and the API becomes useless as a signal to request attention from the user.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### Other Alternatives considered
+
+- Action API is used to control the extension’s button in the browser’s toolbar. It’s exposed if the extension includes the "action" key in the manifest. This is troublesome since an extension could not have an action, but still want to show site access requests. We should not limit requests for extensions with actions
```
Rob--W, line 90: What are use cases benefiting from the proposed API without action button? The current bullet point makes it sound like there are extensions without button that need the API. But the typical use case, i.e. shopping extension, would usually have a button.

An API to draw attention to the action button would be more generic and enable more use cases beyond just granting access to the top-level origin. It would also enable an extension to offer more context to the permission request.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+This API does not directly expose any sensitive data to the extension. However, it could lead to the extension gaining access to the site.
+
+### Abuse Scenarios
+
+Extensions can enable site access requests on every tab, which could be annoying to the user.
+
+#### Mitigations
+Attacks of annoyance are not in our threat model. Extensions can have annoying behaviors, but they risk uninstallation.
+On the browser side, we can place some restrictions to limit noisiness:
```
Rob--W, line 139: The UI is typically bound to visible tabs. The document doesn't explicitly call out the desired behavior when the API is used on background tabs. What do you want to do there?

Supporting background tabs means that the browser needs to keep track of the access request for every tab and extension.

Limiting it to visible tabs only forces extensions to look at the current URL whenever the tab switches.

There are pros and cons to either approach. Given the lack of call-outs in this document, should I assume that you intend for the flag to stick independently of whether the tab is in the foreground?

Another option that solves this issue and limits abuse is to allow an extension to have only one pending request at any time, and clear the previous request when asked again. If justified we can also consider more than one request, but still a limited number.

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: > "permission's API" -> "permissions API"? (There is no extension API called "permission")

Correct, it was a typo.

> Do we want this API to be unconditionally available to all extensions? ...

I'm not sure if I follow this. 
  - Permissions API does not require a warning. Calling `permissions.request(x)` could trigger a warning iff x has a warning.
  - Permissions API doesn't require an entry in manifest permissions
Enforcing responsible use does not have to be connected to whether the permission 'Permissions API' is present, instead each browser can determine it (e.g limit to a number of calls if user has dismissed it on a site). Why would it be useless?

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: In Chrome, an extension without an action can request host permissions, which can be withheld. To grant access, the user can click on the extension icon or extension entry in extensions menu. Once the extension has access, the extension is not clickable because it doesn't have an action.

Now, it's unlikely that an extension doesn't specify an action. The majority wants at least to specify the action's "default icon". However, that is not required and there are extensions that don't specify an action and require host permissions. For example, there is an extension that doesn't have an action and injects a script that makes google docs comments wider.

Originally, I was very undecided between permissions vs action. Went with permission for the reason explained, although I can also see argument why it would be better to have it with action

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Done

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Done

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Done

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Good callout! I think we should allow developers to specify inactive tabs. Request will be shown by the browser when the user activated the tab. As you highlighted, it will require for the browser to keep track of the access request for every tab and extension. This shouldn't be problematic, it's a small amount of data.

Updated proposal (objectives section). What do you think?

proposals/permissions-requestSiteAccess-api.md
Rob--W: > > Do we want this API to be unconditionally available to all extensions? ...
> 
> I'm not sure if I follow this.

I was wondering whether we should add a new permission or manifest key, because requiring either means that the set of extensions that can abuse it drops from all to only a few (that use it), which can enable reviewers in the extension store to pay closer attention to verify whether the use of the API is acceptable.

> * Permissions API does not require a warning. Calling `permissions.request(x)` could trigger a warning iff x has a warning.

By "that doesn't require a warning" I was emphasizing that the new permission/manifest key does not need a permission message. There are several APIs that are behind permissions that do not trigger a warning message in the prompt. This could be another one.

> Enforcing responsible use does not have to be connected to whether the permission 'Permissions API' is present, instead each browser can determine it (e.g limit to a number of calls if user has dismissed it on a site). Why would it be useless?

So from the browser perspective it doesn't matter, since the enforcement is based on the use of the API. In the broader ecosystem, it may be desirable to know whether an API may be used without running all extensions (e.g. for review, auditing, capability analysis, etc.). The manifest.json file is a machine-readable format, and a new permission (or manifest key) enables automatic scanning.

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+### API Schema
+
+```
+// Adds a site access request. Request will only be signaled to the user if
+// extension can be granted access to site in the request (i.e., one specified
+// in host_permissions, optional_host_permissions, a content script match
+// pattern, or an applicable activeTab site).
+// Resolves whether the request is valid.
+Promise<{bool}> <browser>.permissions.addSiteAccessRequest(
```
xeenon: The `{bool}` here looks incorrect. I think @Rob--W was intending it to be around the arguments, not the promise result.

xeenon (approved):

proposals/permissions-requestSiteAccess-api.md
Rob--W: @xeenon is right. I was referring to the API shape. Below is an example of how my feedback can be addressed.

Due to the duplication of the type in `addSiteAccessRequest` and `removeSiteAccessRequest`, it may also make sense to declare a separate type, like you did in one of the userScripts proposals. In reality, when implementing in Chrome, you would probably specify a type.

```suggestion
Promise<boolean> browser.permissions.addSiteAccessRequest({
```

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+  // The id of a document where site access requests can be shown.  Must be
+  // the top-level document within a tab.  If provided, the request is shown on
+  // the tab of the specified document and is removed when the document
+  // navigates to a new origin.
+  // Chrome requires either this or `tabId` to be specified.
+  string?: documentId,
+  // The URL pattern where site access requests can be shown. If provided,
+  // site access requests will only be shown on URLs that match this pattern.
+  // Browsers may require different levels of specificity.
+  string?: pattern
```
Rob--W:
```suggestion
  string?: pattern,
},
```

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+  callback?: function,
+);
+
+// Removes a site access request, if existent.
+// Resolves whther the request was removed.
+Promise<bool> <browser>.permissions.removeSiteAccessRequest(
```
Rob--W:
```suggestion
Promise<boolean> browser.permissions.removeSiteAccessRequest({
```

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+  // The id of the tab where site access request will be removed.
+  // Chrome requires either this or `documentId` to be specified.
+  number?: tabId,
+  // The id of a document where site access request will be removed.
+  // Chrome requires either this or `tabId` to be specified.
+  string?: documentId,
+  // The URL pattern where site access request will be removed.
+  string?: pattern
```
Rob--W:
```suggestion
  string?: pattern,
},
```

proposals/permissions-requestSiteAccess-api.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Existing Workarounds
+
+Users can change an extension's site access. The only way for extensions to signal the user they want to regain access to the site is through `permissions.request()` flow which is noisy and requires a user gesture. Otherwise, they rely on the user realizing through the browser UI (e.g extensions menu).
+
+Extensions have no way to signal the user that they want access to the site, and rely on the user realizing through the browser UI (e.g extensions menu)
+
+### Other Alternatives Considered
+
+Specifying URL patterns instead of tabId or documentId in `permissions.addSiteAccessRequest` or in the extension's manifest. We decided against that because:
```
Rob--W: This sentence states that the proposal decides against "patterns", but the proposal includes the patterns key. I suggest rephrasing it to not have this contradiction in one proposal.

Rob--W (approved):
Approved with the new comments addressed.

FYI: You can easily commit the requested changes through Githubs interface.

proposals/permissions-requestSiteAccess-api.md
EmiliaPaz: Good catch. Done

xeenon (approved):

tophf:
The name of the new API should be using `host`, not `site`, to match the existing API and terms. It's very counter-productive and confusing when the same thing has multiple names. Should I open a new issue?

Rob--W:
> The name of the new API should be using `host`, not `site`, to match the existing API and terms. It's very counter-productive and confusing when the same thing has multiple names. Should I open a new issue?

Thanks for your feedback @tophf . We discussed this during today's meeting, and @oliverdunk is going to check with Chrome engineering about changing to requestHostAccess. Firefox and Safari are supportive of the proposed name.

oliverdunk:
> @oliverdunk is going to check with Chrome engineering about changing to requestHostAccess.

We are supportive of this change, and should be able to make it before the API moves beyond the Canary channel. This is being tracked here: https://issues.chromium.org/380291902

oliverdunk:
To clarify, the new name would be `permissions.addHostAccessRequest` to match the final name chosen in this proposal.

PR #540, opened 2024-02-07
Add userScripts.execute() API proposal

EmiliaPaz:
Proposal for adding `<browser>.userScripts.execute()` API to allow extensions to inject user scripts programmatically into web contents.

### DIFF
--- a/proposals/user-scripts-execute-api.md
+++ b/proposals/user-scripts-execute-api.md
@@ -51,7 +51,7 @@ dictionary UserScriptInjection {
   // Details specifying the target into which to inject the script.
   target: InjectionTarget,
   // The JavaScript "world" to run the script in. The default is `USER_SCRIPT`.
-  world?: ExecutionWorld
+  world?: ExecutionWorld,
 }
 
 dictionary InjectionTarget {
@@ -60,11 +60,11 @@ dictionary InjectionTarget {
   allFrames?: boolean,
   // The IDs of specific documentIds to inject into. This must not be set if
   // frameIds is set.
-  documentIds?: string[]
+  documentIds?: string[],
   // The IDs of specific frames to inject into.
-  frameIds?: number[]
+  frameIds?: number[],
   // The ID of the tab into which to inject.
-  tabId: number
+  tabId: number,
 }
 
 dictionary InjectionResult {
@@ -73,9 +73,9 @@ dictionary InjectionResult {
   // The frame associated with the injection.
   frameId: number,
   // The result of the script execution.
-  result?: any
+  result?: any,
   // The error, if any. error and result are mutually exclusive.
-  error?: any
+  error?: any,
 }
 ```
 
@@ -86,7 +86,7 @@ dictionary InjectionResult {
 // `document_idle` unless the `injectImmediately` property is set. If the
 // script evaluates to a promise, the browser will wait
 // for the promise to settle and return the resulting value.
-Promise<InjectionResult[]> <browser>.userScripts.execute(
+Promise<InjectionResult[]> browser.userScripts.execute(
   // The details of the script which to inject
   injection: UserScriptInjection
 )
--- a/proposals/user-scripts-execute-api.md
+++ b/proposals/user-scripts-execute-api.md
@@ -1,0 +1,148 @@
+# Extension API Proposal: userScripts.execute()
+
+## Background
+
+### User Scripts and User Scripts Managers
+
+User scripts are (usually relatively small) snippets of code that are injected into web pages in order to modify the page's appearance or behavior. User script managers are a type of extension that is used to manage the collection and injection of these user scripts, determining when and how to inject them on web pages.
+
+### Manifest V3
+
+Manifest Version 3 (MV3) restricts the ability for extensions to use remotely-hosted code since it is a major security risk (stores cannot review or audit code that isn't in the extension package). This is directly at odds with user script managers, which fundamentally require the execution of arbitrary code (the user scripts themselves). This created a feature gap between MV2 and MV3.
+
+### User Scripts API
+
+The User Scripts API was introduced in MV3 to register user scripts with arbitrary code, specifically targeting user script managers' use cases. User scripts are injected based on url matching through match patterns and globs. The API requires a new permission, and allows each browser to implement their own restrictions.
+In the User Script API proposal, executing a user script one time was [presented](https://github.com/w3c/webextensions/blob/main/proposals/user-scripts-api.md#execute-user-scripts-one-time) as a potential future enhancement.
+
+## API Overview
+
+### Objective
+
+Allow extensions to inject user scripts programmatically into web content given a target context.
+
+#### Use Cases
+
+The user wants to inject a user script on the page currently loaded in a given tab one time, but does not want to inject the script on that site automatically in the future. That is, the parameter that determines the injection is a tab ID and not based on origin.
+
+### Availability: Public
+
+This API will be public to all extensions (with the appropriate permission).
+
+### Consumers
+
+Any existing user script manager consumers of `chrome.tabs.executeScript()` that inject user scripts programmatically are intended to migrate to this new version in Manifest V3.
+
+## Implementation
+
+### API Schema
+
+#### Types
+
+```
+dictionary UserScriptInjection {
+  // Whether the injection should be triggered in the target as soon as
+  // possible. Note that this is not a guarantee that injection will occur
+  // prior to page load, as the page may have already loaded by the time the
+  // script reaches the target.
+  injectImmediately?: boolean,
+  // The script source to inject into the target.
+  js: ScriptSource,
+  // Details specifying the target into which to inject the script.
+  target: InjectionTarget,
+  // The JavaScript "world" to run the script in. The default is `USER_SCRIPT`.
+  world?: ExecutionWorld
+}
+
+dictionary InjectionTarget {
+  // Whether the script should inject into all frames within the tab. Defaults
+  // to false. This must not be true if `frameIds` is specified.
+  allFrames?: boolean,
+  // The IDs of specific documentIds to inject into. This must not be set if
+  // frameIds is set.
+  documentIds?: string[]
+  // The IDs of specific frames to inject into.
+  frameIds?: number[]
+  // The ID of the tab into which to inject.
+  tabId: number
+}
+
+dictionary InjectionResult {
+  // The document associated with the injection.
+  documentId: string,
+  // The frame associated with the injection.
+  frameId: number,
+  // The result of the script execution.
+  result?: any
+}
+```
+
+#### Methods
+
+```
+// Injects a script into a target context. The script will be run at
+// `document_idle` unless the `injectImmediately` property is set. If the
+// script evaluates to a promise, the browser will wait
+// for the promise to settle and return the resulting value.
+Promise<InjectionResult[]> <browser>.userScripts.execute(
+  // The details of the script which to inject
+  injection: UserScriptInjection
+)
+```
+
+Note: Promises are supported in Manifest V3 and later, but callbacks are provided for backward compatibility. You cannot use both on the same function call. The promise resolves with the same type that is passed to the callback.
+
+### New Permissions
+
+No new permissions will be added; this will be part of the existing `userScripts` permission
+
+### Manifest Changes
+
+N/A
+
+### Custom Bindings
+
+N/A
+
+## User Experience
+
+### UI Elements and User-Visible Effects
+
+There are no new UI elements or user-visible effects (exempting those caused by the script of the extension).
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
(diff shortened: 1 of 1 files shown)


Getfree:
What would be the function's arguments?
Does it accept a "code" parameter?
Does it accept a "runAt" parameter?
Is it going to be affected by the CSP of the target page?
Will the script run in the global scope or a local scope? (i.e. can the script create global variables?)

tophf:
Hopefully you don't mind me posting the answers to these questions which I found while researching the PR as it's something I'll use in my extensions as well:

<details><summary>affected by the CSP of the target page?</summary>

chrome.userScripts.register runs code in the MAIN world despite github.com's `script-src` CSP, so injection itself is not affected and the same should work for chrome.userScripts.execute, **but a clarification for the `MAIN` world would be helpful anyway** as the [documentation mentions](https://developer.chrome.com/docs/extensions/reference/userScripts/index-2.html#type-ExecutionWorld) CSP exemption only for the USER_SCRIPT world, so it could clarify that the code will be injected in the MAIN world regardless of page's CSP, while things created by this code will be subject to the CSP of the page e.g. a `style` element will be blocked by a restrictive `style-src` CSP.

</details>
<details><summary>arguments</summary>

It's `UserScriptInjection` object in the code of the PR, which expectedly looks like a hybrid of chrome.userScripts.register and chrome.scripting.executeScript.

</details>
<details><summary><code>code</code></summary>

Yes, it's `js: [{code: 'console.log(1)'}]` same as chrome.userScripts.register, see [ScriptSource](https://developer.chrome.com/docs/extensions/reference/api/userScripts#type-ScriptSource) definition.

</details>
<details><summary><code>runAt</code></summary>

It's a boolean `injectImmediately` same as chrome.scripting.executeScript, false (by default) [means document_idle](https://crsrc.org/chrome/browser/extensions/api/scripting/scripting_api.cc;l=686;drc=c1857d3b72f27e8259b79108160a475b57d5b2ee).

</details>
<details><summary>global scope or a local scope</summary>

chrome.userScripts.register uses global scope, so it should be the same, but either way we can create global variables explicitly like `window.foo = 123`, which would be necessary when `func` is implemented in the future as it'll run in an IIFE.

</details>

EmiliaPaz:
Thanks @Getfree for taking a look at the proposal. And special thanks to @tophf for providing answers. These are mostly correct, but I am expanding them more:

> What would be the function's arguments?

`userScripts.execute(injection: UserScriptInjection)`. You can see it in more detail on the `API Schema` section in the proposal.

> Does it accept a "code" parameter?

Yes, injection has a User Script API [`ScriptSource`](https://developer.chrome.com/docs/extensions/reference/api/userScripts#type-ScriptSource) parameter which includes code.

> Does it accept a "runAt" parameter?

No. By default in run at `document_idle` but can be set to `injectImmediately`

> Is it going to be affected by the CSP of the target page?

Depends on the world it's injected to. User script can be registered/executed in the `USER_SCRIPT` or `MAIN` world. The `USER_SCRIPT` world is an execution environment specific to user scripts and is exempt from the page's CSP. The `MAIN` world is the execution environment shared with the host page's JavaScript, and thus follows the page's CSP.

@tophf , not sure I follow your answer here. Extension can decide the world in which a script is registered. Script registered in `MAIN` world will follow the page's CSP, and when registered in `USER_SCRIPT` world will follow the extension CSP or a customized one For example:
  -  `script1` is registered in `MAIN` world and it does not have access to `eval()` since page csp dissallows `eval()`. 
  - `script2` is registered in `USER_SCRIPT` world. Extension can configure the `USER_SCRIPT` world to allow `eval()`, does script gets access to `eval()`
 
> Will the script run in the global scope or a local scope? (i.e. can the script create global variables?)

It runs on the global scope.

tophf:
> @tophf , not sure I follow your answer here. 

I was referring to the existing problem in the currently used workaround for MV3 to run arbitrary code by creating a script element inside code that already runs in the MAIN world - this workaround didn't work with a strict CSP of the page. With the new userScripts API the code will run regardless of the CSP of the page (it still affects the artifacts created by this code). I think this is an important info that could be used for an explicit clarification in the documentation.

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+This API does not directly expose any new sensitive data, as it relies on the host permissions of the extension and the userScripts permission. If the extension already has host permissions to a site, it can access all the data associated with that site (including scraping the DOM, accessing cookies, modifying network requests, and more).
+
+### Abuse Scenarios
+
+With any script injection API, there are significant abuse possibilities - such as injecting malicious JavaScript intended to collect user data maliciously.  Additionally, this allows extensions to inject code that's not contained within the extension's package  However, this is inline with the existing userScripts API, and there are no new abuse mechanisms introduced as a part of this API.
```
Rob--W:
```suggestion
With any script injection API, there are significant abuse possibilities - such as injecting malicious JavaScript intended to collect user data maliciously.  Additionally, this allows extensions to inject code that's not contained within the extension's package. However, this is inline with the existing userScripts API, and there are no new abuse mechanisms introduced as a part of this API.
```

Nit: missing period.

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Mitigations
+The existing mitigations for host permissions - most dominantly, runtime host permissions - apply here.  Additionally, we can leverage client-side telemetry for the injected scripts as-needed.
```
Rob--W:
```suggestion
The existing mitigations for host permissions - most dominantly, runtime host permissions - apply here.
```

The part about telemetry looks like a Chrome-specific consideration that ought not be part of a public proposal.

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,148 @@
...
+
+In the past, developers leveraged `chrome.tabs.executeScript()`. However, this method was removed in MV3.
+
+### Existing Workarounds
+
+Developers can utilize `userScripts.register()` to inject JavaScript into a known set of hosts, but this doesn't cover the case of programmatic injections to a specific target or one-time injections.
+
+### Open Web API
+
+Since this API exposes the ability to inject JavaScript on sites that are not in the same origin, it does not belong on the web
```
Rob--W:
```suggestion
Since this API exposes the ability to inject JavaScript on sites that are not in the same origin, it does not belong on the web.
```

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,148 @@
...
+  tabId: number
+}
+
+dictionary InjectionResult {
+  // The document associated with the injection.
+  documentId: string,
+  // The frame associated with the injection.
+  frameId: number,
+  // The result of the script execution.
+  result?: any
```
Rob--W: One `result` implies that there is an unambiguous way to encode the result. That is not the case if there are errors, ranging from runtime errors (while executing the script) to permission errors.

I reported a similar issue about `scripting.executeScript` before at https://issues.chromium.org/issues/40205757
Proposed fix: add an `error` property (MDN documentation with `error` property of scripting.executeScript: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript#return_value):

```suggestion
  // The result of the script execution.
  result?: any
  // The error, if any. error and result are mutually exclusive.
  error?: any
```

rdcronin: Agreed; this has been something I've been thinking about in the executeScript() API for [awhile](https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/data/extensions/api_test/scripting/main_frame/worker.js;l=275-277;drc=bf596c951f436be7f275553a09921064c2de9e93).

EmiliaPaz: This makes sense. Added to the proposal. We can also look to add this to `scripting.executeScript()`

EmiliaPaz: Taking a second look, should error be a string? In chrome, the error associated with the injection is a string (see [FrameResult](https://source.chromium.org/chromium/chromium/src/+/1b571bef02256072be9e289270d6067da1e0de73:extensions/browser/script_executor.h;l=88-90)). Would that work for other browsers?

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,150 @@
...
+  // possible. Note that this is not a guarantee that injection will occur
+  // prior to page load, as the page may have already loaded by the time the
+  // script reaches the target.
+  injectImmediately?: boolean,
+  // The script source to inject into the target.
+  js: ScriptSource,
+  // Details specifying the target into which to inject the script.
+  target: InjectionTarget,
+  // The JavaScript "world" to run the script in. The default is `USER_SCRIPT`.
+  world?: ExecutionWorld
```
Rob--W: Missing comma at end. This is also the case at many of the other types. Let's add them.

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+#### Methods
+
+```
+// Injects a script into a target context. The script will be run at
+// `document_idle` unless the `injectImmediately` property is set. If the
+// script evaluates to a promise, the browser will wait
+// for the promise to settle and return the resulting value.
+Promise<InjectionResult[]> <browser>.userScripts.execute(
```
Rob--W: Since we're working towards standardizing the `browser` namespace, let's just use it instead of the `<browser>` placeholder.

```suggestion
Promise<InjectionResult[]> browser.userScripts.execute(
```

Rob--W (approved):

xeenon (approved):

proposals/user-scripts-execute-api.md
Rob--W: Oops, I missed this question, so answering now.

> Taking a second look, should error be a string? In chrome, the error associated with the injection is a string (see [FrameResult](https://source.chromium.org/chromium/chromium/src/+/1b571bef02256072be9e289270d6067da1e0de73:extensions/browser/script_executor.h;l=88-90)). Would that work for other browsers?

A string would be insufficient to catch user-defined "errors", whether from the last expression being `Promise.reject(12345)` or by an explicit `throw 12345`. For this reason, we have declared the type as `any`, and documented at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript#error as follows:

> If an error occurs, contains the value the script threw or rejected with. Typically this is an error object with a message property but it could be any value (including primitives and undefined).

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Alternatives
+
+### Past Workarounds
+
+In the past, developers leveraged `chrome.tabs.executeScript()`. However, this method was removed in MV3.
+
+### Existing Workarounds
+
+Developers can utilize `userScripts.register()` to inject JavaScript into a known set of hosts, but this doesn't cover the case of programmatic injections to a specific target or one-time injections.
```
rdcronin, line 140: They can also use `scripting.executeScript()` today, but that doesn't allow for remotely-hosted code.

proposals/user-scripts-execute-api.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Open Web API
+
+Since this API exposes the ability to inject JavaScript on sites that are not in the same origin, it does not belong on the web.
+
+## (Potential) Future Enhacements
+
+### Add func to `ScriptSource`
+
+Add `func` and `args` property to `ScriptSource` to specify a JavaScript function to inject. This could be used both by `userScripts.register()` and `userScripts.execute()`.
```
rdcronin, line 150: It's worth highlighting that this isn't as useful in this context, since the extension can trivially construct a string that enables this same functionality.  Additionally, if the extension is using func + args, it isn't using remotely-hosted code, and could also use `scripting.executeScript()`.  The only case it would need to use this in that scenario would be to inject in the user script world.

rdcronin (approved):
LG; thanks, Emilia!

proposals/user-scripts-execute-api.md
tophf: ...also when Chrome makes `args` use the structured clone algorithm.

rdcronin: Yep, that's a reasonable point, too

EmiliaPaz:
This PR has been approved and is ready to merge (I don't have the power to do so :) )

Rob--W:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1930776 to track the implementation of this in Firefox. Is there any Chromium bug for this?

EmiliaPaz:
> I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1930776 to track the implementation of this in Firefox. Is there any Chromium bug for this?
Yes, https://issues.chromium.org/u/2/issues/326657581. Curretly working on the implementation

newRoland:
Launching in Chrome soon (Chrome 134). 

https://developer.chrome.com/blog/extension-news-january-2025#userscriptsexecute_in_canary

PR #542, opened 2024-02-09
Add content scripts section in specification

oliverdunk:
Adds a first draft of information on content scripts in the specification.

There are still some updates needed, in particular around the algorithm for deciding when to inject a script, but I wanted to open something to get some early feedback.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webextensions/pull/542.html" title="Last updated on Jan 29, 2025, 5:28 PM UTC (5093e8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webextensions/542/8962bfa...5093e8f.html" title="Last updated on Jan 29, 2025, 5:28 PM UTC (5093e8f)">Diff</a> (Both outdated!)


specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+
+#### Key `match_about_blank`
+
+If this is `true`, the content script will also be injected into an additional user agent specified set of pages used to represent empty frames. This will only happen if the content script matches the page that embedded the frame. Defaults to `false`.
+
+#### Key `match_origin_as_fallback`
+
+Used to match frames with an opaque or otherwise missing origin. The origin to match against is determined in the following order of priority:
+
+1. If the frame has an [=opaque origin=], such as with a [=blob URLs=], use the non-opaque origin.
```
oliverdunk: @Rob--W @rdcronin Would you be able to take a look at this one and confirm if it is accurate? This was my best understanding based on bugs and documentation in the code.

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
 
 ### Content scripts
 
-#### Isolated worlds
+<dfn>Content scripts</dfn> represent a set of JS and CSS files that should be injected into pages loaded by the user agent.
```
rdcronin:
```suggestion
<dfn>Content scripts</dfn> represent a set of JS and CSS files that should be injected into matching pages loaded by the user agent.
```

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+
+#### Key `matches`
+
+A [=list=] of [=match patterns=] that are used to decide where the content script runs. This key is required.
```
rdcronin:
```suggestion
A [=list=] of [=match patterns=] that are used to which pages the user agent injects the content script into. This key is required.
```

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+
+#### Key `exclude_matches`
+
+A [=list=] of [=match patterns=] that should be used to exclude URLs from where the content script runs.
+
+#### Key `js`
+
+A [=list=] of file paths that should be injected as scripts.
```
rdcronin:
```suggestion
A [=list=] of file paths, relative to the extension's package, that should be injected as scripts.
```

Also below

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+
+#### Key `css`
+
+A [=list=] of file paths that should be injected as stylesheets.
+
+#### Key `all_frames`
+
+If `all_frames` is true, the content script must be injected into subframes. Defaults to false.
```
rdcronin:
```suggestion
If `all_frames` is true, the content script must be injected into any subframes that match the content script's injection criteria. If false, content scripts will only be injected into top-level documents. Defaults to false.
```

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+
+#### Key `match_about_blank`
+
+If this is `true`, the content script will also be injected into an additional user agent specified set of pages used to represent empty frames. This will only happen if the content script matches the page that embedded the frame. Defaults to `false`.
```
rdcronin: Do we (browsers) have different criteria for match_about_blank?

specification/index.bs
rdcronin: I think I'd rephrase this.

(The issue with the current language is that:
a] it doesn't specify what the "non-opaque" origin is or where it comes from
b] it doesn't always use the origin of the parent; it uses the initiator (or "precursor"))

If the URL of a document has a specified scheme**, the user agent will fall back to the origin of the initiator instead.  This is commonly, but not always, the parent or embedding frame.

** In chrome, these schemes are data:, about:, filesystem:, and blob:.  Is that the same in other browsers?

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+1. If available, use the origin of the parent frame.
+1. Otherwise, no origin is found and this frame can never be matched.
+
+#### Key `run_at`
+
+Specifies when the content script should be injected. Valid values are `document_start`, `document_end` and `document_idle`.
+
+#### Key `include_globs`
+
+A list of [=globs=] that a page should match in addition to matches.
```
rdcronin:
```suggestion
A list of [=globs=] that a page should match. A page matches if the URL matches both the [=matches=] field and the [=globs=] field.
```

(Feel free to tweak; as written, it's ambiguous if this is an additional set to inject into or an additional requirement)

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+#### Injecting a content script
+
+Issue: If the same extension specifies the same script twice, what should happen? ([bug](https://crbug.com/324096753))
+
+Issue: The below algorithm needs to be updated to include `match_about_blank` and `match_origin_as_fallback`.
+
+To determine if a content script should be injected in a frame:
+
+1. If the extension does not have access to the origin, return.
+1. If the origin is not included in `matches`, return.
```
rdcronin: matches don't have to be origin-based; they can include paths

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+1. If `include_globs` is present and the origin is not matched, return.
+1. If the origin matches an entry in `exclude_matches` or `exclude_globs`, return.
+1. If this is a frame, and `all_frames` is not `true`, return.
```
rdcronin:
```suggestion
1. If this is a child frame, and `all_frames` is not `true`, return.
```

Everything's a frame : )

rdcronin (changes requested):
Thanks, Oliver!  I had a chance to take a quick pass on this one.

specification/index.bs
```diff
@@ -172,6 +175,12 @@ Issue(62): Specify localization handling.
 
 ## Match patterns
 
+A <dfn>match pattern</dfn> is a pattern used to match URLs.
```
jpmedley: A pattern of what?

oliverdunk: Definitely agree we need to add more. This was just a stub that I hope to replace with more detail :)

specification/index.bs
```diff
@@ -27,11 +28,11 @@ An optional directory containing strings as defined in <a href="#localization">l
 
 ### Other files
 
-An extension may also contain other files, such as those referenced in the <a href="#key-content_scripts">content_scripts</a> and <a href="#key-background">background</a> part of the <a href="#manifest">Manifest</a>.
+An extension may also contain other files, such as those referenced in the [[#key-content_scripts]] and [[#key-background]] part of the [=manifest=].
```
jpmedley:
```suggestion
An extension may also contain other files, such as those referenced in the [[#key-content_scripts]] and [[#key-background]] parts of the [=manifest=].
```

specification/index.bs
```diff
@@ -154,6 +155,8 @@ Filenames beginning with an underscore (`_`) are reserved for use by user agent.
 
 ## Isolated worlds
 
+<dfn>Worlds</dfn> are isolated JavaScript contexts with access to the same underlying DOM tree but their own set of wrappers around those DOM objects.
```
jpmedley: I'm not sure about this definition. In Chrome worlds are main, isolated, and worker. So a main world is never isolated. I don't know what the intent is with regard to the spec, but there's potential for confusion here.

jpmedley (changes requested):
Sorry for the extra review. I didn't remember that I had looked at this already until I got to my first comment halfway through it.

specification/index.bs
dotproto: Devlin's suggestion appears to have a dropped a word.

```suggestion
A [=list=] of [=match patterns=] that are used to decide which pages the user agent injects the content script into. This key is required.
```

specification/index.bs
dotproto: Tagging @Rob--W and @xeenon to request feedback.

specification/index.bs
```diff
@@ -190,7 +199,70 @@ Issue(62): Specify localization handling.
...
+Issue: The below algorithm needs to be updated to include `match_about_blank` and `match_origin_as_fallback`.
+
+To determine if a content script should be injected in a frame:
+
+1. If the extension does not have access to the origin, return.
+1. If the origin is not included in `matches`, return.
+1. If `include_globs` is present and the origin is not matched, return.
+1. If the origin matches an entry in `exclude_matches` or `exclude_globs`, return.
+1. If this is a frame, and `all_frames` is not `true`, return.
+1. Otherwise, inject the content script. This should be done based on the `run_at` setting.
```
dotproto: I think it would be best to move definitions like this into their own sections. Unfortunately I can't find a W3C doc that recommends how to organize a spec, but the two main patterns I've seen skimming over looking over existing specifications I noticed the following patterns:

* Collect all algorithms definitions under a top level section labeled Algorithms ([Service Workeres](https://www.w3.org/TR/service-workers/#algorithms), [IndexedDB](https://www.w3.org/TR/IndexedDB/#algorithms))
* Define algorithms in sub-sections of a broader topic ([CSS Color Level 5](https://www.w3.org/TR/css-color-5/#color-mix-percent-norm))
* Define an algorithm in the flow of related content ([Push API](https://www.w3.org/TR/push-api/#dfn-extract-a-byte-sequence))

Confusingly, [Clipboard API and events](https://www.w3.org/TR/clipboard-apis/#to-process-html-paste-event) seems to do all of these.
* It has an [Algorithms appendix](https://www.w3.org/TR/clipboard-apis/#algorithms)
* It describes the [check clipboard write permission](https://www.w3.org/TR/clipboard-apis/#h-clipboard-write-permission) process in it's own section.
* It describes how to [create a ClipboardItem object](https://www.w3.org/TR/clipboard-apis/#create-a-clipboarditem-object) in the flow of another section.

I'm getting a headache.

dotproto (changes requested):

specification/index.bs
Rob--W: @oliverdunk The semantics have extensively been discussed on Chromium's issue tracker where I and Devlin discussed the API design. If you're interested, the start of the discussion is at https://issues.chromium.org/issues/40443085#comment48. The design that is close to what we have now was sketched in https://issues.chromium.org/issues/40443085#comment61 , with the final name (`match_origin_as_fallback`) at https://issues.chromium.org/issues/40443085#comment67. Devlin summarized the discussion at https://issues.chromium.org/issues/40443085#comment71

Upon reviewing the proposed texts here, I think that there is some confusion on terminology. The current text mentions blob URLs as an opaque origin, but that is not the case.

Relevant to content script matching is the URL of the document (which can have an origin component) and the origin of the document (as a security principal). There may not always be an obvious relation between the two:

- URLs may have visible origin parts in it, such as http(s) and also `blob:` and (Chrome-only) `filesystem:` (e.g. `blob:https://example.com/UUID`).
- URLs may not have a visible origin in it (`about:blank` and `about:srcdoc`), but still have a non-opaque origin: commonly the opener of the frame or window is another http(s) URL. Or even any number of `about:blank`/`srcdoc` documents where the first was initially opened by a http(s) origin.
- The security principal of a document can be an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque), even if the URL of that document looks like it has a non-opaque origin. This happens with `<iframe sandbox>` or `sandbox` directive in the `Content-Security-Policy`. A content script can use `window.origin` to see whether the origin is opaque, as it would serialize to `"null"`.
  - In case of opaque origins, there is almost always a non-opaque initiator that opened the frame. The term "precursor origin" is used here `<iframe sandbox="allow-scripts" src="https://example.com">`
 - The exception is when the initiator of the navigation does not have a non-opaque origin. For example, when the user navigates to a `data:-URL` or to `about:blank`. Since `data:`-URL
   - Chrome does currently not run content scripts in these documents.
   - Firefox currently allows content scripts with `matches` for all URLs AND `match_about_blank: true` to run scripts in top-level about:blank. This is not documented anywhere though.

specification/index.bs
Rob--W: The description here is too vague. `match_about_blank` was designed for `about:blank` and `about:srcdoc`.

If you're looking for clarity, see https://stackoverflow.com/questions/41408936/can-anyone-explain-that-what-is-the-use-of-match-about-blank-in-chrome-extensi, where I previously posted an answer that describes why `match_about_blank` exists and what it does.

Other documentation:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts#match_about_blank
- https://developer.chrome.com/docs/extensions/reference/manifest/content-scripts#frames

Rob--W:
Although not obvious from Github's UI, I just posted additional context on `match_about_blank` and `match_origin_as_fallback`:
- `match_about_blank`: https://github.com/w3c/webextensions/pull/542#discussion_r1580904131
- `match_origin_as_fallback`: https://github.com/w3c/webextensions/pull/542#discussion_r1580892940

specification/index.bs
oliverdunk: Done, thanks :)

specification/index.bs
oliverdunk: I see where you're coming from, but I think I'm comfortable with this since the main world is still isolated from other worlds.

specification/index.bs
oliverdunk: Thanks both, writing the description for these two keys has taken by far the most time in this PR. I've given it another attempt and would appreciate any feedback.

As a general note, concepts like the precursor origin and security principal don't appear to be defined in any other specifications. It seems like they are more informal terms used often in implementations and by implementors. With that in mind, I've tried to describe them as best as possible without talking about them by name.

A few additional notes:

- I've added an informal note to `match_about_blank` describing the Firefox behavior for top-level about:blank pages.
- I've added a note that the path must be a wildcard if `match_origin_as_fallback` is set. This is the behavior today in Chrome. Interestingly, we don't have any restrictions on `include_globs` or `exclude_globs`. This feels like an omission to me and I wonder if we should specify something.
- In Chrome, sandboxing doesn't seem to be relevant. We always apply these fallbacks, even if the parent is inaccessible to the child frame. With that in mind, I haven't mentioned it here.

Clearly there's a lot of detail here so please let me know if I've missed anything or it could be clearer.

specification/index.bs
oliverdunk: That's fair, updated :)

specification/index.bs
oliverdunk: Thanks!

specification/index.bs
oliverdunk: Oh, thanks so much for the detailed comment and looking into this. Given the precedence moving into a new section makes sense to me, and I've gone ahead and done that.

oliverdunk (commented):
Thanks all! I've just done another pass on this, would appreciate any additional feedback.

specification/index.bs
oliverdunk: @Rob--W, could you take another look? I've made some tweaks although it's unclear to me what was too vague.

specification/index.bs
```diff
@@ -190,7 +199,57 @@ Issue(62): Specify localization handling.
...
+
+A [=list=] of file paths, relative to the extension's package, that should be injected as stylesheets.
+
+### Key `all_frames`
+
+If `all_frames` is true, the content script must be injected into any subframes that match the other matching criteria for the content script. If false, content scripts will only be injected into top-level documents. Defaults to false.
+
+### Key `match_about_blank`
+
+If this is true, use the URL of the parent frame when matching a child frame whose document URL has the `about` [=scheme=]. See also [[#determine-the-url-for-content-script-matching]]. Defaults to `false`.
```
dotproto: Chrome currently matches all documents with an `about` scheme as described in the current copy ([source link](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/content_script_injection_url_getter.cc;l=54;drc=24a80999fbc083276e0fbef7f9e2eec36429f6a8) care of @oliverdunk), but Firefox appears to explicitly check against `about:blank` and `about:srcdoc` ([source](https://searchfox.org/mozilla-central/rev/e24f7f322960a39f9ef644a31b6026994db73850/toolkit/components/extensions/MatchPattern.cpp#192-193)). @xeenon, how does Safari handle this key?

Any changes here should be reflected in the algorithm "Determine the URL for content script matching" section below.

specification/index.bs
```diff
@@ -190,7 +199,57 @@ Issue(62): Specify localization handling.
...
+
+Note: In Firefox, setting `match_about_blank` to `true` also allows injection into top-level `about:blank` pages.
```
dotproto: We should add a non-normative label to this block.

specification/index.bs
```diff
@@ -190,7 +199,57 @@ Issue(62): Specify localization handling.
...
+
+If this is true, use fallbacks as described in [[#determine-the-url-for-content-script-matching]].
+
+No path is available when the URL to match against falls back to an origin. Therefore, when set, the user agent must not allow [[#key-matches]] to contain entries with a path other than `/*`.
+
+Defaults to `false`.
+
+### Key `run_at`
+
+Specifies when the content script should be injected. Valid values are `document_start`, `document_end` and `document_idle`.
```
dotproto: Consider adding a WebIDL definition for these values.

specification/index.bs
```diff
@@ -190,7 +199,57 @@ Issue(62): Specify localization handling.
...
+
+A list of [=globs=] that a page should match. A page matches if the URL matches both the [[#key-matches]] field and the [[#key-include_globs]] field.
+
+### Key `exclude_globs`
+
+A list of [=globs=] that should be used to exclude URLs from where the content script runs.
+
+### Key `world`
+
+The [=world=] any JavaScript scripts should be injected into. Defaults to `ISOLATED`. Valid values are `MAIN` and `ISOLATED`.
```
dotproto: Consider adding a WebIDL definition for these values.

specification/index.bs
```diff
@@ -190,7 +199,57 @@ Issue(62): Specify localization handling.
...
+
+Defaults to `false`.
+
+### Key `run_at`
+
+Specifies when the content script should be injected. Valid values are `document_start`, `document_end` and `document_idle`.
+
+### Key `include_globs`
+
+A list of [=globs=] that a page should match. A page matches if the URL matches both the [[#key-matches]] field and the [[#key-include_globs]] field.
```
dotproto: We should probably revise "a page should match" to a more generic concept like "potential injection context." Words are hard. @oliverdunk suggested maybe "document"?

specification/index.bs
```diff
@@ -203,3 +262,33 @@ Issue(62): Specify localization handling.
...
+# Algorithms
+
+## Determine the URL for content script matching
+
+To determine the URL to use for a document when injecting a content script:
+
+1. Let |url| be the document's URL.
+1. If the document is within a child frame:
+    1. If the [=scheme=] of the document's URL is `about`, and `match_about_blank` or `match_origin_as_fallback` is set to true:
+        1. Set |url| to a URL based on the origin of the parent frame.
```
dotproto: The phrase "URL based on the origin of the parent frame" feels a little … awkward? Definitely not required, but maybe we should have a definition or abstract algorithm that describes this and link out to it from here.

specification/index.bs
```diff
@@ -203,3 +262,33 @@ Issue(62): Specify localization handling.
...
+## Inject a content script
+
+Issue: If the same extension specifies the same script twice, what should happen? ([bug](https://crbug.com/324096753))
+
+To determine if a content script should be injected in a frame:
+
+1. Let |url| be the result of running [[#determine-the-url-for-content-script-matching]].
+1. If the extension does not have access to the origin, return.
+1. If |url| is not matched by a match pattern in `matches`, return.
+1. If `include_globs` is present and |url| is not matched by any pattern, return.
```
dotproto: I think this might be a bit easier for the reader to parse. 

```suggestion
1. If `include_globs` is present and |url| is not matched by any glob pattern, return.
```

dotproto (dismissed):

specification/index.bs
```diff
@@ -203,3 +262,33 @@ Issue(62): Specify localization handling.
...
+
+Issue: If the same extension specifies the same script twice, what should happen? ([bug](https://crbug.com/324096753))
+
+To determine if a content script should be injected in a document:
+
+1. Let |url| be the result of running [[#determine-the-url-for-content-script-matching]].
+1. If the extension does not have access to the origin, return.
+1. If |url| is not matched by a match pattern in `matches`, return.
+1. If `include_globs` is present and |url| is not matched by any pattern, return.
+1. If |url| matches an entry in `exclude_matches` or `exclude_globs`, return.
```
dotproto: No change needed. 

On my first pass through this I assumed that we could simplify the language here by combining steps 2 and 3 (similar to how step 4 is written), but @oliverdunk pointed out that`include_globs` behaves as I expected. He also pointed out that the user scripts behavior of `includeGlobs` intentional diverges from content scripts – the [the proposal](https://github.com/w3c/webextensions/blob/main/proposals/user-scripts-api.md#types) notes that globs are:

```
  // Implemented as disjunction: runs in documents whose URL matches
  // "matches" or "includeGlobs", and not "excludeMatches" nor "excludeGlobs".
```

specification/index.bs
oliverdunk: Looking at https://speced.github.io/bikeshed/#conformance, I'm actually getting the sense notes are non-normative by definition. Would you agree?

specification/index.bs
```diff
@@ -190,7 +199,80 @@ Issue(62): Specify localization handling.
...
+</pre>
+
+The {{RunAt}} enum represents when a content script should be injected.
+
+### <dfn>ExecutionWorld</dfn> enum
+
+<pre class="idl">
+enum ExecutionWorld {
+    "ISOLATED",
+    "MAIN"
```
xeenon, line 269: I still would like to see this specified as lowercase (or snake case). See issue #563.

Safari supports case-insensitive here.

oliverdunk: Would you be comfortable landing this to begin with, and then tackling the casing in a follow up PR? Given this is the casing that works across browsers, and we agreed to defer discussions on backwards compatibility, I'd prefer not to hold up the other changes here.

specification/index.bs
oliverdunk: I've changed this to match `about:blank` and `about:srcdoc` exclusively.

oliverdunk:
@Rob--W, I've addressed all of the actionable feedback here. Could you take another look?

specification/index.bs
```diff
@@ -203,3 +283,33 @@ Issue(62): Specify localization handling.
...
 # Version number handling
+
+# Algorithms
+
+## Determine the URL for content script matching
+
+To determine the URL to use for a document when injecting a content script:
+
+1. Let |url| be the document's URL.
+1. If the document is within a child frame:
```
Rob--W: The algorithm from this point onwards is inaccurate. A top-level about:blank opened by a web page (e.g. https://example.com/ ) will also have the `https://example.com` origin.

And if the opener was https://example.com that had `Content-Security-Policy: sandbox`, then the new window would also have an opaque origin, and still be same-origin to the opener. The precursor of the origin will still be example.com, and if `match_origin_as_fallback` is specified, `https://example.com/*` will still match.

I think that it makes more sense to base the decision tree on the document's origin.

Here is a detailed write-up of how the matching works for non obvious cases: https://github.com/w3c/webextensions/issues/575#issuecomment-2079247089 . Does that offer enough context as a starting point to spec the behavior?

I tried looking for "precursor origin" concept in the spec, but could not find any. There is an explainer for the `allow-unique-origin` CSP sandbox keyword (https://github.com/shhnjk/allow-unique-origin) that mentions it too, so if it does not exist, it may make sense to specify the concept more formally elsewhere, and just refer to the "precursor origin" concept in our spec.

Rob--W (changes requested):
I read through it all and considered approving, but then I reached the final section on the algorithm, and since that is inaccurate, I'd like to resolve that before merging.

specification/index.bs
oliverdunk: Ah, thanks for calling that extra case out. Trying to handle top-level frames here definitely makes things a bit trickier to explain simply. I've tried a few changes that use the precursor origin terminology - how does that look? Agreed we should describe that somewhere at some point but I think we should tackle that separately to avoid blocking this PR.

specification/index.bs
```diff
@@ -172,6 +175,12 @@ Issue(62): Specify localization handling.
 
 # Match patterns
 
+A <dfn>match pattern</dfn> is a pattern used to match URLs. They are case-insensitive.
+
+# Globs
+
+A <dfn>glob</dfn> can be any [=string=]. It can contain any number of wildcards where * can match zero or more characters and ? matches exactly one character.
```
xeenon:
```suggestion
A <dfn>glob</dfn> can be any [=string=]. It can contain any number of wildcards where `*` can match zero or more characters and `?` matches exactly one character.
```

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
-### Isolated worlds
+<dfn>Content scripts</dfn> represent a set of JS and CSS files that should be injected into matching pages loaded by the user agent. They are injected using the steps in [[#inject-a-content-script]].
+
+### Key `matches`
+
+A [=list=] of [=match patterns=] that are used to decide which pages the user agent injects the content script into. This key is required.
+
+### Key `exclude_matches`
+
+A [=list=] of [=match patterns=] that should be used to exclude URLs from where the content script runs.
```
xeenon:
```suggestion
A [=list=] of [=match patterns=] that can be used to exclude URLs from where the content script runs.
```

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
+
+A [=list=] of file paths, relative to the extension's package, that should be injected as scripts.
+
+### Key `css`
+
+A [=list=] of file paths, relative to the extension's package, that should be injected as stylesheets.
+
+### Key `all_frames`
+
+If `all_frames` is true, the content script must be injected into any subframes that match the other matching criteria for the content script. If false, content scripts will only be injected into top-level documents. Defaults to false.
```
xeenon:
```suggestion
If `all_frames` is `true`, the content script must be injected into any subframes that match the other matching criteria for the content script. If `false`, content scripts will only be injected into top-level documents. Defaults to `false`.
```

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
+
+### Key `match_about_blank`
+
+If this is true, use the URL of the parent frame when matching a child frame whose document URL is `about:blank` or `about:srcdoc`. See also [[#determine-the-url-for-content-script-matching]]. Defaults to `false`.
```
xeenon:
```suggestion
If this is `true`, use the URL of the parent frame when matching a child frame whose document URL is `about:blank` or `about:srcdoc`. See also [[#determine-the-url-for-content-script-matching]]. Defaults to `false`.
```

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
+
+### Key `match_origin_as_fallback`
+
+If this is true, use fallbacks as described in [[#determine-the-url-for-content-script-matching]].
```
xeenon:
```suggestion
If this is `true`, use fallbacks as described in [[#determine-the-url-for-content-script-matching]].
```

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
+
+Specifies when the content script should be injected. Valid values are defined by the {{RunAt}} enum.
+
+### Key `include_globs`
+
+A list of [=globs=] that a document should match. A document matches if the URL matches both the [[#key-matches]] field and the [[#key-include_globs]] field.
+
+### Key `exclude_globs`
+
+A list of [=globs=] that should be used to exclude URLs from where the content script runs.
```
xeenon:
```suggestion
A list of [=globs=] that can be used to exclude URLs from where the content script runs.
```

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
 # Version number handling
+
+# Algorithms
+
+## Determine the URL for content script matching
+
+To determine the URL to use for a document when injecting a content script:
+
+1. Let |url| be the document's URL.
+1. If |url| is `about:blank` or `about:srcdoc`, and `match_about_blank` or `match_origin_as_fallback` is set to true:
```
xeenon:
```suggestion
1. If |url| is `about:blank` or `about:srcdoc`, and `match_about_blank` or `match_origin_as_fallback` is set to `true`:
```

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
+    1. Set |url| to a URL based on the precursor origin.
+1. If the [=scheme=] of |url| is `data` and `match_origin_as_fallback` is set to true:
```
xeenon:
```suggestion
1. If the [=scheme=] of |url| is `data` and `match_origin_as_fallback` is set to `true`:
```

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
+    1. Set |url| to be a URL based on the precursor origin.
+1. If the [=scheme=] of |url| is `filesystem` or `blob` and `match_origin_as_fallback` is set to true:
```
xeenon:
```suggestion
1. If the [=scheme=] of |url| is `filesystem` or `blob` and `match_origin_as_fallback` is set to `true`:
```

xeenon (approved):

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
+To determine the URL to use for a document when injecting a content script:
+
+1. Let |url| be the document's URL.
+1. If |url| is `about:blank` or `about:srcdoc`, and `match_about_blank` or `match_origin_as_fallback` is set to `true`:
+    1. Set |url| to a URL based on the precursor origin.
+1. If the [=scheme=] of |url| is `data` and `match_origin_as_fallback` is set to `true`:
+    1. Set |url| to be a URL based on the precursor origin.
+1. If the [=scheme=] of |url| is `filesystem` or `blob` and `match_origin_as_fallback` is set to `true`:
+    1. Set |url| to be a URL based on the precursor origin.
+1. Return |url|.
```
Rob--W:
```suggestion
## Determine the URL for matching a document

To determine the URL to use for matching a document, given the document, `match_origin_as_fallback` and `match_about_blank`:

1. Let |url| be the document's URL.
1. If the [=scheme=] of |url| is `http`, `https` or `file`:
    1. Return |url|.
1. If the [=scheme=] of |url| is `blob`, `data` or `filesystem`, or if |url| is `about:blank` or `about:srcdoc`:
    1. If `match_origin_as_fallback` is set to `true`:
        1. If the document's origin is a [=tuple origin=]:
            1. Let |document-origin| be the serialization of the document's origin.
            1. If the [=scheme=] of |document-origin| is `http`, `https` or `file`:
                1. Return |document-origin|.
            1. Else, return null.
        *. NOTE: If not a [=tuple origin=], the document’s origin is an [=opaque origin=].
        1. Let |precursor-origin| be the serialization of the document’s precursor origin, if any.
            1. If the [=scheme=] of |precursor-origin| is `http`, `https` or `file`:
                1. Return |precursor-origin|.
            1. Else, return null.
    1. Else, if `match_about_blank` is set to `true`:
        1. If |url| is `about:blank` or `about:srcdoc`:
            1. Let |opener| be the active document of document’s [=opener browsing context=].
            1. If all of the following conditions are true:
                - |opener| is not null
                - |opener|’s origin is still the same as the document’s [=opener origin at creation=]
                - The algorithm has not been repeated for |opener| yet.
                1. Then repeat the algorithm for |opener|.
1. Return null.
```

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
+
+## Inject a content script
+
+Issue: If the same extension specifies the same script twice, what should happen? ([bug](https://crbug.com/324096753))
+
+To determine if a content script should be injected in a document:
+
+1. Let |url| be the result of running [[#determine-the-url-for-content-script-matching]].
```
Rob--W:
```suggestion
1. Let |url| be the result of running [[#determine-the-url-for-matching-a-document]].
```

specification/index.bs
```diff
@@ -203,3 +283,32 @@ Issue(62): Specify localization handling.
...
+1. If the extension does not have access to the origin, return.
```
Rob--W:
```suggestion
1. If the extension does not have access to |url|, return.
```

Rob--W (commented):
@oliverdunk and I sat together and drafted an updated algorithm that more closely resembles the behavior of Chrome and Firefox. (not Safari because it does not support `match_about_blank` yet).

We still need to define how to get a precursor origin somehow; ideally that would be in a different spec.

specification/index.bs
```diff
@@ -203,3 +283,48 @@ Issue(62): Specify localization handling.
...
+    1. Return |url|.
+1. If the [=scheme=] of |url| is `blob`, `data` or `filesystem`, or if |url| is `about:blank` or `about:srcdoc`:
+    1. If `match_origin_as_fallback` is set to `true`:
+        1. If the document's origin is a [=tuple origin=]:
+            1. Let |document-origin| be the serialization of the document's origin.
+            1. If the [=scheme=] of |document-origin| is `http`, `https` or `file`:
+                1. Return |document-origin|.
+            1. Else, return null.
+        *. NOTE: If not a [=tuple origin=], the document’s origin is an [=opaque origin=].
+        1. Let |precursor-origin| be the serialization of the document’s precursor origin, if any.
```
Rob--W:
```suggestion
        1. Let |precursor-origin| be the serialization of the document’s precursor origin, if any.
            *. TODO: "precursor origin" concept needs to be specified. It is not in the HTML spec at the moment. At least Chrome and Firefox recognize the concept, see e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1715167
```

Rob--W (approved):
Approving to signal that I'm content with the technical accuracy of the document. What remains before merging is formatting so that the spec can render as expected.

Would also be nice to have the "precursor origin" specified somewhere (ideally somewhere not extension-specific because it is a generic concept), but that is not a blocker to merging.

specification/index.bs
oliverdunk: I added this in https://github.com/w3c/webextensions/pull/542/commits/1a40c6569cc81bcad4e3f0170250322f30458d1e.

oliverdunk:
I've updated the formatting following Rob's comment. While CI is passing on this PR, it seems like the preview URL still isn't getting updated. You can find the latest version here: https://github.com/w3c/webextensions/actions/runs/13334183668/artifacts/2594066705

specification/index.bs
```diff
@@ -154,6 +155,8 @@ Filenames beginning with an underscore (`_`) are reserved for use by user agent.
 
 # Isolated worlds
 
+<dfn>Worlds</dfn> are isolated JavaScript contexts with access to the same underlying DOM tree but their own set of wrappers around those DOM objects.
```
rdcronin: maybe referece the fact that the entire variable space is isolated?  (It's not just the DOM objects)

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
-### Isolated worlds
+<dfn>Content scripts</dfn> represent a set of JS and CSS files that should be injected into matching pages loaded by the user agent. They are injected using the steps in [[#inject-a-content-script]].
+
+### Key `matches`
+
+A [=list=] of [=match patterns=] that are used to decide which pages the user agent injects the content script into. This key is required.
+
+### Key `exclude_matches`
+
+A [=list=] of [=match patterns=] that can be used to exclude URLs from where the content script runs.
```
rdcronin:
```suggestion
A [=list=] of [=match patterns=] that can be used to specify URLs where the content script should not run, even if the URL matches an entry in [[#key-matches]].
```

Feel free to wordsmith

specification/index.bs
```diff
@@ -190,7 +199,78 @@ Issue(62): Specify localization handling.
...
+
+Specifies when the content script should be injected. Valid values are defined by the {{RunAt}} enum.
+
+### Key `include_globs`
+
+A list of [=globs=] that a document should match. A document matches if the URL matches both the [[#key-matches]] field and the [[#key-include_globs]] field.
+
+### Key `exclude_globs`
+
+A list of [=globs=] that can be used to exclude URLs from where the content script runs.
```
rdcronin: similar comment here (mostly, "exclude URLs from where the content script runs" sounds awkward to me)

rdcronin (approved):
LGTM; thank you!

PR #546, opened 2024-02-14
update window.browser spec per #532

### LINKED ISSUE #532: Determine the nuances of aliasing `chrome` and `browser`
Now that #508 has been merged (🎉), our team began discussing the specifics on _how_ to alias `chrome` to `browser`. It turns out, what we want to do is different from everyone, but also everyone is different from everyone.

In Firefox, `chrome` and `browser` appear to be clones of the same object. So the values are the same, but updating one does not change another.

```js
browser.foo = 123
console.log(browser.foo)
123
console.log(chrome.foo)
undefined
```

In Safari, chrome and browser appear to be a direct alias to one another

```js
browser.foo = 123
console.log(browser.foo)
123
console.log(chrome.foo)
123
```

What Chrome wants to do is make `chrome` and `browser` separate objects, but make the various properties of each object a direct alias.

So in the above example, the Firefox behavior would be matched

```js
browser.foo = 123
console.log(browser.foo)
123
console.log(chrome.foo)
undefined
```

But modifying a specific property or value _inside_ of `chrome` or `browser` _would_ update the other.

```js
browser.runtime.PlatformArch.ARM = 123
console.log(browser.runtime.PlatformArch.ARM)
123
console.log(chrome.runtime.PlatformArch.ARM)
123
```

The reason why we like this option is the mix of ease of implementation from aliasing, while also allowing us to sluff off APIs we do not want to carry forward into `browser`.

patrickkettner:
Fixes #532

### DIFF
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -70,7 +70,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         globalThis.browser.FAKE = true
         console.log(globalThis.chrome.FAKE); //undefined
 
-        // Individual properies are aliased, so updating a known property
+        // Individual properties are aliased, so updating a known property
         // directly changes the matching property on the other
         globalThis.browser.runtime.FAKE = 123
 
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -70,7 +70,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         // modifying one top level values will not change the other
 
         globalThis.browser.FAKE = true
-        console.log(globalThis.chrome.FAKE); //undefined
+        console.log(globalThis.chrome.FAKE);
+        // ^ undefined when chrome and browser are not direct aliases.
+        // ^ may be true if browser === chrome.
 
         // Individual properties are aliased, so updating a known property
         // directly changes the matching property on the other
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     it MUST be used only for WebExtension related functionality.
   </section>
 
+  <section>
+    For backwards compatibility with older versions of web extensions, a UA MAY
+    alias individual properties of {{browser}} to <code>chrome</code>. {{browser}}
+    and <code>chrome</code> SHOULD NOT be direct aliases. Instead, individual
+    subproperties SHOULD be what is alaised. Any modification of a property of
+    {{browser}} SHOULD be reflected on <code>chrome</code>, and vice versa.
+
+    <div class="example" id="example-1">
+      Here's an example of the expected alias behavior
+      <pre><code class="lang-js">
+        // As `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other
+
+        globalThis.browser.FAKE = true
+
+        console.log(globalThis.chrome.FAKE);
+        // undefined
+
+        // Individual properies are aliased, so updating a known property
+        // directly changes the existing property on the other
+        globalThis.browser.runtime.FAKE = 123
+
+        console.log(globalThis.chrome.runtime.FAKE);
+        // 123
+      </code></pre>
+
+    </div>
+
+
+  </section>
+
+/e
+
 
 <pre class="idl">
   partial interface Window {
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -57,10 +57,11 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     <code>chrome</code> as a global identifier for WebExtension related methods,
     a UA MAY define a global <code>chrome</code> object that includes all
     properties of {{browser}}.
-    {{browser}} and <code>chrome</code> SHOULD NOT be direct aliases. Instead,
-    individual subproperties SHOULD be alaised. Any modification of a property
-    of {{browser}} SHOULD be reflected on the same subproperty of <code>chrome</code>,
-    and vice versa.
+    {{browser}} and <code>chrome</code> MAY be direct aliases of each other.
+    When {{browser}} and <code>chrome</code> are distinct objects, each initial
+    property of {{browser}} SHOULD be defined on the <code>chrome</code> object,
+    and each property descriptor on the <code>chrome</code> object SHOULD return
+    the same value as the corresponding property on the {{browser}} object.
 
     <div class="example" id="example-1">
       Here's an example of the expected alias behavior
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -66,8 +66,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     <div class="example" id="example-1">
       Here's an example of the expected alias behavior
       <pre><code class="lang-js">
-        // As `chrome` and `browser` are not direct aliases,
-        // modifying one top level values will not change the other
+        // When `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other:
 
         globalThis.browser.FAKE = true
         console.log(globalThis.chrome.FAKE);
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -55,7 +55,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   <section>
     For backwards compatibility with older versions of WebExtensions that used
     <code>chrome</code> as a global identifier for WebExtension related methods,
-    a UA MAY alias individual properties of {{browser}} to <code>chrome</code>.
+    a UA MAY define a global <code>chrome</code> object that includes all
+    properties of {{browser}}.
     {{browser}} and <code>chrome</code> SHOULD NOT be direct aliases. Instead,
     individual subproperties SHOULD be alaised. Any modification of a property
     of {{browser}} SHOULD be reflected on the same subproperty of <code>chrome</code>,
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -53,11 +53,13 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   </section>
 
   <section>
-    For backwards compatibility with older versions of web extensions, a UA MAY
-    alias individual properties of {{browser}} to <code>chrome</code>. {{browser}}
-    and <code>chrome</code> SHOULD NOT be direct aliases. Instead, individual
-    subproperties SHOULD be what is alaised. Any modification of a property of
-    {{browser}} SHOULD be reflected on <code>chrome</code>, and vice versa.
+    For backwards compatibility with older versions of WebExtensions that used
+    <code>chrome</code> as a global identifier for WebExtension related methods,
+    a UA MAY alias individual properties of {{browser}} to <code>chrome</code>.
+    {{browser}} and <code>chrome</code> SHOULD NOT be direct aliases. Instead,
+    individual subproperties SHOULD be alaised. Any modification of a property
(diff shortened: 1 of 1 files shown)


xeenon (approved):

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+        console.log(globalThis.chrome.runtime.FAKE);
+        // 123
+      </code></pre>
+
+    </div>
+
+
+  </section>
+
+/e
```
oliverdunk: Nit: Typo

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     it MUST be used only for WebExtension related functionality.
   </section>
 
+  <section>
+    For backwards compatibility with older versions of web extensions, a UA MAY
```
oliverdunk: Should we replace web extensions with WebExtensions to be consistent with the other usages in this document?

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    alias individual properties of {{browser}} to <code>chrome</code>. {{browser}}
```
oliverdunk: Can we say something like "to an exposed global with the identifier chrome"? Since we haven't defined chrome here and it is unclear what it is.

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    and <code>chrome</code> SHOULD NOT be direct aliases. Instead, individual
+    subproperties SHOULD be what is alaised. Any modification of a property of
```
oliverdunk: Can we shorten this to "SHOULD be aliased"?

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    {{browser}} SHOULD be reflected on <code>chrome</code>, and vice versa.
```
oliverdunk: Should we remove this last line? I read it as "if I modify a property on browser, it should be reflected on chrome", i.e exactly what we don't want to support. Or we could try to clarify it, but given the detailed example below, I'm not sure we need to.

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+
+    <div class="example" id="example-1">
+      Here's an example of the expected alias behavior
+      <pre><code class="lang-js">
+        // As `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other
+
+        globalThis.browser.FAKE = true
```
oliverdunk, line 72: I'd personally prefer more concise formatting like:

```
globalThis.browser.FAKE = true;
console.log(globalThis.chrome.FAKE); // undefined
```

specification/window.browser.bs
```diff
@@ -52,6 +52,39 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+
+        console.log(globalThis.chrome.FAKE);
+        // undefined
+
+        // Individual properies are aliased, so updating a known property
+        // directly changes the existing property on the other
```
oliverdunk: "changes the matching property" maybe? Since this makes it sound like it was only existing on the other one, but really it exists on both.

oliverdunk (changes requested):
Thanks Patrick! I left a review with some wording suggestions. I know there was some concern in the last meeting about if it made sense to do this (and I'm still not fully sure) so I think we should get approval from @Rob--W and others before deciding if we merge.

specification/window.browser.bs
patrickkettner: yep!

specification/window.browser.bs
patrickkettner: It is mentioned in the non normative portion at the beginning. But happy to add that

specification/window.browser.bs
patrickkettner: sure

specification/window.browser.bs
patrickkettner: sure

specification/window.browser.bs
patrickkettner: sgtm

patrickkettner:
@oliverdunk thanks for the review! I believe I addressed everything in the last update, so another look at some point would be appreciated! 

@Rob--W I'd love any feedback you have

specification/window.browser.bs
```diff
@@ -52,6 +52,36 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    <div class="example" id="example-1">
+      Here's an example of the expected alias behavior
+      <pre><code class="lang-js">
+        // As `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other
+
+        globalThis.browser.FAKE = true
+        console.log(globalThis.chrome.FAKE); //undefined
+
+        // Individual properies are aliased, so updating a known property
```
Rob--W:
```suggestion
        // Individual properties are aliased, so updating a known property
```

specification/window.browser.bs
```diff
@@ -52,6 +52,36 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    {{browser}} and <code>chrome</code> SHOULD NOT be direct aliases. Instead,
+    individual subproperties SHOULD be alaised. Any modification of a property
+    of {{browser}} SHOULD be reflected on the same subproperty of <code>chrome</code>,
+    and vice versa.
+
+    <div class="example" id="example-1">
+      Here's an example of the expected alias behavior
+      <pre><code class="lang-js">
+        // As `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other
```
Rob--W: Let's add a period to sentences (also below), and semicolons at line 70 and 75. The code works without it due to ASI, but for clarity let's just use semicolons.

specification/window.browser.bs
```diff
@@ -52,6 +52,36 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     it MUST be used only for WebExtension related functionality.
   </section>
 
+  <section>
+    For backwards compatibility with older versions of WebExtensions that used
+    <code>chrome</code> as a global identifier for WebExtension related methods,
+    a UA MAY alias individual properties of {{browser}} to <code>chrome</code>.
```
Rob--W: Backward support does not necessarily imply that properties of the namespaces are aliases of each other. Rather, it implies that there may be two globals that basically offer equivalent functionality. So I suggest to rephrase the current sentence to:

```suggestion
    a UA MAY define a global <code>chrome</code> object that includes all
    properties of {{browser}}.
```

specification/window.browser.bs
```diff
@@ -52,6 +52,36 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    {{browser}} and <code>chrome</code> SHOULD NOT be direct aliases. Instead,
+    individual subproperties SHOULD be alaised. Any modification of a property
+    of {{browser}} SHOULD be reflected on the same subproperty of <code>chrome</code>,
+    and vice versa.
```
Rob--W: I thought that we were going to specify that they may be direct aliases. Here is a different phrasing that captures this, and also defines the relation more precisely:

```suggestion
    {{browser}} and <code>chrome</code> MAY be direct aliases of each other.
    When {{browser}} and <code>chrome</code> are distinct objects, each initial
    property of {{browser}} SHOULD be defined on the <code>chrome</code> object,
    and each property descriptor on the <code>chrome</code> object SHOULD return
    the same value as the corresponding property on the {{browser}} object.
```

The original wording said "Any modification of a property of {{browser}} SHOULD be reflected on the same subproperty of <code>chrome</code> and vice versa." Besides the direct value alias, this would (unintentionally?) also imply:

- `browser.tabs = 1; assert(chrome.tabs === 1);` (the tabs property has been modified)
- `browser.whatever = 2; assert(chrome.whatever === 2);` ("Any modification of a property" does not say that user-defined properties aren't mirrored)
- `chrome.csi = 2; assert(browser.csi === 2);`  ("Any modification of a property" + "and vice versa" implies that `chrome` and `browser` should share the same set of properties?)

Rob--W (commented):
Besides the review, another thought related to your comment at https://github.com/w3c/webextensions/issues/532#issuecomment-1922063924

> Whether or not `chrome.foo = "bar"` would automatically make `browser.foo === "bar"` is left up to the user agent to decide. (With our proposal, this would be false. My understanding was that Firefox was fine with changing to match behavior, but Safari was not interested in changing their strict aliasing as they didn't have anything on `chrome` they didn't also want on `browser`)

The thought occurred to be that there may be little point in distinguishing the `browser` and `chrome` namespaces. Any existing property of `chrome` is de facto reserved, because if another browser were to introduce a new API under `browser.loadTimes`, for example, then developers would be surprised when they use `chrome.loadTimes` with the expectation of `browser.loadTimes` pointing to the same object. `browser` as a namespace explicitly permits UA-specific additions, which is good enough to allow Chrome to expose `chrome.csi` as `browser.csi`, but as a browser-specific "extension" API. Thoughts on this one?

xeenon (dismissed):

patrickkettner:
> The thought occurred to be that there may be little point in distinguishing the browser and chrome namespaces

@Rob--W My thought process was that a developer may be surprised to see that modifying `browser.foo` has different effect to `chrome.foo` in different browsers. Since testing showed that to be true, it makes sense to attempt to standardize behavior to remove that surprise. Since we all have unique behavior (if you count our plan as a behavior), there would be some kind of change somewhere for that behavior to be consistent.
If no one plans on changing their behavior, then I agree that there is no point in specing out the behavior, since it is just saying "anyone can do whatever they want". But if we want to align on expectations, then it makes sense to me to write down what those expectations are.

specification/window.browser.bs
```diff
@@ -52,6 +52,38 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+    When {{browser}} and <code>chrome</code> are distinct objects, each initial
+    property of {{browser}} SHOULD be defined on the <code>chrome</code> object,
+    and each property descriptor on the <code>chrome</code> object SHOULD return
+    the same value as the corresponding property on the {{browser}} object.
+
+    <div class="example" id="example-1">
+      Here's an example of the expected alias behavior
+      <pre><code class="lang-js">
+        // As `chrome` and `browser` are not direct aliases,
+        // modifying one top level values will not change the other
```
Rob--W:
```suggestion
        // When `chrome` and `browser` are not direct aliases,
        // modifying one top level values will not change the other:
```

Rephrased to avoid the impression that chrome and browser couldn't be direct aliases.

specification/window.browser.bs
```diff
@@ -52,6 +52,38 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
...
+
+        globalThis.browser.FAKE = true
+        console.log(globalThis.chrome.FAKE); //undefined
```
Rob--W:
```suggestion
        console.log(globalThis.chrome.FAKE);
        // ^ undefined when chrome and browser are not direct aliases.
        // ^ may be true if browser === chrome.
```

Rob--W (approved):
Since Chrome is unwilling to have `chrome === browser` that Safari and Firefox already supports, we will spec the "not direct alias" behavior.

oliverdunk (approved):

Rob--W (approved):
@xeenon Does this look good to you too?

xeenon (approved):

Issue #279, opened 2022-09-20
User scripts in Manifest V3


dotproto:
_PLACEHOLDER: We have a couple of issues related to user scripts support in Manifest V3, but nothing that currently feels like a good tracking issue. This issue intends to fill that gap by consolidating discussion of Manifest V3 API considerations in one place._

Rob--W:
@dotproto (on behalf of Google / Chrome) announced the intent to support user script managers in the WECG meeting at TPAC (#232). The meeting notes have been merged in #277 and this topic can be found under "User script manager support in MV3" at https://github.com/w3c/webextensions/blob/main/_minutes/2022-09-15-wecg-tpac.md

I (on behalf of Mozilla / Firefox) previously stated that we'd like to support user script managers, ideally in a more secure way, at https://blog.mozilla.org/addons/2022/05/18/manifest-v3-in-firefox-recap-next-steps/#comment-227473.

EDIT: The API design is ongoing. Significant updates are listed here:

- https://github.com/w3c/webextensions/issues/279#issuecomment-1255843961 by @Rob--W (2022-09-23)
   Summary of Google's "code" proposal + Mozilla's proposal for a secure "userscript" "world".

- https://github.com/w3c/webextensions/issues/279#issuecomment-1302762663 by @EmiliaPaz (2022-11-03)
  Google's API proposal document, "User Script Support in Scripting API".

- https://github.com/w3c/webextensions/issues/279#issuecomment-1309614401 by @Rob--W (2022-11-10)
  Analysis of the requirements for User script managers + proposal to cover cross-world communication.

- https://github.com/w3c/webextensions/issues/279#issuecomment-1326736211 (2022-11-24)
  Meeting notes and summary of the 2022-11-15 user script kickoff meeting.

- https://github.com/w3c/webextensions/issues/279#issuecomment-1339843225 by @EmiliaPaz (2022-12-06)
   Google's updated API proposal, PR #331.

- https://github.com/w3c/webextensions/pull/331#discussion_r1048532828 Discussion with clarification of API requirements and renewed API design (2022-12-14 until 2023-01-12)
  - The full discussion is part of the thread connected to https://github.com/w3c/webextensions/pull/331#discussion_r1048532828. The permalinks below may be broken, so expand the discussion (multiple times) via Github's UI and search for the fragments listed below.
  - API requirements gathering starts at https://github.com/w3c/webextensions/pull/331#discussion_r1058641524 by @rdcronin
[...]

PR #560, opened 2024-03-08
Proposal: Multiple user script worlds

rdcronin:
This adds a new API proposal for supporting multiple user script worlds, allowing user script managers to better isolate individual user scripts when multiple may be injected on a given site.

This proposal follows the new API proposal process.

### DIFF
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -179,7 +179,7 @@ does not have a corresponding configuration, it uses the default user script
 world properties. Any existing worlds are not directly affected by
 `userScripts.configureWorld()` calls; however, the browser may revoke
 certain privileges (for instance, message calls from existing user script worlds
-may beging to fail if the extension sets `messaging` to false). This is inline
+may beging to fail if the extension sets `messaging` to false). This is in line
 with behavior extensions encounter when e.g. the extension is unloaded and the
 content script continues running.
 
@@ -253,13 +253,20 @@ No additional security considerations. This API may (mildly) increase security
 by (somewhat) isolating individual user scripts from interacting with one
 another.
 
+Note that since isolated world boundaries are not considered "hard" boundaries,
+we would not consider this isolation a security measure from the browser's
+security model (just as we don't consider isolated worlds to be a strict
+security boundary today).
+
 ## Alternatives
 
 ### Existing Workarounds
 
 Today, user script managers go to various lengths to try to isolate behavior
-between user scripts. This includes having various frozen types, making heavy
-use of function closures, ensuring injection before untrusted code, and more.
+between user scripts (for more information, see conversation in
+[issue 279](https://github.com/w3c/webextensions/issues/279). This includes
+having various frozen types, making heavy use of function closures, ensuring
+injection before untrusted code, and more.
 
 These workarounds are fragile and require significantly more complex code in
 user script managers, while providing a lesser guarantee of isolation than a
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -56,7 +56,7 @@ Relevant methods and types:
 ```diff
  export namespace userScripts {
    export interface WorldProperties {
-     /**
++    /**
 +     * Specifies the ID of the specific user script world to update.
 +     * If not provided, updates the properties of the default user script
 +     * world.
@@ -150,9 +150,12 @@ Relevant methods and types:
    ...
 
 +   /**
-+    * Removes the configuration for a given world.
++    * Resets the configuration for a given world. Any scripts that inject into
++    * the world with the specified ID will use the default world configuration.
++    * Does nothing (but does not throw an error) if provided a `worldId` that
++    * does not correspond to a current configuration.
 +    */
-+   export function removeWorld(worldId: string): Promise<void>;
++   export function resetWorldConfiguration(worldId: string): Promise<void>;
 +
 +   /**
 +    * Returns a promise that resolves to an array of the the configurations
@@ -173,10 +176,15 @@ Worlds may be configured via `userScripts.configureWorld()` by indicating the
 given `worldId`. User scripts injected into a world with the given `worldId`
 will have the associated properties from the world configuration. If a world
 does not have a corresponding configuration, it uses the default user script
-world properties.
+world properties. Any existing worlds are not directly affected by
+`userScripts.configureWorld()` calls; however, the browser may revoke
+certain privileges (for instance, message calls from existing user script worlds
+may beging to fail if the extension sets `messaging` to false). This is inline
+with behavior extensions encounter when e.g. the extension is unloaded and the
+content script continues running.
 
-World configurations can be removed via the new `userScripts.removeWorld()`
-method.
+World configurations can be removed via the new
+`userScripts.resetWorldConfiguration()` method.
 
 Additionally, `runtime.Port` and `runtime.MessageSender` will each be extended
 with a new, optional `userScriptWorldId` property that will be populated in the
@@ -193,7 +201,7 @@ a user script).
 
 Configured worlds are persisted until the owning extension is updated (in order
 to align with the behavior of the rest of the `userScripts` API) or manually
-removes them via the new `userScripts.removeWorld()` method.
+removes them via the new `userScripts.resetWorldConfiguration()` method.
 
 #### World Limits
 
@@ -202,9 +210,9 @@ number of worlds that may be active in a given document. These limits may be
 different (since an extension may want more individual world configurations than
 they would expect to be practically encountered on a given site).
 
-If an extension reaches the limit of the number of registered worlds, attempts
-to register a new world via `userScripts.configureWorld()` will fail with an
-error.
+If an extension reaches the limit of the number of registered world
+configurations, attempts to register a new world via
+`userScripts.configureWorld()` will fail with an error.
 
 If an extension tries to inject more scripts into a single document than the
 per-document limit, all additional scripts will be injected into the default
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -146,16 +146,37 @@ Relevant methods and types:
 +     */
 +    worldId?: string;
    }
+
+   ...
+
++   /**
++    * Removes the configuration for a given world.
++    */
++   export function removeWorld(worldId: string): Promise<void>;
++
++   /**
++    * Returns a promise that resolves to an array of the the configurations
++    * for all user script worlds.
++    */
++   export function getWorldConfigurations(): Promise<WorldProperties[]>;
  }
 ```
 
 Note that the signatures of the related functions, including `configureWorld()`,
 `register()`, and others are left unchanged.
 
-When the developer specifies a `worldId` in either the `WorldProperties` or
-a `RegisteredUserScript`, the browser will create a separate user script world
-for those cases. If `worldId` is omitted, the default user script world will
-be used.
+When the developer specifies a `RegisteredUserScript`, the browser will use a
+separate user script world when injecting the scripts into a document. If
+`worldId` is omitted, the default user script world will be used.
+
+Worlds may be configured via `userScripts.configureWorld()` by indicating the
+given `worldId`. User scripts injected into a world with the given `worldId`
+will have the associated properties from the world configuration. If a world
+does not have a corresponding configuration, it uses the default user script
+world properties.
+
+World configurations can be removed via the new `userScripts.removeWorld()`
+method.
 
 Additionally, `runtime.Port` and `runtime.MessageSender` will each be extended
 with a new, optional `userScriptWorldId` property that will be populated in the
@@ -165,6 +186,29 @@ script world. We do not need to populate a value for messages from the default
 user script world, since this is unambiguous given the distinction in the
 event listeners (`onUserScriptMessage` already indicates the message was from
 a user script).
+
+### Behavioral Notes
+
+#### World Persistence
+
+Configured worlds are persisted until the owning extension is updated (in order
+to align with the behavior of the rest of the `userScripts` API) or manually
+removes them via the new `userScripts.removeWorld()` method.
+
+#### World Limits
+
+User agents may place limits on both the number of registered worlds and the
+number of worlds that may be active in a given document. These limits may be
+different (since an extension may want more individual world configurations than
+they would expect to be practically encountered on a given site).
+
+If an extension reaches the limit of the number of registered worlds, attempts
+to register a new world via `userScripts.configureWorld()` will fail with an
+error.
+
+If an extension tries to inject more scripts into a single document than the
+per-document limit, all additional scripts will be injected into the default
+world.
 
 ### New Permissions
 
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -264,7 +264,7 @@ security boundary today).
 
 Today, user script managers go to various lengths to try to isolate behavior
 between user scripts (for more information, see conversation in
-[issue 279](https://github.com/w3c/webextensions/issues/279). This includes
+[issue 279](https://github.com/w3c/webextensions/issues/279)). This includes
 having various frozen types, making heavy use of function closures, ensuring
 injection before untrusted code, and more.
 
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -133,102 +133,102 @@ export namespace userScripts {
 ```
 
 We will add a new property, `worldId`, to the `WorldProperties` and
-`RegisteredUserScript` types, as below. Changed fields are indicated with
-"+".
+`RegisteredUserScript` types, as below.
 
-```
-export namespace userScripts {
-  export interface WorldProperties {
-    /**
-+    * Specifies the ID of the specific user script world to update.
-+    * If not provided, updates the properties of the default user script world.
-+    * Values with leading underscores (`_`) are reserved.
-+    */
-+   worldId?: string;
+```diff
+ export namespace userScripts {
+   export interface WorldProperties {
+     /**
++     * Specifies the ID of the specific user script world to update.
++     * If not provided, updates the properties of the default user script
++     * world.
++     * Values with leading underscores (`_`) are reserved.
++     */
++    worldId?: string;
 
-    /**
-     * Specifies the world's CSP. The default is the `ISOLATED` world CSP.
-     */
-    csp?: string;
+     /**
+      * Specifies the world's CSP. The default is the `ISOLATED` world CSP.
+      */
+     csp?: string;
 
-    /**
-     * Specifies whether messaging APIs are exposed. The default is `false`.
-     */
-    messaging?: boolean;
-  }
+     /**
+      * Specifies whether messaging APIs are exposed. The default is `false`.
+      */
+     messaging?: boolean;
+   }
 
-  /**
-   * Properties for a registered user script.
-   */
-  export interface RegisteredUserScript {
-    /**
-     * If true, it will inject into all frames, even if the frame is not the
-     * top-most frame in the tab. Each frame is checked independently for URL
-     * requirements; it will not inject into child frames if the URL
-     * requirements are not met. Defaults to false, meaning that only the top
-     * frame is matched.
-     */
-    allFrames?: boolean;
+   /**
+    * Properties for a registered user script.
+    */
+   export interface RegisteredUserScript {
+     /**
+      * If true, it will inject into all frames, even if the frame is not the
+      * top-most frame in the tab. Each frame is checked independently for URL
+      * requirements; it will not inject into child frames if the URL
+      * requirements are not met. Defaults to false, meaning that only the top
+      * frame is matched.
+      */
+     allFrames?: boolean;
 
-    /**
-     * Specifies wildcard patterns for pages this user script will NOT be
-     * injected into.
-     */
-    excludeGlobs?: string[];
+     /**
+      * Specifies wildcard patterns for pages this user script will NOT be
+      * injected into.
+      */
+     excludeGlobs?: string[];
 
-    /**
-     * Excludes pages that this user script would otherwise be injected into.
-     * See Match Patterns for more details on the syntax of these strings.
-     */
-    excludeMatches?: string[];
+     /**
+      * Excludes pages that this user script would otherwise be injected into.
+      * See Match Patterns for more details on the syntax of these strings.
+      */
+     excludeMatches?: string[];
 
-    /**
-     * The ID of the user script specified in the API call. This property
-     * must not start with a '_' as it's reserved as a prefix for generated
-     * script IDs.
-     */
-    id: string;
+     /**
+      * The ID of the user script specified in the API call. This property
+      * must not start with a '_' as it's reserved as a prefix for generated
+      * script IDs.
+      */
+     id: string;
 
-    /**
-     * Specifies wildcard patterns for pages this user script will be
-     * injected into.
-     */
-    includeGlobs?: string[];
+     /**
+      * Specifies wildcard patterns for pages this user script will be
+      * injected into.
+      */
+     includeGlobs?: string[];
 
-    /**
-     * The list of ScriptSource objects defining sources of scripts to be
-     * injected into matching pages.
-     */
-    js: ScriptSource[];
+     /**
+      * The list of ScriptSource objects defining sources of scripts to be
+      * injected into matching pages.
+      */
+     js: ScriptSource[];
 
-    /**
-     * Specifies which pages this user script will be injected into. See
-     * Match Patterns for more details on the syntax of these strings. This
-     * property must be specified for ${ref:register}.
-     */
-    matches?: string[];
+     /**
+      * Specifies which pages this user script will be injected into. See
+      * Match Patterns for more details on the syntax of these strings. This
+      * property must be specified for ${ref:register}.
+      */
+     matches?: string[];
 
-    /**
-     * Specifies when JavaScript files are injected into the web page. The
-     * preferred and default value is document_idle
-     */
-    runAt?: RunAt;
+     /**
+      * Specifies when JavaScript files are injected into the web page. The
+      * preferred and default value is document_idle
+      */
+     runAt?: RunAt;
 
-    /**
-     * The JavaScript execution environment to run the script in. The default
-     * is `USER_SCRIPT`
-     */
-    world?: ExecutionWorld;
+     /**
+      * The JavaScript execution environment to run the script in. The default
+      * is `USER_SCRIPT`
+      */
+     world?: ExecutionWorld;
 
-+   /**
-+    * If specified, specifies a specific user script world ID to execute in.
-+    * Only valid if `world` is omitted or is `USER_SCRIPT`. If omitted, the
-+    * script will execute in the default user script world.
-+    * Values with leading underscores (`_`) are reserved.
-+    */
-+   worldId?: string;
-  }
-}
++    /**
++     * If specified, specifies a specific user script world ID to execute in.
++     * Only valid if `world` is omitted or is `USER_SCRIPT`. If omitted, the
++     * script will execute in the default user script world.
++     * Values with leading underscores (`_`) are reserved.
++     */
++    worldId?: string;
+   }
+ }
 ```
 
 Note that the signatures of the related functions, including `configureWorld()`,
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -15,7 +15,7 @@ Allow developers to configure and use multiple user script worlds in the
 
 **Created:** 2024-03-07
 
-**Related Issues:** TODO
+**Related Issues:** [565](https://github.com/w3c/webextensions/issues/565)
 
 ## Motivation
 
@@ -78,43 +78,51 @@ export namespace userScripts {
      * frame is matched.
      */
     allFrames?: boolean;
+
     /**
      * Specifies wildcard patterns for pages this user script will NOT be
      * injected into.
      */
     excludeGlobs?: string[];
+
     /**
      * Excludes pages that this user script would otherwise be injected into.
      * See Match Patterns for more details on the syntax of these strings.
      */
     excludeMatches?: string[];
+
     /**
      * The ID of the user script specified in the API call. This property
      * must not start with a '_' as it's reserved as a prefix for generated
      * script IDs.
      */
     id: string;
+
     /**
      * Specifies wildcard patterns for pages this user script will be
      * injected into.
      */
     includeGlobs?: string[];
+
     /**
      * The list of ScriptSource objects defining sources of scripts to be
      * injected into matching pages.
      */
     js: ScriptSource[];
+
     /**
      * Specifies which pages this user script will be injected into. See
      * Match Patterns for more details on the syntax of these strings. This
      * property must be specified for ${ref:register}.
      */
     matches?: string[];
+
     /**
      * Specifies when JavaScript files are injected into the web page. The
      * preferred and default value is document_idle
      */
     runAt?: RunAt;
+
     /**
      * The JavaScript execution environment to run the script in. The default
      * is `USER_SCRIPT`
@@ -125,7 +133,8 @@ export namespace userScripts {
 ```
 
 We will add a new property, `worldId`, to the `WorldProperties` and
-`RegisteredUserScript` types, as below.
+`RegisteredUserScript` types, as below. Changed fields are indicated with
+"+".
 
 ```
 export namespace userScripts {
@@ -133,11 +142,12 @@ export namespace userScripts {
     /**
 +    * Specifies the ID of the specific user script world to update.
 +    * If not provided, updates the properties of the default user script world.
++    * Values with leading underscores (`_`) are reserved.
 +    */
 +   worldId?: string;
 
     /**
-     * Specifies the world csp. The default is the `ISOLATED` world csp.
+     * Specifies the world's CSP. The default is the `ISOLATED` world CSP.
      */
     csp?: string;
 
@@ -159,43 +169,51 @@ export namespace userScripts {
      * frame is matched.
      */
     allFrames?: boolean;
+
     /**
      * Specifies wildcard patterns for pages this user script will NOT be
      * injected into.
      */
     excludeGlobs?: string[];
+
     /**
      * Excludes pages that this user script would otherwise be injected into.
      * See Match Patterns for more details on the syntax of these strings.
      */
     excludeMatches?: string[];
+
     /**
      * The ID of the user script specified in the API call. This property
      * must not start with a '_' as it's reserved as a prefix for generated
      * script IDs.
      */
     id: string;
+
     /**
      * Specifies wildcard patterns for pages this user script will be
      * injected into.
      */
     includeGlobs?: string[];
+
     /**
      * The list of ScriptSource objects defining sources of scripts to be
      * injected into matching pages.
      */
     js: ScriptSource[];
+
     /**
      * Specifies which pages this user script will be injected into. See
      * Match Patterns for more details on the syntax of these strings. This
      * property must be specified for ${ref:register}.
      */
     matches?: string[];
+
     /**
      * Specifies when JavaScript files are injected into the web page. The
      * preferred and default value is document_idle
      */
     runAt?: RunAt;
+
     /**
      * The JavaScript execution environment to run the script in. The default
      * is `USER_SCRIPT`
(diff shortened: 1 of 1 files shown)


proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Author:** rdcronin
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:** N/A
+
+**Created:** 2024-03-07
+
+**Related Issues:** TODO
```
rdcronin: @Rob--W I didn't see an existing bug for this.  Do you know if one is filed?

rdcronin:
@Rob--W @xeenon @mukul-p please take a look (or delegate) on behalf of the other browsers.

@EmiliaPaz , FYI, as this (somewhat) relates to your upcoming work on `userScripts.execute()`.

@tophf FYI, as well, given your interest and contribution in previous user script discussions.

proposals/multiple_user_script_worlds.md
Rob--W: Multiple world has been mentioned from the start in the userScripts API discussions at https://github.com/w3c/webextensions/issues/279

E.g. fragment from https://github.com/w3c/webextensions/issues/279#issuecomment-1255843961

> At present, I propose only one new world. This is strictly an improvement over the current state of art. Support for more than one world could be useful to extensions, but raises the complexity of implementation, so I'm leaving that for a follow-up and consider it to be out of scope in this proposal.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+## Specification
+
+### Schema
+
+This change would modify existing schemas.
+
+Relevant methods and types:
+
+```
+export namespace userScripts {
```
Rob--W: Instead of copy-pasting the schema and letting the reader eyeball-diff it, could you just specify the final schema, and mention what changed compared to before? And consider linking the original proposal in this repo.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+    excludeMatches?: string[];
+    /**
+     * The ID of the user script specified in the API call. This property
+     * must not start with a '_' as it's reserved as a prefix for generated
+     * script IDs.
+     */
+    id: string;
+    /**
+     * Specifies wildcard patterns for pages this user script will be
+     * injected into.
```
Rob--W: Although the behavior of includeGlobs/matches did not change, I'd like to emphasize that the original proposal explicitly called out the expected matching behavior at https://github.com/w3c/webextensions/blob/14ebd85ac034e5e4138d0653ae797a4aa9dc6703/proposals/user-scripts-api.md#L66-L67 upon my request, because the expected glob/matches behavior from user scripts differs from what has been implemented in the content script APIs.

Why is the original note omitted from the current "full" API description?

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
++    */
++   worldId?: string;
+  }
+}
+```
+
+Note that the signatures of the related functions, including `configureWorld()`,
+`register()`, and others are left unchanged.
+
+When the developer specifies a `worldId` in either the `WorldProperties` or
```
Rob--W: `WorldProperties` was not defined before. For clarity, can you list all methods that are affected by this change.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Implementation Notes
+
+N/A
+
+## Future Work
+
+With the addition of a `userScripts.execute()` function (as described in
+this [PR](https://github.com/w3c/webextensions/pull/540) and
+[issue](https://github.com/w3c/webextensions/issues/477), we should also
```
Rob--W:
```suggestion
[PR 540](https://github.com/w3c/webextensions/pull/540) and
[issue 477](https://github.com/w3c/webextensions/issues/477), we should also
```

Added number so that readers don't have to hover over the link to see what it is referring to.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+```
+
+Note that the signatures of the related functions, including `configureWorld()`,
+`register()`, and others are left unchanged.
+
+When the developer specifies a `worldId` in either the `WorldProperties` or
+a `RegisteredUserScript`, the browser will create a separate user script world
+for those cases. If `worldId` is omitted, a "default" user script world will
```
Rob--W:
```suggestion
for those cases. If `worldId` is omitted, the default user script world will
```

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+export namespace userScripts {
+  export interface WorldProperties {
+    /**
++    * Specifies the ID of the specific user script world to update.
++    * If not provided, updates the properties of the default user script world.
++    */
++   worldId?: string;
+
+    /**
+     * Specifies the world csp. The default is the `ISOLATED` world csp.
```
Rob--W:
```suggestion
     * Specifies the world's CSP. The default is the `ISOLATED` world's CSP.
```

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+### Open Web API
+
+The concept of user scripts should not be exposed to the web; this should not
+be an open web API.
+
+## Implementation Notes
+
+N/A
+
+## Future Work
```
Rob--W, line 284: Would we ever expose `worldId` to the content side? A use case that I can see is enabling (synchronous) cross-world communication or execution. I've wanted to create such a proposal for a while.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+     * is `USER_SCRIPT`
+     */
+    world?: ExecutionWorld;
+
++   /**
++    * If specified, specifies a specific user script world ID to execute in.
++    * Only valid if `world` is omitted or is `USER_SCRIPT`. If omitted, the
++    * script will execute in the default user script world.
++    */
++   worldId?: string;
```
Rob--W: Are there any constraints on the format of `worldId`? For example, if we reserve `_` for internal use, then we can choose default worldIds as follows.

- `world: "MAIN"` -> `worldId` defaults to `_MAIN_WORLD` and an error raised if anything else is passed.
-  `world: "USER_SCRIPT"` -> `worldId` defaults to `_USER_SCRIPT_WORLD` and can be set to anything else.

Another option is to use `""` as the default value, and permit any value in `worldId`.

Rob--W (changes requested):
In addition to the feedback below, I'd also like to point out that the `runtime.Port` and `runtime.MessageSender` API definitions should be extended with a `userScriptWorldId` property to enable extensions to distinguish user script worlds from each other.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * Specifies when JavaScript files are injected into the web page. The
+     * preferred and default value is document_idle
+     */
+    runAt?: RunAt;
+    /**
+     * The JavaScript execution environment to run the script in. The default
+     * is `USER_SCRIPT`
+     */
+    world?: ExecutionWorld;
```
xeenon: I filed #563 about this, but IMO we should use lowercase strings. This has recently changed with these newer APIs, while older APIs use lowercase strings — leading to an inconsistency.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+
+Relevant methods and types:
+
+```
+export namespace userScripts {
+  export interface WorldProperties {
+    /**
+     * Specifies the world csp. The default is the `` `ISOLATED` `` world csp.
+     */
+    csp?: string;
```
xeenon: Ditto regarding #563.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+  export interface RegisteredUserScript {
+    /**
+     * If true, it will inject into all frames, even if the frame is not the
+     * top-most frame in the tab. Each frame is checked independently for URL
+     * requirements; it will not inject into child frames if the URL
+     * requirements are not met. Defaults to false, meaning that only the top
+     * frame is matched.
+     */
+    allFrames?: boolean;
+    /**
```
xeenon: Consider adding empty lines before all these comments, you do in some places and it helps readability.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,285 @@
...
+     * Specifies when JavaScript files are injected into the web page. The
+     * preferred and default value is document_idle
+     */
+    runAt?: RunAt;
+    /**
+     * The JavaScript execution environment to run the script in. The default
+     * is `USER_SCRIPT`
+     */
+    world?: ExecutionWorld;
```
xeenon: Ditto regarding #563.

proposals/multiple_user_script_worlds.md
rdcronin: I think this would be better tracked individually (rather than lumped into the meta "user scripts in MV3" issue), so I've filed a new issue ( issue #565 ) for it, but referenced the original comment.

proposals/multiple_user_script_worlds.md
rdcronin: I prefer having the original schema so that, when looking back at these in <n> years, we can easily see what changed.

I already explicitly called out the changed lines with "+" indicators, but I've called out explicitly that I've done that.

proposals/multiple_user_script_worlds.md
rdcronin: Good question, but orthogonal from this issue.  I can update that separately in Chrome's repo.

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
rdcronin: I was debating that.  I don't want to surface "_" to developers (so wouldn't want them to have to pass  "_DEFAULT" or anything), but I'm fine saying _-prefixed values are reserved, allowing us to use these internally (if we want).

proposals/multiple_user_script_worlds.md
rdcronin: World properties were (and are) defined.  They allow the configuration of whether messaging APIs are exposed to the world.  In chrome, these are [here](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/user_scripts.idl;l=77-87;drc=0345221adefeb4f5f77cb57dd2dd1e8159b30762).

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
rdcronin: I'm not quite sure what you mean by "content side" -- do you mean to the user script?

In either case, I've indicated we may leverage the world ID for inter-world communication.

proposals/multiple_user_script_worlds.md
rdcronin: Good call; added.

proposals/multiple_user_script_worlds.md
rdcronin: Thanks for raising this.  I agree with aligning on a uniform standard, but would prefer using all caps for the reasons I highlighted in the response there.

I'd like to avoid blocking this issue on the resolution of that (it seems like it is very likely to have lots of impassioned bikeshed discussions : )); for the purpose of this proposal, I'd propose we keep the all caps since that aligns with the rest of the user script API, and we can establish what to do for other cases as part of issue #563  (since we'll need to address other APIs anyway).  Does that sound reasonable to you?

rdcronin (commented):
Thanks, Rob and Timothy!

@Rob--W , good callout on the new properties on runtime.Port and runtime.MessageSender; I've added those as well.  (Github won't let me reply to your comment there directly...)

rdcronin:
@Rob--W friendly ping : )

proposals/multiple_user_script_worlds.md
Rob--W: Markdown supports diff rendering, by putting "diff" after the triple backtick. Then when every line is indented with one space, a + instead of a space at the start renders as an addition and - as a removal.

```diff
 Line unchanged
+Line added
-Line removed
```

proposals/multiple_user_script_worlds.md
Rob--W: By "content" I meant in the renderer. Because doing so opens two things: firstly the ability to do interworld communication and two, enabling extensions to use it as a hack to encode a world-specific configuration in the string.

Since you included the mention of the first, the answer appears to be yes and I'll resolve this thread.

proposals/multiple_user_script_worlds.md
rdcronin: Ooh, that's rad!  Thanks for the tip; done.

With that, I think it's also clear to condense these into one section, so did that as well.

rdcronin:
@Rob--W , back to you!

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,234 @@
...
+
+This change would modify existing schemas. We will add a new property,
+`worldId`, to the `WorldProperties` and `RegisteredUserScript` types, as below.
+
+Relevant methods and types:
+
+```diff
+ export namespace userScripts {
+   export interface WorldProperties {
+     /**
```
Rob--W:
```suggestion
+    /**
```

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,234 @@
...
++    worldId?: string;
+   }
+ }
+```
+
+Note that the signatures of the related functions, including `configureWorld()`,
+`register()`, and others are left unchanged.
+
+When the developer specifies a `worldId` in either the `WorldProperties` or
+a `RegisteredUserScript`, the browser will create a separate user script world
```
Rob--W: Important design question: what are the persistence/lifetime characteristics of worldId + configureWorld? worldId should not persist across extension updates for consistency with the existing userScripts API design (https://developer.chrome.com/docs/extensions/reference/api/userScripts#extension_updates), but there is still an open question of the lifetime of worldId (and consequently, potentially reduce the duration of persistence).

Previously, `configureWorld` would have a limited scope, exactly one. Now, with the introduction of `worldId`, there may be an unbounded number. According to the current proposal, the concept of a specific world (identified by worldId) becomes concrete whenever it is used in `RegisteredUserScript` or `WorldProperties`. Because the number of worlds is now dynamic, we should clarify when a worldId exists and when it ceases to exist. As currently written, worldIds are instantiated and never discarded.

This section seems to have been written with `userScripts.register` in mind. The issue that I'm thinking of would have become much more obvious if the proposal explicitly spelled out the expected behavioral change (or non-change) per method (from the proposal process perspective - I think that it would be a good idea for future proposals to exhaustively enumerate all uses of a type to make sure that the behavior is fully specified where needed):
- `userScripts.register()` - may introduce a worldId. Does not necessarily have to do anything until the script is used in the renderer (unless there is a desire to broadcast all known worldIds?).
- `userScripts.configureWorld()` - may create a persistent configuration for a world. What if the worldId does not exist yet?
- `userScripts.getScripts()` - does this have to "create a separate user script world"?
- `userScripts.unregister()` - when the last use of `worldId` disappears, would that delete the configuration?
- `userScripts.update()` - when the `worldId` changes, and the previous one was the last one, does that delete the registered configured worlds?

I think that there are roughly two ways to address this issue (not mutually exclusive):

1. Explicit world management (`userScripts.registerWorld` + `userScripts.unregisterWorld`). Using an unregistered `worldId` in RegisteredUserScript or WorldProperties is an error.
2. Automatic world cleanup: a world configuration is guaranteed to persist when there is at least one RegisteredUserScript that references it. What happens when there is no reference to it is more interesting:
   - When there are no references, we could unregister the world as soon as the refcount drops to zero would be obvious (by `userScripts.unregister` or `userScripts.update`).
   - Do we want to support calling `userScripts.configureWorld` before `userScripts.register`? If so, then `configureWorld` should register a world implicitly, despite there not being any scripts that reference it.
     - to avoid accumulating junk, unused worlds should eventually be discarded. This is non-trivial.
     - an alternative is to enable the world to be configured along with registering a user script.
     - another alternative is to not support worldId in `configureWorld` unless the worldId has been allocated explicitly by `userScripts.register`. Extensions that really care about ensuring setup of the world before execution can register a dummy script with a worldId, then configure the world, then register the actual script for that world, and then unregister the dummy script. Is this an acceptable pattern?

Rob--W (commented):
I wanted to sign off, but upon doing a last pass where I examined the impact on individual API methods, I got one new question. Let's resolve that and then this should be good to merge.

proposals/multiple_user_script_worlds.md
rdcronin: Thanks, @Rob--W  !

I was thinking about most of these already (complete with a few TODOs, e.g. [here](https://chromium-review.googlesource.com/c/chromium/src/+/5408252/4/extensions/browser/api/user_scripts/user_scripts_api.cc#565)) and was planning on submitting an update to the proposal as we got nearer, but I hadn't considered the update behavior and allowing developers to clear out old registrations -- both of which I agree are valuable.

The way I'm thinking about worlds is that:
- A user script is registered with a world.  That determines where it injects.
- The configuration for that world may or may not exist, depending on whether the developer wants to use non-default values (for messaging and CSP, but potentially other APIs in the future).

As such, there isn't really the "creation" of a world when you register a script -- worlds are created in the document in which they run, and are created according to the configuration provided (if any).

I don't like the idea of either reference counting or requiring a user script be registered for every specified worldId -- that sounds very difficult for both us (the browser) and for developers to keep track of, and seems like it would very likely lead to some unpredictable or unexpected edge cases (e.g., a developer updates a user script world via unregister + register, and doesn't realize that unregistered the user script world).  It also leads to a (very small) race where it'd be impossible to ensure a user script injected with given world properties, since you'd have to register the script _before_ the configuring the world, but then the script may inject before that configuration took hold.  I'd much rather lean on the developers handling the registration.

I've expanded the proposal to include:
- Two new functions, `removeWorld()` and `getWorldConfigurations()`.
- Notes about how the worlds behave when injecting.
- Notes on world persistence and world limits.

Please take a look and let me know what you think.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,278 @@
...
++     */
++    worldId?: string;
+   }
+
+   ...
+
++   /**
++    * Removes the configuration for a given world.
++    */
++   export function removeWorld(worldId: string): Promise<void>;
```
Rob--W: "Remove world" gives the impression that the world ceases to exist. The way I see it, the concept of a world (identified by worldId) starts to exist whenever it is used by a script or config. A world is only removed when all scripts referencing the worldId are unregistered (AND the config removed too).

So I suggest renaming to something more obvious, such as one of the following:

- removeWorldConfiguration
- resetWorldConfiguration
- clearWorldConfiguration

And can you also clarify the behavior when the worldId doesn't have a special configuration? Especially with the reset or clear name, I would not expect errors. I would be fine with any defined behavior.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,278 @@
...
+
+When the developer specifies a `RegisteredUserScript`, the browser will use a
+separate user script world when injecting the scripts into a document. If
+`worldId` is omitted, the default user script world will be used.
+
+Worlds may be configured via `userScripts.configureWorld()` by indicating the
+given `worldId`. User scripts injected into a world with the given `worldId`
+will have the associated properties from the world configuration. If a world
+does not have a corresponding configuration, it uses the default user script
+world properties.
```
Rob--W: Should also add that configuration changes only apply to worlds/execution contexts created after that point. Existing worlds stick to the previous configuration.

If you envision different behavior, please speak up.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,278 @@
...
+removes them via the new `userScripts.removeWorld()` method.
+
+#### World Limits
+
+User agents may place limits on both the number of registered worlds and the
+number of worlds that may be active in a given document. These limits may be
+different (since an extension may want more individual world configurations than
+they would expect to be practically encountered on a given site).
+
+If an extension reaches the limit of the number of registered worlds, attempts
```
Rob--W: "registered worlds" -> "registered world configurations"

World registration is ambiguous because there are two distinct ways to have a concept of a world. And I think that you are trying to encourage extensions to unregister obsolete world configurations here.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,278 @@
...
+
+#### World Persistence
+
+Configured worlds are persisted until the owning extension is updated (in order
+to align with the behavior of the rest of the `userScripts` API) or manually
+removes them via the new `userScripts.removeWorld()` method.
+
+#### World Limits
+
+User agents may place limits on both the number of registered worlds and the
```
Rob--W, line 208: Should we really limit the number of worlds? What are the numbers you are thinking of?

A too low limit reduces the practical value of the multiple worlds concept, and user script managers are then still forced to implement script wrappers to avoid conflicts between user scripts.

A reasonably high limit reduces the impact for users with a few scripts, but punishes power users with lots of scripts. If a user is willing to install many scripts and considers the overhead acceptable, why would we stop them?

tophf: Assuming the world's overhead is ~1MB, how about 256 worlds max?

rdcronin: Short answer: yes, I strongly think we should limit the number of worlds.  In particular, I think this is true for the active number of worlds on a page.

The registration is relatively cheap, so I don't think we need to be _too_ limiting for total number of registered worlds (I was originally thinking ~100, but I think that one's more flexible).  The real cost is in the number of active worlds on pages.  Each of these results in a separate JS context which, depending on the size of the script, may (very) non-trivially increase the weight of a given document.  Then, given the number of documents that can be actively running, this can become _much_ heavier weight.  Every additional user script world is roughly equivalent to the cost of a whole other content script-injecting extension.  From @tophf 's suggestion of 256 worlds assuming a 1MB overhead, that's 256MB _per document_.  (Assuming that was meant as a suggestion for "active worlds on a page", rather than total registered worlds.  I think 256 total registered worlds is perhaps a bit high, but not unreasonable.)  I just opened cnn.com (my go-to for a "noisy" site), and there are 30 different frames there.  That means that, with 256 worlds at 1MB a piece, the user would have 7.5GB of memory used on _that one tab_ just from user scripts.  I think that's more than we'd allow in Chrome.

Now, since a lot of user scripts are only intended to run on specific pages, there are lots of cases when two scripts would never "collide" and result in multiple worlds on the same page, so I'm hopeful that most user script managers wouldn't run into this limit frequently.  But for all the scripts that inject on every page, I think there is a risk of major performance impact.

I also think the exact values here may be something left up to individual browser vendors to establish, as we've done with historical limitations with performance (in which case, Firefox may have no limit, Safari on iOS may establish a different limit than Safari on MacOS, etc).  But rather than fully leave out the expectations from the specification and then have an individual browser impose an intervention in ways that are unpredictable to the developer, I think it's worth describing what the behavior the developer can expect will be.

tophf: Chrome/Firefox similarly limit the number of OS processes they use and they combine sites in one process if necessary. Arguably extensions could set the maximum number of the worlds or at least read it from the API so they can display the warning to the user and add a "virtualizer" to the userscript code, specifically `with (globalProxy) {....}`, which is undesirable as `with` incurs overhead.

FWIW, my 1MB estimate might be too high. What if it's just 100kB?

tophf: I see 46kB for an extension `Window` in Chrome Devtools -> Memory -> Snapshot.

Roughly 20 bytes \* (1000 global objects + their 1000 prototypes).

rdcronin: Unfortunately, it's not _just_ the overhead of the JS items -- there's more that needs to happen to create a new isolated world (on the Chrome side, we need a new v8::Context, ScriptState, execution environment, etc).  The impact I'm seeing for a completely empty script injection is about 200KB per world when injecting into a very simple site (example.com).  Because the isolated world also has a copy of the DOM (the DOM is shared, but the JS objects for the DOM are not), this increases with the complexity of the page.

I am supportive of exposing the maximum number of worlds so that extensions can handle this appropriately (as @tophf called out, potentially showing user-visible UI or modifying their injections to be "semi-isolated" from one another).

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
tophf: So what happens to a matched userscript in the case its separate world cannot be created? There should be some event or another method of detecting it.

proposals/multiple_user_script_worlds.md
rdcronin: That's a good point on the naming; thanks, Rob.  "clearWorldConfiguration" SGTM, so I've tentatively incorporated that (if anyone has strong preferences for "reset", I'm good with that one, too).

proposals/multiple_user_script_worlds.md
rdcronin: This one is a bit tricky.

I don't think we'd:
- Update the CSP of existing worlds
- Manipulate JS objects in the existing worlds

However, I do think we _may_ prevent APIs from working if they get proxied through the browser process.  In particular, I think we may want to prevent messaging APIs from working (the browser may see that the user script world does not currently allow messaging, and prevent any message communication).  This seems desirable to me, and is also inline with the behavior we have for e.g. when the extension is unloaded but content scripts continue to run: the JS is unaffected, but anything reaching _outside_ that world (e.g. through the browser process) may be.

I've added this in.

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
rdcronin: This is called out in the proposal -- it will fallback to using the default world.

I'd be happy to consider:
- Having an additional trait in WorldProperties like "shouldFallBackIfWorldLimitReached" (perhaps with a better name) that controls whether the script should fall back or just bail and not inject, and / or
- Having a new event like "onUserScriptCouldNotInjectBecauseWorldLimitReached" (definitely with a better name)

However, given the scope creep we're seeing in this proposal, I'd like to do those in a separate change.

tophf: ...because the userscript manager extension would need to run the userscript's function inside `with(windowProxy)`

tophf: Okay, as long as it's not postponed to some distant future.

rdcronin: I think this can be a reasonably fast-follow.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,286 @@
...
+`worldId` is omitted, the default user script world will be used.
+
+Worlds may be configured via `userScripts.configureWorld()` by indicating the
+given `worldId`. User scripts injected into a world with the given `worldId`
+will have the associated properties from the world configuration. If a world
+does not have a corresponding configuration, it uses the default user script
+world properties. Any existing worlds are not directly affected by
+`userScripts.configureWorld()` calls; however, the browser may revoke
+certain privileges (for instance, message calls from existing user script worlds
+may beging to fail if the extension sets `messaging` to false). This is inline
```
Rob--W:
```suggestion
may beging to fail if the extension sets `messaging` to false). This is in line
```

proposals/multiple_user_script_worlds.md
Rob--W: I'm not resolving the thread yet because the expressed questions has not been resolved, but feel free to merge even if this thread is still open.

Consider calling out improvements in this area in "future work".

Rob--W (approved):
From my perspective, this proposal looks like it is in a good shape.

I would encourage solliciting feedback from developers of User Script managers, perhaps timeboxed to e.g. a week or two, to make sure that the API works for them without unsurmountable issues.

proposals/multiple_user_script_worlds.md
tophf: An event is necessary anyway, because managers need it to set up the script's environment before execution of any matched userscript e.g. set `window[randomId]` with the secure communication channel name in the MAIN world right right before execution so that no JS code can run in-between. Something like chrome.userScripts.onExec with (scriptId, worldId) parameters and chrome.runtime.lastError inside the listener.

Rob--W: Synchronous cross-world communication and code execution has been discussed before, including recently at the in-person WECG meetup. That will become a separate proposal, and does not need to be specific to user scripts.

tophf: My point was that userscripts need an event right before execution, it's not the same as cross-world communication. This event can be used to report the world creation error.

tophf: Same-world userscripts in the default world need it too to set up the internal bindings for `GM` stuff which is implemented in a common controller script. BTW we need chrome.userScripts.registerController for the code that runs before the first matched userscript in the tab and receives all matched userscript ids.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,286 @@
...
+
+No additional security considerations. This API may (mildly) increase security
+by (somewhat) isolating individual user scripts from interacting with one
+another.
+
+## Alternatives
+
+### Existing Workarounds
+
+Today, user script managers go to various lengths to try to isolate behavior
```
mukul-p, line 265: Adding references of these examples (stackoverflow link?) would "complete" the document. Assuming the proposal guideline is not against that. This is not a blocking comment.

mukul-p (approved):

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,286 @@
...
+injecting scripts into different web pages. A user may have many different
+user scripts that execute on the same page; however, these scripts are
+conceptually distinct from one another and should generally not interact with
+each other.
+
+Today, user script managers take various steps to isolate user scripts as they
+are injected, since all user scripts will run either in the context of the
+main world or in a single user script world for the extension. This API would
+allow user script managers to instead have multiple user script worlds, each
+isolated from one another, to reduce the likelihood of user scripts negatively
```
anforowicz, line 34: nit: Is it worth clarifying somewhere (in the paragraph here?  in the "Additional Security Considerations" section below?  in a separate doc?) that "isolated" is mostly in "hygiene" sense and not necessarily in "security" sense, because:

* Cross-world contact is possible (e.g. via `window`)
* OS process offers a more robust security boundary (e.g. Spectre allows cross-world reading of the whole process memory)

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,286 @@
...
+
+We anticipate user script managers such as Tampermonkey, Violentmonkey, and
+others to use this new capability.
+
+## Specification
+
+### Schema
+
+This change would modify existing schemas. We will add a new property,
+`worldId`, to the `WorldProperties` and `RegisteredUserScript` types, as below.
```
anforowicz, line 52: Q1: I assume that world ids are generated by the browser (when an extension asks to create a new world or world id)?  (as opposed to having an extension generate unique world id strings, and then pass those IDs to the browser/extension APIs)

Q2: Can a malicious extension try to spoof/forge world ids that belong to another extension?

proposals/multiple_user_script_worlds.md
tophf: Side-channel attacks notwithstanding, it's much more than hygiene, because isolation means JavaScript artifacts are isolated such as variables and DOM expandos and there's no way to access them directly, same as extension's content scripts are isolated from the MAIN world scripts. The "cross-world contact" is only possible via events and modifying DOM elements, which doesn't violate JavaScript isolation concept.

tophf: What about `JavaScript environment is isolated from one another (note that a side-channel attack is still possible as it's the same OS process)`?

proposals/multiple_user_script_worlds.md
rdcronin: I've added a link to issue 279, where many of these were discussed.

proposals/multiple_user_script_worlds.md
rdcronin: Sounds like we're agreed we can discuss these events in a follow-up proposal, so I'll resolve this conversation for now.

proposals/multiple_user_script_worlds.md
rdcronin: Done

proposals/multiple_user_script_worlds.md
rdcronin: Q1: not quite -- these IDs _are_ provided by extensions.  On the chrome side, we'll segment these by extension ID, so that the unique world configuration ID would be pair<extension id, world id> (though note that there may be multiple worlds with that same pair across different renderers -- this is intentional, those should inject in the same world configuration).

Q2: Nope, because of this pairing.  (Extension A's "my world" has no overlap with Extension B's "my world").

proposals/multiple_user_script_worlds.md
rdcronin: I think @anforowicz is looking at this from the browser security model, where we don't consider this to be a strict security boundary.  I already had language that hinted at this in the "Additional Security Considerations" (with its "mildly" and "somewhat" language : )), but I've made this more clear by adding an additional note that says this is not a hard security boundary and we wouldn't consider it in the browser's security model.

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,293 @@
...
+      * The JavaScript execution environment to run the script in. The default
+      * is `USER_SCRIPT`
+      */
+     world?: ExecutionWorld;
+
++    /**
++     * If specified, specifies a specific user script world ID to execute in.
++     * Only valid if `world` is omitted or is `USER_SCRIPT`. If omitted, the
++     * script will execute in the default user script world.
++     * Values with leading underscores (`_`) are reserved.
```
anforowicz, line 145: optional nit: Maybe require **two** leading underscores?  It seems that double-leading-underscore is less likely to be used in practice than a single leading underscore.  And this would also be [consistent with C++](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#:~:text=Each%20name%20that%20contains%20a,name%20in%20the%20global%20namespace.).

QUESTION: Just to double-check my understanding - reserved world ids may lead to some undesirable behavior **within** an extension, but this is okay, because they won't have any effect **outside** of an extension.  In other words, a malicious extension won't be able to abuse reserved world ids to somehow cross a security boundary somewhere, right?

rdcronin: Thanks, Lukasz!

> optional nit: Maybe require two leading underscores?

Unfortunately, we have a decent amount of precedent in extension APIs for using single-underscore for reserved (a few different APIs, file paths in the extension directory, etc).  We might revisit that at some point in the future, but I think being consistent is best here for now.

> In other words, a malicious extension won't be able to abuse reserved world ids to somehow cross a security boundary somewhere, right?

They certainly shouldn't be able to! ; )  I'd expect any implementation to simply reject any call from an extension that specifies a world ID beginning with '_'.  The reservation of '_' IDs was just a suggestion in case implementors wanted to allow all user script worlds to have a world ID (e.g., using "_default" for the default world in the underlying implementation).  On the Chromium side, we just use a std::optional<>, so there's really no way an extension should be able to abuse this.

rdcronin:
I think that's everyone!

@Rob--W , is this good to merge in?

proposals/multiple_user_script_worlds.md
```diff
@@ -0,0 +1,293 @@
...
+security model (just as we don't consider isolated worlds to be a strict
+security boundary today).
+
+## Alternatives
+
+### Existing Workarounds
+
+Today, user script managers go to various lengths to try to isolate behavior
+between user scripts (for more information, see conversation in
+[issue 279](https://github.com/w3c/webextensions/issues/279). This includes
```
Rob--W:
```suggestion
[issue 279](https://github.com/w3c/webextensions/issues/279)). This includes
```

Added missing closing parenthesis.

Rob--W (approved):
One nit, I'll suggest a change and then use Github's UI to make the change before merging.

proposals/multiple_user_script_worlds.md
Rob--W: Actually, I cannot use Github's UI to commit the change because I don't have write access to the repository from which this PR is pulling. @rdcronin could you apply this change and then merge the PR?

xeenon (approved):

proposals/multiple_user_script_worlds.md
rdcronin: Done!

PR #561, opened 2024-03-08
Update window.browser spec status

patrickkettner:
The current version of the spec is listed as an unofficial proposal. I believe this is no longer accurate, since the group approved it.


oliverdunk (approved):
LGTM! Thanks.

PR #562, opened 2024-03-09
Browser-specific needs-triage labels

dotproto:
Replace the general "needs-triage" label with browser-specific versions of the needs-triage label.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -8,8 +8,3 @@
 "needs-triage: firefox": 
   - ""
 
-"needs-triage%3A safari": 
-  - ""
-
-"needs-triage%3A firefox": 
-  - ""
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
-# Add a 'needs triage' label for each browser to all new issues
+# Add a 'needs triage' label for each browser engine to all new issues
 "needs-triage: chrome":
   - ""
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,9 @@
-# Add needs triage label to all new issues
-needs-triage:
+# Add a needs triage for each browser to all new issues
+"needs-triage%3A chrome":
+  - ""
+
+"needs-triage%3A safari": 
+  - ""
+
+"needs-triage%3A firefox": 
   - ""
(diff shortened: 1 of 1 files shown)


xeenon (approved):

.github/labeler.yml
```diff
@@ -1,3 +1,9 @@
-# Add needs triage label to all new issues
-needs-triage:
+# Add a needs triage for each browser to all new issues
+"needs-triage%3A chrome":
```
oliverdunk: Since we're in quotes here, I don't think we need to encode the `:`? I did a quick test in https://github.com/oliverdunk/triage-test and it seems like that's the case.

oliverdunk (changes requested):
Thanks for this! Left one nit, otherwise looks good.

.github/labeler.yml
dotproto: Changes suggested by Oliver.

```suggestion
"needs-triage: chrome":
  - ""

"needs-triage: safari": 
  - ""

"needs-triage: firefox": 
```

.github/labeler.yml
```diff
@@ -1,3 +1,15 @@
...
+  - ""
+
+"needs-triage: firefox": 
+  - ""
+
+"needs-triage%3A safari": 
+  - ""
+
+"needs-triage%3A firefox": 
   - ""
```
dotproto: Remove extra lines.

```suggestion
```

.github/labeler.yml
```diff
@@ -1,3 +1,10 @@
-# Add needs triage label to all new issues
-needs-triage:
+# Add a needs triage for each browser to all new issues
```
dotproto:
```suggestion
# Add a 'needs triage' label for each browser to all new issues
```

oliverdunk (approved):
LGTM, thanks :)

.github/labeler.yml
```diff
@@ -1,3 +1,10 @@
-# Add needs triage label to all new issues
-needs-triage:
+# Add a 'needs triage' label for each browser to all new issues
```
Rob--W:
```suggestion
# Add a 'needs triage' label for each browser engine to all new issues
```

Let's emphasize that it is a distinct engine. @mukul-p just confirming - this is what we decided, right? In practice Edge hasn't made significant use of the consensus labels in the group, since Edge mostly follows Chromium in its extension implementation. Thus, it would reduce overhead to not require Edge to triage issues too.

mukul-p: sgtm.

mukul-p (approved):

Rob--W (approved):
Approved, with the requested "engine" addition included.

PR #564, opened 2024-03-12
Update Simeon's affiliation

dotproto:
I just updated my affiliation in the W3C to reflect that I'm working more closely with Mozilla. This PR updates the charter to reflect that change.

### DIFF
--- a/charter.md
+++ b/charter.md
@@ -228,7 +228,7 @@ standardization in an appropriate venue.
 
 The Chairs of the WebExtensions Community Group are:
 
-  * Simeon Vincent, Independent (and before at Google)
+  * Simeon Vincent, Mozilla (previously Google)
   * Timothy Hatcher, Apple
 
 The Chairs are responsible for the day-to-day running of the group,


tantek (approved):

xeenon (approved):

charter.md
```diff
@@ -228,7 +228,7 @@ standardization in an appropriate venue.
 
 The Chairs of the WebExtensions Community Group are:
 
-  * Simeon Vincent, Independent (and before at Google)
+  * Simeon Vincent, Mozilla (previously Google)
```
Rob--W, line 231: Should we include more specific dates in affiliation, possibly including the Independent affiliation? That would make it easier to rate the affiilation of older comments.

Even if we don't include the date, one could also bisect the meeting notes to find the date of "official" affiliation changes.

patrickkettner: I have never heard of that practice in other groups, have you?

Rob--W: I don't have enough experience in other groups to give an authorative answer. I asked the question, because "previously" is vague and becomes increasingly less relevant in the future. I have no issues with keeping the text as is though.

I looked up a name of someone involved in standards whose affiliation changed (Opera -> Mozilla -> Apple), and in that example only the latest/current company is mentioned.

hanguokai: I think adding a date would be helpful. Different browser vendors have different opinions, such as: whether to support service worker or event page and webRequest in MV3. It can be confusing if a person's perspective completely reverses due to a change in company affiliation. Adding a date can help others understand it.

dotproto: Keeping historical information for each participant inline feels like it could get unwieldy. The Privacy Interest Group has some relevant prior art; the [About section](https://www.w3.org/2019/09/privacy-ig-charter.html#about) of their charter has a table with the date of and short description of each change.  Perhaps we could do something like this?

dotproto: I've created #584 to track adding change history to the charter.

Issue #252, opened 2022-08-04
Proposal: `i18n.getPreferredSystemLanguages()` and `i18n.getSystemUILanguage()`


carlosjeurissen:
### Motivation
As browsers often don't directly use the OS language but some reflection of it based on what languages are supported by the browser UI. Having the original OS language is useful for language-related extensions and for extensions who want to provide translations more true to the OS language (like be able to adapt to a specific sub-language, like Belgium-Dutch (nl-BE). While `i18n.getUILanguage` would return 'nl'.

### Syntax
`i18n.getSystemUILanguage()` would synchronously return an `IETF tag` just like `i18n.getUILanguage()` does right now.

It would follow the following signature for [i18n.json](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/extensions/common/api/i18n.json):
```json
{
  "name": "getSystemUILanguage",
  "type": "function",
  "nocompile": true,
  "description": "Gets the UI language of the Operating System. This is different from $(ref:i18n.getUILanguage) which returns the UI language of the web browser.",
  "parameters": [],
  "returns": {
    "type": "string",
    "description": "An IETF BCP 47 language tag such as en-US or pt-BR."
  }
},
```

See proposal: https://github.com/w3c/webextensions/pull/569

xeenon:
This sounds fine to me for Safari.

hanguokai:
Why not use `navigator.language` and `navigator.languages`?

xeenon:
> Why not use `navigator.language` and `navigator.languages`?

It isn't available in service workers, IIRC.

birtles:
> > Why not use `navigator.language` and `navigator.languages`?
> 
> It isn't available in service workers, IIRC.

It seems like it should be available given [`ServiceWorkerGlobalScope`](https://w3c.github.io/ServiceWorker/#serviceworkerglobalscope) inherits from [`WorkerGlobalScope`](https://html.spec.whatwg.org/multipage/workers.html#workerglobalscope) whose `navigator` member has type [`WorkerNavigator`](https://html.spec.whatwg.org/multipage/workers.html#workernavigator) which includes [`NavigatorLanguage`](https://html.spec.whatwg.org/multipage/system-state.html#navigatorlanguage).

That said, I believe `navigator.language` / `navigator.languages` is a separate concern to this.

As I understand it we have:

1. The language of the OS (`browser.i18n.getOSLanguage`, i.e. this proposal)
[...]

Issue #274, opened 2022-09-15
Proposal: `i18n.withLanguage()`


carlosjeurissen:
As requested by @Rob--W, splitting this proposal off from https://github.com/w3c/webextensions/issues/258

### Why
- To be able to fetch messages from a different language than the browser UI language. This  can for example be useful if you want to display your extension in a different language.
- Fetching the language files manually is not really a doable alternative as it does not handle fallbacks to less specific language tags and English while adding a lot of boilerplate.

### Proposal A
Introduce a `i18n.getLanguageDictionary` method. In which the first and only parameter is the language tag as string. Returning a promise with the languageDictionary allowing you to request messages from as follows:

```js
browser.i18n.getLanguageDictionary('pt-BR').then((api) => {
  let someMessage = api.getMessage('some_id');
});
```

Or with async/await:
```js
const brazilApi = await browser.i18n.getLanguageDictionary('pt-BR');
let someMessage = brazilApi.getMessage('some_id');
```

The name of this method can be worked on. Potential alternatives are `createLanguageInstance` and `getLanguageInstance`.

A variant on this is directly passing the `getMessage` method as suggested by @hanguokai in https://github.com/w3c/webextensions/issues/274#issuecomment-2021652070 .

```js
browser.i18n.getLanguageDictionary('pt-BR').then((getMessage) => {
  let someMessage = getMessage('some_id');
});
```

Or with async/await:
```js
const getPtBrMessage = await browser.i18n.getLanguageDictionary('pt-BR');
let someMessage = getPtBrMessage('some_id');
```

### Proposal B
Introduce a `i18n.getLanguageDictionary` method. In which the first and only parameter is the language tag as string. Returning a promise with the languageDictionary allowing you to request messages from as follows:

```js
browser.i18n.getLanguageDictionary('pt-BR').then((languageDictionary) => {
  let someMessage = browser.i18n.getMessage('some_id', substitutions?, {
    dictionary: languageDictionary,
  });
});
```

### Proposal C
Introduce an async`i18n.initLanguage` method. In which the first and only parameter is the language tag as string. This would initialise the messagedictionary on the background. The messages can then be fetched sync by passing the language parameter as option.

```js
browser.i18n.initLanguage('pt-BR').then(() => {
  let someMessage = browser.i18n.getMessage('some_id', substitutions?, {
    language:'pt-BR',
  });
});
```

### Proposal D
Allow passing the requested language synchronously. Most convenient for developers however potentially tricky implementation-wise.

[...]

PR #569, opened 2024-03-18
Proposal: i18n-system-languages

carlosjeurissen:
This PR adds the proposal for both `i18n.getPreferredSystemLanguages()` and `i18n.getSystemUILanguage()` previously called `i18n.getOSLanguage()`, which all vendors are supportive on as discussed in https://github.com/w3c/webextensions/issues/252


xeenon (approved):

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Created:** 2024-03-18
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/252
+
+## Motivation
+
+### Objective
+
+As browsers often do not directly use the OS language but some reflection of it based on what languages are supported by the browser UI.
```
rdcronin: This reads a little incompletely to me.

Maybe:

```suggestion
Allow extensions to display UI in the user's preferred language, even when this is distinct from the browser's UI language (which is restricted to languages supported by the browser).
```

Feel free to wordsmith.

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Known Consumers
+
+Language-related extensions and extensions wanting to match the language of the OS more closely independent of the langauge of the browser.
+
+## Specification
+
+### Schema
+
+`i18n.getOSLanguage()` would synchronously return an IETF tag just like `i18n.getUILanguage()` does right now.
```
rdcronin: For completeness, can we link to the IETF spec for IETF tag?

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+  "returns": {
+    "type": "string",
+    "description": "An IETF BCP 47 language tag such as en-US or pt-BR."
+  }
+}
+```
+
+### New Permissions
+
+As browser.i18n.getUILanguage does not require a permission. browser.i18n.getOSLanguage also should not.
```
rdcronin:
```suggestion
As `browser.i18n.getUILanguage()` does not require a permission, browser.i18n.getOSLanguage also should not.
```

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Exposed Sensitive Data
+
+The language of the OS will be purposely exposed.
+
+## Alternatives
+
+### Existing Workarounds
+
+There is no current workaround.
```
rdcronin:
```suggestion
Currently, if an extension wants to offer languages outside of what the browser supports, it must do so by presenting a user-facing options page.  There is no way to reasonably detect a language code outside of the browser's UI language with existing APIs.
```

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Open Web API
+
+To combat fingerprinting, this should not be an Open Web API. Just like browser.i18n.getUILanguage is not.
```
rdcronin:
```suggestion
To combat fingerprinting, this should not be an Open Web API, just like browser.i18n.getUILanguage is not.
```

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+  "type": "function",
+  "nocompile": true,
+  "description": "Gets the UI language of the Operating System. This is different from $(ref:i18n.getUILanguage) which returns the UI language of the web browser.",
+  "parameters": [],
+  "returns": {
+    "type": "string",
+    "description": "An IETF BCP 47 language tag such as en-US or pt-BR."
+  }
+}
+```
```
rdcronin: nit: I'd lean towards describing these in typescript (for now), since it's a more portable format than Chromium's use of JSON schema.  Especially things like "nocompile" are very Chromium-specific.

That said, if other browser vendors are fine with (or prefer) this, I don't feel strongly, since, well, this is obviously good enough for Chromium : )

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+```json
+{
+  "name": "getOSLanguage",
+  "type": "function",
+  "nocompile": true,
+  "description": "Gets the UI language of the Operating System. This is different from $(ref:i18n.getUILanguage) which returns the UI language of the web browser.",
+  "parameters": [],
+  "returns": {
```
rdcronin: This implies a synchronous return.  Is that a requirement?

(I'm not sure if we'd have the OS language available in the renderer today, unlike the UI language)

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
+# Proposal: i18n.getOSLanguage>
+
+**Summary**
+
+Allow to get the language of your operating system as an `IETF tag`
+
+**Document Metadata**
+
+**Author:** carlosjeurissen
+
```
rdcronin: This dropped the "sponsoring browser" item [here](https://github.com/w3c/webextensions/blob/e02ace0c34a17ae99a446c6e73fb4db778f41a53/proposals/proposal_template.md?plain=1#L19).

That was added to ensure the proposals here don't become a graveyard of aspirational APIs and instead reflect ones that at least one browser is committed to implementing in the immediate future.

I'll need to do some quick investigation into the implementation.  If it looks straightforward, I'm comfortable putting Google Chrome as the sponsoring browser.

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,74 @@
+# Proposal: i18n.getOSLanguage>
+
+**Summary**
+
+Allow to get the language of your operating system as an `IETF tag`
+
+**Document Metadata**
+
+**Author:** carlosjeurissen
```
rdcronin: optional: maybe link to your github?  (Though I imagine you have carlosjeurissen just about everywhere : ))

rdcronin (changes requested):
Thanks, Carlos!

proposals/i18n-getOSLanguage.md
carlosjeurissen: Added.

proposals/i18n-getOSLanguage.md
carlosjeurissen: In Mozilla Firefox, the OS language is available in the renderer synchronously. In general the motivation to return synchronously is to match `i18n.getUILanguage`. Having it synchronously is also very welcome to load locale data as quickly as possible for bidi, lang attributes, locale specific files and potentially in the future, passing it to an API like [i18n.getLanguageDictionary](https://github.com/w3c/webextensions/issues/274)

proposals/i18n-getOSLanguage.md
carlosjeurissen: Rephrased to:
`Currently, if an extension wants to offer languages outside of what the current browser language uses, it must do so by presenting a user-facing options page. However, currently there is no way to reasonably detect a language code outside of the browser's UI language with existing APIs.`

proposals/i18n-getOSLanguage.md
carlosjeurissen: Added.

proposals/i18n-getOSLanguage.md
carlosjeurissen: Rephrased to:
`Allow extensions to display UI in the user's system language, even when this is distinct from the browser's UI language (which is restricted to languages supported by the browser).`

carlosjeurissen:
@xeenon Thanks for reviewing and the willingness to implement this.

@rdcronin Thanks for reviewing and your comments. Most feedback has been implemented in the proposal. Thanks for offering to sponser the feature as browser.

@Rob--W Reading from your comment in the reading notes of [Today's (2024-03-28) meeting](https://github.com/w3c/webextensions/pull/578):

> This proposal doesn’t follow the new process. One of the aspects of the new process template is that there should be at least one browser committing to implementing the API.

It was under my impression there was general support for this feature across browsers and the next step would be a proposal. However I understand your concerns regarding ending up with many proposals without browsers moving forward to actually implement them. So I went ahead and created a patch for Firefox.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1888486

proposals/i18n-getOSLanguage.md
carlosjeurissen: Added back the Sponsoring browser with a TBD.

xeenon: I'm supportive of sponsoring for Safari and plan to implement it as soon as we have consensus.

proposals/i18n-getOSLanguage.md
xeenon: Agreed, TypeScript would make more sense.

Rob--W:
> @Rob--W Reading from your comment in the reading notes of [Today's (2024-03-28) meeting](https://github.com/w3c/webextensions/pull/578):
> 
> > This proposal doesn’t follow the new process. One of the aspects of the new process template is that there should be at least one browser committing to implementing the API.
> 
> It was under my impression there was general support for this feature across browsers and the next step would be a proposal.

Part of the proposal template includes an item about the "sponsoring" browser. As Devlin commented elsewhere in the PR, the purpose of that is to encourage proposals that reflect APIs that have a path towards realization rather than just being aspirational.

> However I understand your concerns regarding ending up with many proposals without browsers moving forward to actually implement them. So I went ahead and created a patch for Firefox.
> See: https://bugzilla.mozilla.org/show_bug.cgi?id=1888486

Thanks for the bug and patch!

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Known Consumers
+
+Language-related extensions and extensions wanting to match the language of the OS more closely independent of the langauge of the browser.
+
+## Specification
+
+### Schema
+
+`i18n.getOSLanguage()` would synchronously return an [IETF tag](https://www.w3.org/International/core/langtags/rfc3066bis.html) just like `i18n.getUILanguage()` does right now.
```
aphillips: The proper reference here is IETF BCP47. The current RFCs for that are 5646 and 4647. Generally speaking its common to talk about a "BCP47 language tag" or just "language tag". See also the [I18N Glossary](https://www.w3.org/TR/i18n-glossary) for useful links to I18N terminology for the Web.

carlosjeurissen: @aphillips Thanks for the references! Updated the proposal to reflect this!

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,78 @@
+# Proposal: i18n.getOSLanguage>
```
Rob--W: What is the `>` doing at the end?

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,78 @@
...
+
+Language-related extensions and extensions wanting to match the language of the OS more closely independent of the langauge of the browser.
+
+## Specification
+
+### Schema
+
+`i18n.getOSLanguage()` would synchronously return a [BCP47 language tag](https://www.w3.org/International/core/langtags/rfc3066bis.html) like `i18n.getUILanguage()` does right now.
+
+It would follow the following signature for [i18n.json](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/extensions/common/api/i18n.json):
```
Rob--W: When referencing source code, try to use a permalink where possible. In this case there has been a request to use IDL instead of JSON, so this line will probably end up being dropped. If not, here is the permalink: https://chromium.googlesource.com/chromium/src/+/4299ce68496b32ba309e2f012e0db5b4b8cd478a/extensions/common/api/i18n.json

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,78 @@
+# Proposal: i18n.getOSLanguage>
+
+**Summary**
+
+Allow to get the language of your operating system as an [BCP47 language tag](https://www.w3.org/International/core/langtags/rfc3066bis.html).
```
Rob--W: Given the discussion starting from https://github.com/w3c/webextensions/issues/252#issuecomment-2027025304, should we rename the proposal to `i18n.getOSLanguages` and let it return an array?

And also, guarantee the array to be non-empty?

hanguokai:
As the discussion deepens (#252), I think there are four concepts here:
1. The Browser UI Language
2. The Browser Accept Languages
3. The OS UI Language
4. The OS Accept Languages

So, ideally there should be four functions:
1. `i18n.getBrowserUILanguage()`
2. `i18n.getBrowserAcceptLanguages()`
3. `i18n.getOSUILanguage()`
4. `i18n.getOSAcceptLanguages()`

And deprecate the current `i18n.getUILanguage()` and `i18n.getAcceptLanguages()`.

Or using another design:
```
// return a single string for UI language
i18n.getUILanguage(
    prefer: "browser" // or "os"
);

// return an array for accept languages
i18n.getAcceptLanguages(
    prefer: "browser" // or "os"
);
```

proposals/i18n-getOSLanguage.md
carlosjeurissen: Eh, it is not supposed to be there. Thanks for noting! Fixed.

proposals/i18n-getOSLanguage.md
carlosjeurissen: Good to know! Updated the proposal to reflect this!

proposals/i18n-getOSLanguage.md
```diff
@@ -0,0 +1,78 @@
+# Proposal: i18n.getOSLanguage
+
+**Summary**
+
+Allow to get the language of your operating system as an [BCP47 language tag](https://www.w3.org/International/core/langtags/rfc3066bis.html).
```
aphillips: This is a really ancient link (thank you for discovering it so that we/i18n can fix it). Consider linking to the real deal:

```suggestion
Allows getting the language/locale of your operating system as a [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html)
```

I18N discussed this issue in our call of 2024-04-18. I will add comments to #252.

carlosjeurissen: Thanks for the update @aphillips ! Great to hear it was discussed. looking forward to the comments.

carlosjeurissen:
Updated the proposal which addresses the following:
`i18n.getOSLanguage()` has been renamed to `i18n.getSystemUILanguage()` to better indicate it is supposed to be the user interface of the operating system. Not the first preferred language specified in the operating system.

`i18n.getPreferredSystemLanguages()` has been added which returns the above mentioned preferred languages specified in the operating system. This aligns with the same API and name available in Electron apps.

For `i18n.getPreferredSystemLanguages()`, a patch for Firefox has been made, see: https://bugzilla.mozilla.org/show_bug.cgi?id=1888486

proposals/i18n-getOSLanguage.md
carlosjeurissen: @aphillips Updated the PR to address some of the feedback raised in #252. Namely splitting the method into two separate ones with better naming. See https://github.com/w3c/webextensions/pull/569#issuecomment-2126669092

Seems all vendors have no objection to the changes and this PR. Is there anything you would like to add here?

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,97 @@
...
+
+Allow extensions to display UI in the user's system language(s), even when this is distinct from the browser's UI language (which is restricted to languages supported by the browser).
+
+#### Use Cases
+
+Having the original system language(s) is useful for language-related extensions and for extensions who want to provide translations more true to the system language. For example, to adapt to a specific sub-language, like Belgium-Dutch (nl-BE). While i18n.getUILanguage would return 'nl'.
+
+### Known Consumers
+
+Language-related extensions and extensions wanting to match the language of the operating system more closely independent of the langauge of the browser.
```
xPaw:
```suggestion
Language-related extensions and extensions wanting to match the language of the operating system more closely independent of the language of the browser.
```

carlosjeurissen: Thanks!

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+### Objective
+
+Allow extensions to display UI in the user's system language(s), even when this is distinct from the browser's UI language (which is restricted to languages supported by the browser).
+
+#### Use Cases
+
+Having the original system language(s) is useful for language-related extensions and for extensions who want to provide translations more true to the system language. For example, to adapt to a specific sub-language, like Belgium-Dutch (nl-BE). While i18n.getUILanguage would return 'nl'.
```
aphillips: The example here is not very clear. There is a tendency to think that the primary language subtag (`nl` in the examples here) is "the language", but there are language variations that depend on additional subtags, such as script or region. For example, `zh-Hans` (Simplified Chinese) vs. `zh-Hant` (Traditional Chinese). Thus, the UI language is not guaranteed to be a single subtag.

It's also more useful/helpful to talk about the [locale](https://www.w3.org/TR/ltli#locale) than it is to talk about the language, even if, for historical reasons, we use the word "Language" in the API names.

Note that locale means _more_ than just language adaptation. Frequently, locale-based variations in number, date, time, measurement, etc. formatting constitute a key part of adaptation. 

Note too that the use cases here are incomplete. The PR lists a number of different reasons, but you only have one in the document itself.

Thus:

```suggestion
Extension authors might want to obtain the locale used by the user agent's host environment in order to format values (such as numbers, dates, and so forth) or do other locale-affected operations (such as list sorting) according to expectations of the user. For example, formatting the date `2024-06-15` varies between `en-US` (6/15/2024) and `en-GB` (15/6/2024), whereas `i18n.getUILanguage` might return only `en` (English).

Extension authors might want to obtain the locale used by the user agent's host environment to better match the runtime environment's localization. For example, the operating system might be running in `ff-NG` (Fulani as used in Nigeria), but the user agent might not be localized into this language and return `en` (English) for `i18n.getUILanguage`.
```

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Known Consumers
+
+Language-related extensions and extensions wanting to match the language of the operating system more closely independent of the language of the browser.
+
+## Specification
+
+### Schema
+
+`i18n.getPreferredSystemLanguages()` would asynchronously return a list of [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html) like `i18n.getAcceptLanguages()` does right now.
```
aphillips: Probably you want to say that it returns an **_ordered_** list (BCP47 calls this a "language priority list").

Note that most operating systems don't provide a feature like this.

carlosjeurissen: @aphillips thanks for noting and clarifying! Updated the sentence to reflect these suggestions. It is indeed meant to be an ordered list. As far as I know both mac and windows support this. But I could be wrong?

proposals/i18n-system-languages.md
carlosjeurissen: @aphillips Thanks so much for the suggestions! Applied the suggested change.

Also added a better example of the purest form use-case which describes `i18n.getUILanguage` returning a language tag with additional subtags which might not be sufficient for the user.

Lastly rewritten sentences to use locale instead of language where it makes sense based on your suggestion.

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,101 @@
...
+
+**Summary**
+
+Allows getting the language(s)/locale(s) of your operating system as a [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html)
+
+**Document Metadata**
+
+**Author:** [carlosjeurissen](https://github.com/carlosjeurissen)
+
+**Sponsoring Browser:** TBD
```
dotproto: In today's meeting @xeenon said he was interested in sponsoring.

```suggestion
**Sponsoring Browser:** Safari
```

xeenon:
@carlosjeurissen There seems to have been a rebase issue, where this PR picked up a bunch of other changes from the repo outside of your proposal.

carlosjeurissen:
@xeenon Thanks for noting! Has been addressed.

carlosjeurissen:
Based on some of the feedback during TPAC, updated this PR and updated the Firefox patch.
See:
https://phabricator.services.mozilla.com/D206064
https://bugzilla.mozilla.org/show_bug.cgi?id=1888486

proposals/i18n-system-languages.md
```diff
@@ -67,11 +70,14 @@ It would follow the following signature for [i18n.json](https://chromium.googles
   "name": "getSystemUILanguage",
   "type": "function",
   "nocompile": true,
-  "description": "Gets the current UI language of the Operating System. This is different from $(ref:i18n.getUILanguage) which returns the UI language of the web browser.",
+  "description": "Gets the current UI locale of the Operating System. This is different from $(ref:i18n.getUILanguage) which returns the UI locale of the web browser. This locale could include additional information like preferred datetime format and number system if supported by the OS.",
```
xeenon: Is the "additional information" part a standard? I know Apple platforms can include it, but that isn't something we have had in web extensions before.

carlosjeurissen: @xeenon yes, this is specified by unicode. This was discussed during the shared session with the WECG/I18NWG meeting at TPAC. However, giving a further look at this I see how this should not be part of the getSystemUILanguage. Updated the PR to reflect this. Still, providing this information can be useful, but it should be implemented using some other method / proposal.

carlosjeurissen:
@Rob--W were you able to take a look at this?

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,107 @@
...
+
+The returned language tag can be different from the first entry returned by getPreferredSystemLanguages as the operating system could not support all languages specified by the user for its user interface.
+
+It would follow the following signature for [i18n.json](https://chromium.googlesource.com/chromium/src/+/4299ce68496b32ba309e2f012e0db5b4b8cd478a/extensions/common/api/i18n.json):
+
+```json
+{
+  "name": "getSystemUILanguage",
+  "type": "function",
+  "nocompile": true,
```
Rob--W:
```suggestion
```

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,107 @@
...
+      "type": "array",
+      "items": {"$ref": "LanguageCode"},
+      "description": "Array of LanguageCode"
+    }]
+  }
+}
+
+```
+
+`i18n.getSystemUILanguage()` would synchronously return a [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html) like `i18n.getUILanguage()` does right now.
```
Rob--W: This states that `getSystemUILanguage` is synchronous, but the API description below states that the API is asynchronous. Which of the two is it?

There is relevant prior discussion at https://github.com/w3c/webextensions/pull/569#discussion_r1543266374. If there is a compelling use case for it being synchronously available, and other browser vendors are onboard, I'd be supportive of the API being synchronous.

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,107 @@
...
+      "type": "string",
+      "description": "A BCP47 language tag such as en-US or pt-BR."
+    }]
+  }
+}
+```
+
+### New Permissions
+
+As `browser.i18n.getUILanguage()` does not require a permission, browser.i18n.getSystemUILanguage and browser.i18n.getPreferredSystemLanguages also should not.
```
Rob--W:
```suggestion
As `browser.i18n.getUILanguage()` does not require a permission, `browser.i18n.getSystemUILanguage` and `browser.i18n.getPreferredSystemLanguages` also should not.
```

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Alternatives
+
+### Existing Workarounds
+
+Currently, if an extension wants to offer locales outside not localised for by the browser, it must do so by presenting a user-facing options page. However, currently there is no way to reasonably detect a language code outside of the browser's UI language with existing APIs.
+
+### Open Web API
+
+To combat fingerprinting, this should not be an Open Web API, like browser.i18n.getUILanguage is not.
```
Rob--W:
```suggestion
To combat fingerprinting, this should not be an Open Web API, like `browser.i18n.getUILanguage` is not.
```

proposals/i18n-system-languages.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Schema
+
+`i18n.getPreferredSystemLanguages()` would asynchronously return a language priority list (ordered list of [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html). This is similar to `i18n.getAcceptLanguages()`, with the difference that the priority list is configured by the user in the OS and the method would not add less specific language tags in place.
+
+```json
+{
+  "name": "getPreferredSystemLanguages",
+  "type": "function",
+  "description": "Gets the preferred languages of the operating system. This is different from the languages set in the browser; to get those, use $(ref:i18n.getAcceptLanguages).",
```
Rob--W: The introduction of this md document emphasizes the distinction between language and locale, but here the description of `getPreferredSystemLanguages` refers to language only. The description below mentions locale, based on the feedback from https://github.com/w3c/webextensions/pull/569#discussion_r1784857029

Unless you really intended to limit the meaning to "language (not locale)", let's try to be consistent.

Rob--W (commented):
The general idea and proposal looks good to me, but I'm waiting with signing off until the next review, to confirm that inconsistencies have been resolved (in particular on async vs sync and locale vs language).

proposals/i18n-system-languages.md
Rob--W: Safari has implemented this; it is async.

dotproto:
@rdcronin @oliverdunk to follow up on this on the Chrome side.

proposals/i18n-system-languages.md
carlosjeurissen: Updated the PR. Now states async.

proposals/i18n-system-languages.md
carlosjeurissen: Updated the PR. Now states locale.

Rob--W (approved):

Rob--W:
I have squashed the changes instead of creating a merge commit because otherwise there would be 31 noisy commits as part of the merge.

Thanks for your work here @carlosjeurissen !

PR #576, opened 2024-03-27
Add guidance on the review expectations of proposals

rdcronin:
As we discussed at the in-person WECG meet-up, we should add expectations around review practices. This PR adds documentation for those practices.

### DIFF
--- a/proposals/proposal_process.md
+++ b/proposals/proposal_process.md
@@ -184,3 +184,58 @@ sponsored by a browser vendor) and a reviewing browser vendor are unable to
 reach consensus (with regard to either the API's concept or an individual
 aspect of it).  In this scenario, the reviewing vendor must choose a stance
 (which will very likely be "Opposed") to allow the API to move forward.
+
+## Review Process
+
+Reviewing new proposals is one of the most important and impactful roles of the
+Web Extensions Community Group. This ensures browsers have the opportunity to
+align on shared functionality when desirable for new functionality and helps
+prevent drift between the API surfaces. As such, it is critical to ensure
+timely reviews.
+
+### Authors
+
+Authors should request a review from at least one representative from each
+browser engine.
+
+When responding to feedback from reviewers, authors should resolve conversations
+if they feel they have been fully addressed. This makes it easy to see if there
+are outstanding comments. Authors can (and should) leave comments open if they
+would like further input from reviewers (e.g., "Is this what you were looking
+for").
+
+If reviewers haven't responded within a few business days, authors should ping
+the relevant reviewers.
+
+### Reviewers
+
+Reviewers should strive to respond to a review within one or two business days;
+this is not a strict SLO, but is a target -- we understand that this may not
+always be possible (e.g. if multiple reviewers are at a summit).
+
+Reviewers should indicate if they are supportive of a proposal by approving it.
+
+Reviewers should resolve any comment threads they began if they feel the comment
+has been sufficiently addressed. This makes it easy to see if there are
+outstanding comments.
+
+Reviewers should *not* block on other reviewers in most cases. A reviewer's
+approval is only indicative of their own approval, not of the readiness of a
+proposal to merge. This prevents a deadlock where reviewer A waits on reviewer
+B and reviewer B waits on reviewer A.
+
+Reviewers should be clear whether they have blocking requests by indicating
+they "Request changes" on the review.
+
+Reviewers may approve a proposal "with nits" if they trust the author to update
+the proposal before (or shortly after) it's merged. As an example, a typo or
+slight rephrasing need not block a proposal from being merged.
+
+### Merge Managers
+
+A proposal should be merged only when no reviewers have requested additional
+changes. Just because a single reviewer has approved a proposal does *not*
+mean it should be merged.
+
+When all reviewers have approved (or abstained by not leaving any comments),
+a proposal should be merged.
--- a/proposals/proposal_process.md
+++ b/proposals/proposal_process.md
@@ -189,9 +189,8 @@ aspect of it).  In this scenario, the reviewing vendor must choose a stance
 
 Reviewing new proposals is one of the most important and impactful roles of the
 Web Extensions Community Group. This ensures browsers have the opportunity to
-align on shared functionality when desirable for new functionality and helps
-prevent drift between the API surfaces. As such, it is critical to ensure
-timely reviews.
+align on shared functionality when desirable and helps prevent drift between the
+API surfaces. As such, it is critical to ensure timely reviews.
 
 ### Authors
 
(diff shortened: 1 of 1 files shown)


oliverdunk (approved):
LGTM, thanks for writing this up. Of course, my approval does not indicate that we should merge this yet 😉

xeenon (approved):

proposals/proposal_process.md
```diff
@@ -184,3 +184,58 @@ sponsored by a browser vendor) and a reviewing browser vendor are unable to
...
 aspect of it).  In this scenario, the reviewing vendor must choose a stance
 (which will very likely be "Opposed") to allow the API to move forward.
+
+## Review Process
+
+Reviewing new proposals is one of the most important and impactful roles of the
+Web Extensions Community Group. This ensures browsers have the opportunity to
+align on shared functionality when desirable for new functionality and helps
+prevent drift between the API surfaces. As such, it is critical to ensure
+timely reviews.
```
dotproto: Current draft uses the word "functionality" twice in quick succession. We should either reword to remove one or use a synonym. 

```suggestion
Reviewing new proposals is one of the most important and impactful roles of the
Web Extensions Community Group. This ensures browsers have the opportunity to
align on shared behavior when desirable for new functionality and helps prevent
drift between the API surfaces. As such, it is critical to ensure timely
reviews.
```

proposals/proposal_process.md
```diff
@@ -184,3 +184,58 @@ sponsored by a browser vendor) and a reviewing browser vendor are unable to
...
+If reviewers haven't responded within a few business days, authors should ping
+the relevant reviewers.
+
+### Reviewers
+
+Reviewers should strive to respond to a review within one or two business days;
+this is not a strict SLO, but is a target -- we understand that this may not
+always be possible (e.g. if multiple reviewers are at a summit).
+
+Reviewers should indicate if they are supportive of a proposal by approving it.
```
dotproto: Conversely does this mean reviewers should mark opposition to a proposal by rejecting it?

There's an important distinction between "I think this PR is in a good state to merge" and "I believe this is something the group should work on." Typically GitHub PR approval signals the former, and the WECG has been using the "supportive: <vendor>" labels to signal that a browser vendor wishes to pursue a topic.

proposals/proposal_process.md
```diff
@@ -184,3 +184,58 @@ sponsored by a browser vendor) and a reviewing browser vendor are unable to
...
+for").
+
+If reviewers haven't responded within a few business days, authors should ping
+the relevant reviewers.
+
+### Reviewers
+
+Reviewers should strive to respond to a review within one or two business days;
+this is not a strict SLO, but is a target -- we understand that this may not
+always be possible (e.g. if multiple reviewers are at a summit).
```
dotproto: Reworded to read a little more clearly and to use more common punctuation.

```suggestion
Reviewers should strive to respond to a review within one or two business days.
This is a target, not a strict SLO: we understand that this may not always be
possible (e.g. if multiple reviewers are at a summit).
```

dotproto (changes requested):

Rob--W (approved):
Description is quite wordy, but matches my understanding of what we have discussed.

The distinction between "proposal is in an acceptable state (well formed, etc.)" and "browser engine supports the implementation" isn't clearly expressed, but I see that Simeon had already raised that and will let him handle that. We try to align in this group, but there will eventually be APIs that are adopted by part of the group and opposed by others. That shouldn't prevent a proposal from being merged.

proposals/proposal_process.md
rdcronin: Good call; I just dropped the second part entirely ("shared functionality when desirable for new functionality" -> "shared functionality when desirable")

proposals/proposal_process.md
rdcronin: why you no like emdashes : P

Mostly took yours, but I'm keeping a semicolon in there!  (I think it's more appropriate than a colon after the "not a strict SLO").

proposals/proposal_process.md
rdcronin: Thanks for catching this.  By "support" here, I really meant "you should indicate you think this is good to merge."

Reviewers indicate general support or opposition to a proposal via the labels, as described already in the "Browser Vendor Support" section.  I've clarified the intention here.

rdcronin:
I think all comments are addressed.  @dotproto and @Rob--W , please take another look.

hanguokai:
I think this process lacks reflection of developers' opinions and is more about collaboration between browser vendors. So, I recommend that at least one of the reviewers is not affiliated with any browser vendor and has developed at least one extension with at least(>=) 500 users.

rdcronin:
@hanguokai I definitely agree that developer feedback is a critical aspect in API proposals!  Since the current editors and chairs of the WECG are affiliated with browser vendors, I'm hesitant to add "find an extension developer" as a necessary step in this process.  Additionally, we're already at the point of having 3+ different reviewers, and review by consensus doesn't scale.

I think we absolutely should (and do!) welcome developer feedback on proposals and in the issues that likely led up to them, but I don't think we'll have it as a gating switch on proposal submission.  However, I do think that if there's clear developer feedback that isn't addressed, browser vendors will take that into account and request it be addressed before they sign off.  (For instance, if I were a reviewer on a proposal and a developer raised significant questions that weren't addressed, I'd ask the author to please address those before I approved the proposal.)

hanguokai:
I don't want to block the approval of this PR here because these issues can be gradually improved in the future. Here are my personal opinions:

1. Developers' position and perspective are not always consistent with browser vendors' position and perspective. For example, "pushing more burdens to developers" or "significant breaking changes". So no matter how many browser vendors participate, they cannot replace developers.
2. No matter how tech-savvy a person is, if they lack practical extension development experience, they may not be able to fully understand the issues that developers are concerned about and will not be able to come up with good proposals.
3. "Find a developer as a reviewer" is much easier than "Find a "sponsoring browser" that is committed to implementing the proposal soon" (if you are a developer).
4. Historically, some developer feedback was ignored by browser vendors. After thinking deeply, my conclusion is that developers "人微言轻"[1] and lack decision-making power. So, adding developers as reviewers could help improve this aspect.

[1] In one's humble position, one's word does not carry much weight.; The words of the lowly carry little weight.; When a man is in a low position, his words are of little effect.

rdcronin:
I think we're all aligned that incorporating developer feedback is critical for success, and the intent here is definitely not to circumvent or minimize that.  We will definitely solicit (and listen to) developer feedback on proposals, though, of course, we may not always be able to satisfy all parties.

I don't think we'll change this to have a developer have sign-off power on proposals, but I will create a separate PR that indicates more clearly that developers should be empowered to, and encouraged to, provide feedback on API proposals and that browser vendors should consider this feedback during the review.

hanguokai:
> I don't think we'll change this to have a developer have sign-off power on proposals, but I will create a separate PR that indicates more clearly that developers should be empowered to, and encouraged to, provide feedback on API proposals and that browser vendors should consider this feedback during the review.

I don't pursue unrealistic results, on the contrary, I pursue the best possible results. So, this result is acceptable to me (I represent myself here, not other developers).

In other words, it also implies that if a person joins or leaves a browser vendor, his or her reviewer role will change automatically, right?

rdcronin:
> In other words, it also implies that if a person joins or leaves a browser vendor, [their] reviewer role will change automatically, right?

Yep, I'd say so.  They can, of course, continue to provide feedback as a developer and someone familiar with the ecosystem, but they would no longer have a "blocking" review.  (And note also that even vendors don't have fully blocking reviews.  These guidelines are more about timeliness to ensure we don't get bogged down.)

dotproto (commented):
LGTM. Thanks, @rdcronin!

Issue #401, opened 2023-06-01
Expose SHA256 value of DownloadItem


patrickkettner:
@ericlaw1979 [opened an issue on crbug](https://bugs.chromium.org/p/chromium/issues/detail?id=1450291) about exposing SHA256 of [DownloadItem](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/DownloadItem). This is a fairly trivial change in chromium as the value is already calculated and stored, but were unclear if adding this functionality would be burdensome elsewhere and want to check that before discussing more internally if we want to do this.

Rob--W:
Regardless of currently existing implementations, I don't think that it would be a good design to include this information in DownloadItem by default. It being there implies that the information should be present at all times, even for the majority of the extensions that don't use it. This increases the required complexity for browsers that implement the downloads API.

It would be more acceptable to provide the functionality through a dedicated method, e.g. a new `downloads.getFileHash`:

- behind a new `downloads.getFileHash` permission, even if there is no permission message.
- only meaningful for completed downloads.
- rejects if a hash is not available.
- to be decided: desired behavior if the file does not exist.
- to be decided: desired behavior if the file has changed.
- to be decided: desired behavior if the file has relocated.

Beyond the API design aspect, there is the privacy/security consideration: this functionality is essentially an oracle that emits the sha256 hash for any web content. This hash can easily be reversed to the original value when the number of possible options is relatively small. A way to counter this is to add a permission warning for the extension permission.

Rob--W:
https://github.com/w3c/webextensions/labels/follow-up%3A%20chrome label here means: Patrick to follow up with the engineers to answer the (design) questions/considerations raised above and in the meeting.

ericlaw1979:
- I agree that putting this API behind a permission makes sense, and agree that _most_ extensions, including _many_ download-monitoring extensions will not need it. 

- From a privacy/security POV: Yes, this does reveal information about the response content, and because a malicious extension could cause _any_ URL to be treated as a "Download", it does seem like we should probably put this behind a permission prompt at install time.

- Making this a function rather than a property seems entirely reasonable for performance and ergonomics reasons.
- Returning an error if the file has been moved or deleted seems entirely reasonable.
[...]

Issue #563, opened 2024-03-09
New API enum values should use lowercase-dash-delimited style


xeenon:
At present, [`ExecutionWorld`](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ExecutionWorld) requires values to be uppercase, specifically `MAIN` or `ISOLATED`. Similarly, [`StyleOrigin`](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/scripting/insertCSS#origin) also requires the uppercase values `AUTHOR` or `USER`. This approach, for these newer APIs, is in contrast to the convention used for older properties, such as [`RunAt`](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/extensionTypes/RunAt), which uses lowercase values like `document_start` or `document_idle`.

To improve consistency across the Web Extensions API, I recommend standardizing in favor of lowercase for these values. Additionally, for backward compatibility, browsers should adopt a case-insensitive comparison for these values until the uppercase values are sufficiently phased out. 

Moving forward, I suggest the adoption of lowercase string constants in all future APIs, to establish a uniform standard.

erosman:
Original MV2 [tabs.insertCSS(): cssOrigin](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/insertCSS#cssorigin) & [CSSOrigin](https://developer.chrome.com/docs/extensions/reference/api/extensionTypes#type-CSSOrigin) also used lowercase "user" & "author".

rdcronin:
I agree with deciding on a uniform standard.

ExecutionWorld is not the first (or by any means, only) API to use SCREAMING_STYLE for the enum values.  We have historically been very inconsistent about this, using combinations of SCREAMING_STYLE, camelCaseStyle, and snake_case_style.  We eventually somewhat standardized in Chrome on SCREAMING_STYLE because these were similar to enums or constants, both of which are represented with SCREAMING_STYLE in the [Google JS Style Guide](https://google.github.io/styleguide/jsguide.html#naming-enum-names), and is also the [referenced "common convention"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage) in MDN to represent constants -- though of course, MDN would never require a certain style.

I would prefer aligning on using all-caps to better fit with these existing patterns and conventions.

[...]

PR #587, opened 2024-04-08
Proposal: downloads.getFileHash()

bershanskiy:
This formalizes #401 into a concrete proposal.

### DIFF
--- a/proposals/downloads_get_file_hash.md
+++ b/proposals/downloads_get_file_hash.md
@@ -1,0 +1,160 @@
+# Proposal: `downloads.getFileHash()`
+
+**Summary**
+
+Allow extensions to query hashes of contents of downloaded files
+programmatically.
+
+**Document Metadata**
+
+**Author:** @bershanskyi
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** @Rob--W
+
+**Created:** 2024-04-08
+
+**Related Issues:** [WECG GitHub issue 401](https://github.com/w3c/webextensions/issues/401)
+
+## Motivation
+
+### Objective
+
+This API will enable extensions to privately and efficiently scan user
+downloads for "Known Unsafe" contents and validate the integrity of "Known Safe"
+files without exposing the raw content of the files.
+
+#### Use Cases
+
+##### Filtering for unwanted downloads
+
+Many antivirus products detect previously seen "safe" and "unsafe" files by
+computing a file content hash and checking it against a set of known hashes.
+These checks require minimal effort on the client side, yet are powerful
+enough to limit the spread of known malware downloads since effectively every
+client gets "inoculated" against a particular download as soon as a single
+antivirus detects a malicious file.
+User-Agents may store file hashes even after the original file is deleted to
+facilitate tracing of malware exposure even if the file is deemed to be
+dangerous after the user initially downloads it.
+
+##### Download integrity assurance
+
+Files can get damaged while they are shared on a file hosting server or NAS.
+While some file formats explicitly support integrity checks like embedded
+checksums or even cryptographic signatures, other files do not. This API
+would allow file managers or extensions to programmatically check the
+integrity of downloaded files if the file source elects to provide them.
+
+### Known Consumers
+
+None at the moment.
+
+## Specification
+
+### Schema
+
+Schema in Chromium-style IDL:
+```idl
+namespace downloads {
+    callback GetFileHashCallback = void(DOMString hash);
+
+    [supportsPromises, permissions="downloads.getFileHash"]
+    static void getFileHash(
+        long downloadId,
+        [legalValues=("SHA-256")] DOMString algorithm,
+        GetFileHashCallback callback);
+}
+```
+
+Schema in TypeScript:
+```ts
+namespace downloads {
+  enum FileHashAlgorithm {
+    SHA256 = "SHA-256"
+  }
+
+  getFileHash: (
+    downloadId: number,
+    algorithm?: keyof downloads.FileHashAlgorithm)
+      => Promise<string>;
+}
+```
+
+### New Permissions
+
+No new permission. `downloads.getFileHash()` is honored only if extension has
+`"downloads"` permission and the host access for `DownloadItem.finalUrl`. For
+`data:`-URLs, host permissions are not required, because the content of the
+file is already fully visible through the URL.
+
+### Manifest File Changes
+
+No changes.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Hashes (cryptographic) are considered one-way operations, but they can easily
+be reversed to the original input when the number of possible inputs is
+sufficiently small. So extension may be able to infer the raw contents of the
+file if the possible space of file contents is small.
+A malicious extension could:
+ - Observe hashes of downloaded files and send them to a home server
+ - Initiate a download with an arbitrary URL and obtain a hash of the response
+ - Initiate a download for a ranged HTTP request and obtain a hash of the
+   corresponding portion of the response
+
+### Abuse Mitigations
+
+The API method requires host access to the `DownloadItem.finalUrl`.
+
+### Additional Security Considerations
+
+User Agents must not provide obsolete hash algorithms (like MD5 and SHA-1).
+User Agents must expose only hashes of complete files and never provide
+partial hash results while the file is being downloaded. This is necessary to
+prevent malicious or compromised extensions from calculating incremental hashes
+and constructing a "hash oracle" to infer file contents one byte (or just a few
+bytes) at a time.
+
+## Alternatives
+
+### Existing Workarounds
+
+As of writing, extensions can calculate the digest of a download by
+implementing a native host component that would read the desired file from the
+disc directly. This approach is a significant elevation of privileges (requires
+unsandboxed application), largely expands the complexity of implementing such
+an extension (since the extension has to listen for download completion, notify
+the native component, and ensure that the file does not change or move before
+the hash is calculated) and raises friction for users (due to the complexity of
+installation). Additionally, calculating file content hash separately is
+slightly less efficient since it requires reading it back into memory.
+
+### Open Web API
+
+There are no equivalent APIs on the Open Web. Open Web has the concept of
+subresource integrity, but it applies only to page resources like images and
+other media and does not apply to downloaded files.
+
+## Implementation Notes
+
+ 1. Hash algorithm descriptor matching must be case-sensitive (which differs
+    from some other Open Web APIs).
+ 2. User Agent may throw an error upon a call to `downloads.getFileHash()`
+    if the required hash is not available. User Agent may compute the hash on
+    the fly if the original file is still available.
+ 3. User Agents must not support obsolete hash algorithms, see "Additional
+    Security Considerations" for more details.
+ 4. User Agents must encode (binary) hash digest into a hex string, using
+    lower-case letters.
+
+## Future Work
+
+In theory, some use cases could be served via a declarative API. For example,
+an antivirus extension could provide a set of hashes of known bad content for
+the browser to use directly. However, the development of such a schema is
+outside of the scope of this proposal.


proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,152 @@
...
+Additionally, calculating file content hash separately is slightly less
+efficient since it requires reading it back into memory.
+
+### Open Web API
+
+There are no equivalent APIs on the Open Web. Open Web has the concept of
+subresource integrity, but it applies only to page resources like images and
+other media and does not apply to downloaded files.
+
+## Implementation Notes
```
bershanskiy, line 143: @Rob--W One more question: should the encoded hash be upper-case or lower-case? I have no particular preference, just would like the spec to be unambiguous.

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,152 @@
...
+enough to limit the spread of known malware downloads since effecively every
+client gets "innoculated" against a particular download as soon as a single
+antivirus detects a malicious file.
+User Agents may store file hashes even after the original file is deleted to
+facilitate tracing of malvare exposure even if the file is deemed to be
+dangerous after the user initially downloads it.
+
+##### Download integrity assurance
+
+Files can get damaged while they are sored on file hosting server or NAS.
```
Rob--W:
```suggestion
Files can get damaged while they are shared on file hosting server or NAS.
```

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,152 @@
...
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Hashes (cryptographic) are considered one-way operations, but they can easily
+be reversed to the original input when the number of possible inputs is
+sufficiently small. So extension may be able to infer the raw contents of the
+file if the possible space of file contents is small.
+A malicious extension could:
+ - Observe hashes of downloaded files and send them to a home server
```
Rob--W, line 105:
```suggestion
 - obtain the hash of the response to any URL.
```

The downloads API can create downloads without user interaction, so this effectively allows any extension to get the hash for any URL.

proposals/downloads_get_file_hash.md
Rob--W: Lowercase seems more common.

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,152 @@
...
+
+  getFileHash: (downloadId: number, algorithm?: downloads.FILE_HASH_ALGO) => Promise<string>;
+}
+```
+
+### New Permissions
+
+|      Permission Added     | Suggested Warning |
+| ------------------------- | ----------------- |
+| `"downloads.getFileHash"` | `"View cryptographic digests of downloads"` |
```
Rob--W: Open questions:
- do we want a permission warning? If not, no more questions. If yes, continue.
- is this the permission warning that we'd like to settle on? It is quite technical and I am not sure if the average user would know what they are consenting to.
- do we want this API permission for any API?
- do we want to enable extensions with host permissions to finalUrl to get the hash without additional permission warnings?

bershanskiy: These are great questions, let's discuss them separately:
 > - do we want a permission warning?

Yes, I believe every permission should have a corresponding permission "message", even if it not a "warning". I have an ambition to (eventually) make every permission optional and adding a toggle in browser-provided UI.

> - is this the permission warning that we'd like to settle on?

Better proposals are welcome. I would like to add a "learn more" UI of some form eventually.

> - do we want this API permission for any API?

> - do we want to enable extensions with host permissions to `finalUrl` to get the hash without additional permission warnings?

Yes, but I intentionally left this permission relaxation out of this proposal and first implementation in Chromium. I would like to eventually move to host-based permission model. Specifically, if extension has a host permission for a site, it should be granted `"tabs"`, `"scripting"`, `"downloads"`, `"favicon"`, and many other permissions for that site specifically.

proposals/downloads_get_file_hash.md
bershanskiy: I added point 4 "User Agents must encode (binary) hash digest into a hex string, using lower-case letters." Is this OK?

proposals/downloads_get_file_hash.md
bershanskiy: I added two points:
> - Initiate a download with an arbitrary URL and obtain a hash of the response
> - Initiate a download for a ranged HTTP request and obtain a hash of the
   corresponding portion of the response

proposals/downloads_get_file_hash.md
bershanskiy: Per WECG meeting minutes, we removed the permission warning and added requirement for host access to `DownloadItem.finalURL`. Should we mark this as resolved?

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,161 @@
...
+        long downloadId,
+        [legalValues=("SHA-256")] DOMString algorithm,
+        GetFileHashCallback callback);
+}
+```
+
+``` idl
+namespace downloads {
+  enum FILE_HASH_ALGO {
+    "SHA-256"
```
Rob--W: Not blocking, but I wonder how this is exposed as the constant, since there is no obvious approach as seen at https://github.com/w3c/webextensions/issues/563

- lowercase or upper case?
- what happens with the `-`?

Rob--W (approved):
Looks good from my perspective. Before merging this PR also needs approval from at least Chrome and Safari.

proposals/downloads_get_file_hash.md
bershanskiy: This is a good point and this proposal lacks clarity about enums. Here we actually have two hings to consider: the enum key and the enum value literal. This document currently does not specify the enum key explicitly. I copied enum value literal `"SHA-256"` from [Web Cryptography API §30.2 Registration](https://www.w3.org/TR/WebCryptoAPI/#sha-registration).

What do you think about this?
```idl
namespace downloads {
    enum FileHashAlgorithm {
        SHA256 = "SHA-256"
    }
}
```

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,163 @@
...
+
+    [supportsPromises, permissions="downloads.getFileHash"]
+    static void getFileHash(
+        long downloadId,
+        [legalValues=("SHA-256")] DOMString algorithm,
+        GetFileHashCallback callback);
+}
+```
+
+``` idl
```
oliverdunk: Nit: This is TypeScript right?

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+### New Permissions
+
+|      Permission Added     | Suggested Warning |
+| ------------------------- | ----------------- |
+| `"downloads.getFileHash"` | No warning        |
+
+The new `"downloads.getFileHash"` permission is optional and is honored only if
```
oliverdunk: > is honored only if extension has `"downloads"` permission as well

What was the thinking here? I think in general these sub-permissions work without the parent permissions so this might be inconsistent with downloads.open for example.

oliverdunk (commented):
Thanks Anton! This generally looks good to me, but I want to run this by folk internally before approving to make sure this matches what we expected.

proposals/downloads_get_file_hash.md
bershanskiy: Sub-permissions work a bit differently across browsers: Firefox requires both `"downloads"` and `"downloads.open"` to call `downloads.open()` method. I do not have a strong opinion on this. Personally, I would be more happy with the way Chrome does it (since more fine-grained permissions carry greater agency for the user/user-agent). Would be nice to discuss this during the call today.

proposals/downloads_get_file_hash.md
bershanskiy: Thanks, fixed!

oliverdunk: Thanks! The top one should still be `idl` I think?

bershanskiy: I changed it [back](https://github.com/w3c/webextensions/compare/9f55418fcf335142dac5377f233cda35b63ba189..2cfdc1baf93855ad9196b79d3ed5dc3517d77624), does it look OK now?

oliverdunk: Perfect!

xeenon (approved):

proposals/downloads_get_file_hash.md
Rob--W: I tested in Chrome and it behaves like Firefox: The `downloads` permission is required for access to the `downloads` namespace. The `downloads.open` permission has a separate permission warning, and unlocks the `downloads.open` method.

After having thought this through a bit, I think that the behavior as described is good enough. It is not only consistent with what we already have, the most compelling point is that the `downloads.getFileHash` method is of no use without access to the rest of the `downloads` API. In particular, the `downloadId` needed to identify specific downloads can only be obtained through the `downloads` permission.

oliverdunk: Thanks Rob, it seems like I was mistaken - I confirmed with Devlin and your description matches the intended behavior.

On the existence of the permission in general, we want to add permissions for new namespaces but not necessarily new APIs - as an example we have a `userScripts` permission but not one for each of the individual methods. The one exception is if the API is particularly more dangerous, but given the host permission requirement, Devlin and I don't feel that it meets that bar. Do you feel strongly on having a permission here / have a case for why this is significantly more dangerous than host permissions? If so I think there will be some additional work to align.

Rob--W: This permission predates my suggestion of relying on host permissions. With the use of host permissions, the only remaining value of the permission is to audit API usage more easily. The API is not distinctive enough (e.g. compared to fetch/XHR) to really need that though.

oliverdunk: Ok, great - in that case it seems like we're aligned on removing the permission. @bershanskiy, does that sound good to you?

oliverdunk: FYI @xeenon

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,164 @@
...
+extension has `"downloads"` permission as well, and the extension has host
```
Rob--W: @oliverdunk The discussion reminds me that not all URLs have an origin. E.g. data:-URLs. What is your recommendation for these cases?

In the case of data:-URLs specifically, since the data is already fully revealing its content, I don't think that any origin permissions are needed there.

oliverdunk: Good point! It looks like we do expose the full URL already there so I don't have any strong feelings. One thing that came up in discussion about this proposal is that we want to run it by the downloads API owners (which is not the general extensions team) and privacy / security review. I'm going to drive that but suspect this may be something where they will have more thoughts.

bershanskiy:
I rebased the PR [here](https://github.com/w3c/webextensions/compare/2cfdc1baf93855ad9196b79d3ed5dc3517d77624..57c0ca994058f858c7d78eec6bff02c95f5a098a) and incorporated the feedback [here](https://github.com/w3c/webextensions/compare/57c0ca994058f858c7d78eec6bff02c95f5a098a..591f58d1b321e9bc9682696d64e2df4bcb75c4c1).

proposals/downloads_get_file_hash.md
bershanskiy: I removed `"downloads.getFileHash"` permission [here](https://github.com/w3c/webextensions/compare/57c0ca994058f858c7d78eec6bff02c95f5a098a..591f58d1b321e9bc9682696d64e2df4bcb75c4c1). Please mark as "resolved" if this works OK.

proposals/downloads_get_file_hash.md
bershanskiy: @oliverdunk Did you have a chance to discuss this internally?

oliverdunk: I confirmed with the downloads API owners last week that they are comfortable with this change. I haven't yet been able to go through privacy/security review though, which I'll do as soon as possible.

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,158 @@
...
+file if the possible space of file contents is small.
+A malicious extension could:
+ - Observe hashes of downloaded files and send them to a home server
+ - Initiate a download with an arbitrary URL and obtain a hash of the response
+ - Initiate a download for a ranged HTTP request and obtain a hash of the
+   corresponding portion of the response
+
+### Abuse Mitigations
+
+The API method requires host access to the`DownloadItem.finalUrl`.
```
Rob--W:
```suggestion
The API method requires host access to the `DownloadItem.finalUrl`.
```

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,158 @@
...
+    downloadId: number,
+    algorithm?: keyof downloads.FileHashAlgorithm)
+      => Promise<string>;
+}
+```
+
+### New Permissions
+
+No new permission. `downloads.getFileHash()` is honored only if extension has
+`"downloads"` permission and the host access for `DownloadItem.finalUrl`.
```
Rob--W:
```suggestion
`"downloads"` permission and the host access for `DownloadItem.finalUrl`.
For data:-URLs, host permissions are not required, because the content of the file is already fully visible through the URL.
```

Added extra remark about data:-urls.

proposals/downloads_get_file_hash.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Manifest File Changes
+
+No changes besides the new optional permission `"downloads.getFileHash"`.
```
Rob--W:
```suggestion
No changes.
```

Dropped mention of the permission that was present in an earlier draft but not the final version.

Rob--W (approved):

proposals/downloads_get_file_hash.md
bershanskiy: [Done](https://github.com/w3c/webextensions/compare/591f58d1b321e9bc9682696d64e2df4bcb75c4c1..ddf0ca8fb4737dcd78a9b58409ea6cf7d7c7c901)

proposals/downloads_get_file_hash.md
bershanskiy: [Done](https://github.com/w3c/webextensions/compare/591f58d1b321e9bc9682696d64e2df4bcb75c4c1..ddf0ca8fb4737dcd78a9b58409ea6cf7d7c7c901)

proposals/downloads_get_file_hash.md
bershanskiy: [Done](https://github.com/w3c/webextensions/compare/591f58d1b321e9bc9682696d64e2df4bcb75c4c1..ddf0ca8fb4737dcd78a9b58409ea6cf7d7c7c901)

Rob--W:
@oliverdunk Chromium is listed as a supporting browser. Could you sign off on this PR to confirm that the proposal is in a good shape and that you agree to sponsor this proposal?

After approval this can be merged (no need to squash because there is only one commit).

oliverdunk:
> @oliverdunk Chromium is listed as a supporting browser. Could you sign off on this PR to confirm that the proposal is in a good shape and that you agree to sponsor this proposal?

I'd like to get through security and privacy review internally before approving this - I've started that process but haven't heard back yet. I'll update this PR as soon as I do.

dotproto:
WECG F2F note: We should make sure that no browsers have security concerns. If nothing else, this means Chrome needs to go through their Security & Privacy Review process.

Closing as no browsers have active plans to implement this feature. We are generally supportive, though, and are open to revisiting if a contributor is interested.

PR #614, opened 2024-05-14
Use a plural array type like `"permissions"` for simplicity

solomonkinard:


dotproto (approved):
LGTM

xeenon (approved):

Rob--W (approved):

mukul-p (approved):

carlosjeurissen:
There is one "color-schemes" instead of "color_schemes" in the commit.

Basically with this PR the ability to set the color scheme to a single string value is dropped?

solomonkinard:
This PR was created a week before the previous Tuesday more as a question than a directive. I like the idea of `color_scheme` accepting string or string[]. However, I was thinking about "permissions" as a string[] when I wrote this PR. This is not necessarily set in stone here, so it could have been discussed in this thread or in the meeting today before merge. Even still, it's not too late to make a case for something other than `color_schemes` as a string[]. Some ideas:
1. Revert this change to return to `color_scheme`: string | string[]
2. Keep this change which supports `color_schemes`: string[].
3. Support `color_schemes`: string | string[].
4. Support {`color_scheme`: string, `color_schemes`: string[]}.

Maybe you have other ideas beyond those. Perhaps you could share your ideas here or better yet create a PR with your preference and rationale.

carlosjeurissen:
Thinking out loud the most common use case seems to be either set `color_scheme` to "light" or "dark", or omit it to indicate it is for all color schemes.

Requiring an array would mean the syntax is more verbose with the benefit of having a single syntax for any situation and indicate other color_schemes can be added here.

In any case, I feel we either go with this PR, always use `color_schemes` which requires an array (string[]). OR keep as was before (`color_scheme` as string | string[]).

Would love to hear about other contributors who helped decide on this before @hanguokai @xeenon

PR #629, opened 2024-06-05
Fix a warning about Node version in deploy workflow

danielhjacobs:


dotproto:
@danielhjacobs, can you share more about what warning you observed and how this change addresses it? Is there a known issue in actions/checkout that might have more info?

danielhjacobs:
See https://github.com/w3c/webextensions/actions/runs/9398624429:


See https://github.com/actions/checkout/releases/tag/v4.0.0

danielhjacobs:
Also see https://github.com/actions/checkout/releases/tag/v3.0.0, which updated Node 12 to Node 16, plus see the links from the action, https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/ and https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

danielhjacobs:
Do you want any more info on this? In my opinion it's generally not a good idea to rely on the auto-upgrade behavior of the Node versions for GitHub actions when newer versions of the actions exist.

danielhjacobs:
The new timeline for Node16 to reach end of life in actions is June 30, 2024, according to https://github.blog/changelog/2024-05-17-updated-dates-for-actions-runner-using-node20-instead-of-node16-by-default/

xeenon (approved):

Issue #258, opened 2022-08-19
Per-extension language preferences


hanguokai:
**Update**: In order to express it completely and clearly, I reorganized the proposal and edited it many times after 2022-09-12.

## Summary
Currently, `browser.i18n` only display one default language to users. Users can't change the language to other extension supported languages independently. Developers need to use their own solutions to provide users with multiple language select menu.

It makes sense for an app to use another language independently of the operating system. For example, Android supports per-app language preferences (here is its [docs](https://developer.android.com/guide/topics/resources/app-languages) and [video](https://www.youtube.com/watch?v=DUKnNWwcNvo)). The same goes for extensions. This proposal brings per-app language preferences to browser extensions.

Tracking bugs: [Chromium-1365283](https://bugs.chromium.org/p/chromium/issues/detail?id=1365283) , [Firefox-1791356](https://bugzilla.mozilla.org/show_bug.cgi?id=1791356)

### 1. Browser built-in supply a language select menu per extension
<img width="331" alt="Screen Shot 2022-09-04 at 14 30 17" src="https://user-images.githubusercontent.com/776467/189578313-e1125407-ffa4-489b-b689-c525a9859698.png">

Since this is a general purpose feature for all users and extensions, browser built-in support is best.

If browsers built-in support a language select menu per extensions, all developers, all existing extensions and users can get this benefit right away. Ideally, developers just use current i18n api without doing anything. Another benefit is that it's much easier for developers to test i18n.

### 2. New APIs for developers
```
/**
 * get the current language used by i18n.getMessage()
 * return the language tag,  e.g. en-US, zh-CN.
 */
i18n.getCurrentLanguage() => code: String.

/**
 * set a new language used in this extension.
 * if code is null, revert to the default state(follow the browser UI language).
 * if code is not valid, reject it.
 * return a Promise, resolved when the operation is complete.
 */
i18n.setCurrentLanguage(code: String) => Promise<void>

/**
 * get all languages that supported by this extensions.
 * return a Promise, resolved with an array of language tags.
 */
i18n.getAllLanguages() => code_array: Promise<Array<String>>

/**
 * After i18n changed to a new language, the browser triggers a language changed event.
 * callback is (new_language_code: String) => void
 */
i18n.onLanguageChanged.addListener(callback)
```
[...]

Issue #296, opened 2022-10-17
`i18n.getMessage()` language fallback paths


carlosjeurissen:
Not all browsers handle language fallbacks the same. Considering the following situation:

An extension is using the native i18n APIs with `"default_locale": "en"` in `manifest.json`, and three `messages.json` files in the languages `en`, `pt` and `pt-BR`.

Both `en` and `pt` include the message ids `message1` and `message2`. While `pt-BR` includes only `message1`.

In the above situation, browsers handle fetching `i18n.getMessage('message2')` different.

Chromium first checks `pt_BR/messages.json`, if the message is not present, it checks `pt/messages.json`, and finally, if the message is still not found, it will check the `default_locale`, in this case `en/messages.json`. In the above situation, this means it gets the message2 value from `pt`.

In Firefox, however, the browser first checks `pt_BR/messages.json`. If the message is not in this file, it will directly fallback to `default_locale`. so it checks `en/messages.json`. Resulting in message2 value becomes the one from `en`.
Edit: The behaviour of Firefox has been changed and now aligns with Safari and Chrome.

Interestingly enough, in Firefox, if `pt_BR/messages.json` is not present in general, it will check `pt/messages.json` first, before checking `en/messages.json`.

What is the behaviour we want in these cases?

hanguokai:
I support Chrome's [behavior](https://developer.chrome.com/docs/extensions/reference/i18n/#searching-for-messages), which seems more reasonable. Usually, developers want `language-region` locale to fallback to `language` locale first, then the default locale.

If the browser wants to support multiple different behaviors at the same time, I recommend add a new property in the 3rd parameter(options) in this api.

carlosjeurissen:
@hanguokai generally speaking this can be useful to reduce the overall package size.

However, there are cases when the fallback might not always be welcome. Say your `zh/messages.json` is in Simplified Chinese script, and `zh_TW` is in Traditional Chinese. Would it be great if a different script is used as fallback? Same can happen with other languages with multiple scripts, like Serbian (Latin and Cyrillic).

hanguokai:
I know the difference, it's better than the default(complete another language like English). For example,
```
en: "Software"
zh: "软件"
zh_TW: "軟體"
```
The difference between "软件" and "軟體" is smaller than that of English.

For the best user experience, developers need to supply full message map(1:1) if they are different. Only when they are the same or acceptable, they can be omitted.

carlosjeurissen:
[...]

Issue #131, opened 2021-11-03
Documentation: get a complete list of all locales supported in each browser


carlosjeurissen:
To make the extensions available to as much users as possible, extension developers need to know what languages and language-variants each browser supports.

Chrome for example seems to support both `es` and `es_419`. What we need is a complete list for each browser and potentially OS which includes all locale/language codes (like `es_419`) the browser supports using the `_locales/language_code/messages.json` and `i18n.getMessage` APIs.

The request is not about accept-languages or what languages the actual extension stores support.

carlosjeurissen:
Overview:
https://docs.google.com/spreadsheets/d/1N0f9Nl-PM_nafpJEfYbS8xV9Qvb6o6DvuBFWLTXRr2Y/edit#gid=1518318975

PR #641, opened 2024-06-18
Proposal: Per-extension language preferences

hanguokai:
This proposal has been discussed a lot in #258 and is supportive by all browsers. This is the formal version of the proposal. I followed the latest proposal template.

Welcome everyone to review this proposal. I suggest focusing on the main issues first, followed by minor issues, and finally fixing textual typos.

### DIFF
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -1,0 +1,234 @@
+# Proposal: Per-extension language preferences
+
+**Summary**
+
+This proposal allows developers (users) to set a specific language for their extension (which may be different from the browser or system default language) and create a language selection menu for users.
+
+**Document Metadata**
+
+**Author:** hanguokai
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** 
+
+**Created:** 2024-06-16
+
+**Related Issues:** [#258](https://github.com/w3c/webextensions/issues/258)
+
+## Motivation
+
+### Objective
+
+Prior to this proposal, browsers automatically selected a `browser.i18n` language for extensions, and developers (users) could not set a specific language.
+
+For multilingual users, sometimes they want to set a specific application (extension) to another language that is different from the system (browser). This need is not uncommon.
+[Android 13](https://developer.android.com/guide/topics/resources/app-languages), [iOS 13 and macOS Catalina](https://developer.apple.com/news/?id=u2cfuj88) support a feature
+called "Per-app language preferences (settings)". This proposal brings the same function to browser extensions.
+
+#### Use Cases
+
+- Extensions create a language selection menu for users.
+- [Optional] Browsers could potentially provide a built-in language selection menu for users, just like Android and iOS.
+- [Optional] Extensions could integrate third-party i18n implementations (frameworks) with the browser's built-in language selection menu.
+
+### Known Consumers
+
+This is a generic i18n feature, which means that all types of extensions can use it. Some extensions already provide this functionality via workarounds or non-browser.i18n implementations.
+
+## Specification
+
+### Definitions
+
+In this document:
+- **locale** or **language code** or **language tag** is a string that represent a language, defined in [BCP 47](https://www.rfc-editor.org/info/bcp47).
+For example, `en-US`, `zh-CN`, `fr`. It is used by `browser.i18n`, `Date`, `Intl` and various other APIs.
+- **the user preferred language** is the language set by `i18n.setCurrentLanguage(code)`. The user may not have set it.
+- **the extension displayed language** is the language which the extension is displayed in. Note that the extension displayed language may be different from the browser UI language. For example, if the user preferred language is not set, the browser UI language is English, and the extension only supports French (default) and Japanese, then the extension displayed language is French because `i18n.getMessage()` returns French message at this situation.
+
+### Schema
+
+```ts
+/**
+ * Get the extension displayed language.
+ * return the language tag, e.g. en-US.
+ */
+i18n.getCurrentLanguage() : string
+
+/**
+ * Set the user preferred language.
+ * if code is null, revert to the unset state.
+ * if code is not valid, return a rejected Promise.
+ * else return a Promise resolved when the operation is complete.
+ */
+i18n.setCurrentLanguage(code: string) : Promise<void>
+
+/**
+ * Get all languages that supported by this extension.
+ * return a Promise resolved with an array of language tags.
+ */
+i18n.getAllLanguages(): Promise<string[]>
+
+/**
+ * After changing to a new language, the browser triggers a language changed event.
+ * The callback is function(code: string) : void.
+ * The code parameter in the callback is the new language.
+ */
+i18n.onLanguageChanged.addListener(callback)
+```
+
+### Behavior
+
+#### Behavior of `i18n.getCurrentLanguage()`
+
+This method return the language that the extension is displayed in.
+
+- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return `undefined`.
+- If the preferred language is not set by `i18n.setCurrentLanguage()`, returns the current language used by `i18n.getMessage()`, assuming that all languages support all possible keys.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, and the extension supports this language, then return this language.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language (no message file for this language), then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
+
+##### Use Case 1: use the extension displayed language with other locale-related APIs.
+
+For example, if the preferred language is not set, the language of the browser UI is English (en-US), the extension supports French (default) and Japanese, then `i18n.getCurrentLanguage()` return `"fr"` because `i18n.getMessage()` return French message at this situation.
+
+```js
+// use current extension displayed language to format date
+const locale = browser.i18n.getCurrentLanguage();
+const today_date = new Date().toLocaleString(locale); // format date string in 'fr'
+
+// if don't use current extension displayed language, it may be formatted in other locale
+const today_date = new Date().toLocaleString(); // format date string in 'en-US'
+```
+
+##### Use Case 2: show the current language in the extension language selection menu.
+
+#### Behavior of `i18n.setCurrentLanguage(code)`
+
+This method sets the user preferred language for this extension.
+
+- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return a rejected Promise.
+- If the code is an invalid language tag or an unsupported language by this extension, return a rejected Promise.
+- If the code is null, revert to the unset state.
+- else do the following:
+  1. save the language persistently, and make sure that the language is prioritized for use next time.
+  1. return a resolved Promise.
+  1. trigger `i18n.onLanguageChanged` event.
+
+In addition, when `i18n.setCurrentLanguage(code)` success:
+- the browser should update related browser UI and extension UI, because some values in manifest.json may be changed, like `name`, `short_name`, `description` and `action.default_title`.
+- Text from `i18n.getMessage()` and css files don't update. These are handled by developers.
+
+##### Persistence of `i18n.setCurrentLanguage(code)`
+
+This setting is persistent when:
+- restart or upgrade the browser.
+- disable then re-enable the extension.
+- upgrade the extension to a new version.
+
+#### Behavior of `i18n.getAllLanguages()`
+
+This method returns an array of language tags that the extension supported, e.g. `["en", "fr", "ja", "zh-CN"]`.
+This is a convenient method. Without this method, developers need to hardcode it into their code.
+
+#### Behavior of `i18n.onLanguageChanged` event
+
+When the user preferred language is changed to a new different language by `i18n.setCurrentLanguage(code)`, this event is fired.
+Developers use this event to know the language changed, and might do the following:
+
+- Update the content of extension pages in place (without refreshing the page)
+- Reload extension pages:
+  - If the page is stateless, refresh the page directly.
+  - If the page is stateful, save the current state and then refresh the page.
+  - Prompts the user that the language setting has changed, and asks the user if they want to refresh the page immediately.
+- Re-create the extension context menus, since the title of the context menus shoule be updated.
+- Update the action title and badge text if needed.
+
+#### Behavior of `i18n.getMessage()`, manifest and CSS files
+
+When this proposal is implemented, the behavior of these existing functions will change as follows:
+
+- If the preferred language is not set by `i18n.setCurrentLanguage()`, it is consistent with the existing behavior.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, and the extension supports this language, then prioritize using that language.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language, then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
+
+### Browser built-in language selection menu
+
+Based on the capabilities provided by this proposal, browsers could provide a unified built-in language selection menu for extensions, like Android and iOS.
+Whether to provide the built-in language selection menu and how to implement it is up to the browser to decide. This is just a suggestion.
+
+The built-in language selection menu has the following benefits:
+- Easy for developers: Developers do not need to implement it themselves, they only need to adapt to this proposal.
+- Easy for users: A unified UI makes it easier for users to use, otherwise each extension might be set up differently.
+- It is convenient for developers to develop and test i18n functions by switching the extension language.
+
+### Integrate other i18n implementations (frameworks) with browser.i18n
+
+This proposal can also be integrated with third-party i18n frameworks, as these frameworks typically allow developers to specify the language to be used.
+For example, developers use `i18n.getCurrentLanguage()` and `i18n.setCurrentLanguage()` for the user preferred language, but don't use `i18n.getMessage()`.
+
+```js
+otherI18nFramework.setLanguage(browser.i18n.getCurrentLanguage());
+
+let text = otherI18nFramework.getMessage(key);
+```
+
+### New Permissions
+
+N/A
+
+### Manifest File Changes
+
+If the browser would like to support the built-in language selection menu for extensions, the manifest file should add a new key for developers to opt-in this feature, like the following:
+```
+{
+  "name": "__MSG_extName__",
+  "default_locale": "en",
+  "builtin_languages_menu": true | false(default)
+}
+```
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+N/A
+
+### Abuse Mitigations
+
+N/A
+
+### Additional Security Considerations
+
+N/A
+
+## Alternatives
+
+### Existing Workarounds
+
+##### Workaround-1: fetch message files
+
+This is not ideal. Developers need to solve many problems on their own, such as saving the preferred language, implementing placeholders, and fallback mechanisms.
+
+##### Workaround-2: use other i18n frameworks or implement one yourself
+
+There are some JavaScript i18n frameworks that provide similar features, which allow developers to specify the language they want to use.
+In fact, this workaround is equivalent to asking developers to give up using `browser.i18n`.
+In addition, developers need to save the preferred language by themself.
+
+##### Workaround-3: mix browser.i18n and other workarounds.
+
+Because some text can only be localized by `browser.i18n`, such as the extension name and description, developers often mix different implementations.
+
+### Open Web API
+
+The Web provides some related tools, such as `Intl`, but there is no unified framework that provides functionality such as `browser.i18n.getMessage()`.
+There are some third-party i18n frameworks, but they use custom mechanisms rather than the `browser.i18n` mechanism.
+
+## Implementation Notes
+
+When the extension is upgraded, if the new version removes the language that was set by `i18n.setCurrentLanguage(code)`, the user preferred language should be reverted to the unset state.
+
+## Future Work
+
+N/A
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -196,7 +196,7 @@ N/A
 
 ### Abuse Mitigations
 
-N/A
+A case of abuse is changing the extension name/description dynamically by `i18n.setCurrentLanguage()`. This is not a security threat because all metadata published in the extension store is audited. And for the extension management page (e.g. chrome://extensions/), the extension name should be displayed in the same language as the management page (i.e. the Browser UI language), so it is not affected by `i18n.setCurrentLanguage()`.
 
 ### Additional Security Considerations
 
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -81,14 +81,14 @@ i18n.onLanguageChanged.addListener(callback)
 
 #### Content Scripts Availability
 
-In history, all `browser.i18n` APIs were available in content scripts. But some of methods in this proposal should not be available in content scripts because of their purpose.
+Unlike other APIs, all `browser.i18n` APIs are historically available to content scripts. To counter abuse, some of the features in this proposal should not be available in content scripts.
 
-| Method  | Content Scripts Availability |
+| API  | Content Scripts Availability |
 | ------------- | ------------- |
 | `getCurrentLanguage()` | Yes |
 | `setCurrentLanguage()` | No |
 | `getAllLanguages()` | No |
-| `onLanguageChanged()` | Yes |
+| `onLanguageChanged` | Yes |
 
 #### A New Predefined Message `@@current_locale`
 There is an exsiting predefined message `@@ui_locale`, that reflects the value of `i18n.getUILanguage()`, but the value uses underscore (e.g. "en_US") as separator in Chrome and hyphen (e.g. "en-US") in Firefox.
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -116,7 +116,9 @@ This method sets the user preferred language for this extension.
   1. trigger `i18n.onLanguageChanged` event.
 
 In addition, when `i18n.setCurrentLanguage(code)` success:
-- the browser should update related browser UI and extension UI, because some values in manifest.json may be changed, like `name`, `short_name`, `description` and `action.default_title`.
+- Because some values in manifest.json may be changed, like `name`, `short_name`, `description` and `action.default_title`:
+  - The browser should update related extension UI, e.g. `action.default_title`.
+  - The browser might update related browser UI, but this part is up to the browser to decide.
 - Text from `i18n.getMessage()` and css files don't update. These are handled by developers.
 
 ##### Persistence of `i18n.setCurrentLanguage(code)`
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -196,7 +196,7 @@ N/A
 
 ### Abuse Mitigations
 
-A case of abuse is changing the extension name/description dynamically by `i18n.setCurrentLanguage()`. This is not a security threat because all metadata published in the extension store is audited. And for the extension management page (e.g. chrome://extensions/), the extension name should be displayed in the same language as the management page (i.e. the Browser UI language), so it is not affected by `i18n.setCurrentLanguage()`.
+A case of abuse is changing the extension name/description dynamically by `i18n.setCurrentLanguage()` in an attempt to confuse the user and hinder their ability to identify the extension. A mitigation to this issue is to ignore changes from `setCurrentLanguage` in the extension management parts of the browser UI, and always use the browser's configured UI language instead. This includes extension management pages (e.g. `chrome://extensions` in Chrome and `about:addons` in Firefox) and permission dialogs.
 
 ### Additional Security Considerations
 
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -152,16 +152,6 @@ When this proposal is implemented, the behavior of these existing functions will
 - If the preferred language is set by `i18n.setCurrentLanguage()`, and the extension supports this language, then prioritize using that language.
 - If the preferred language is set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language, then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
 
-### Browser built-in language selection menu
-
-Based on the capabilities provided by this proposal, browsers could provide a unified built-in language selection menu for extensions, like Android and iOS.
-Whether to provide the built-in language selection menu and how to implement it is up to the browser to decide. This is just a suggestion.
-
-The built-in language selection menu has the following benefits:
-- Easy for developers: Developers do not need to implement it themselves, they only need to adapt to this proposal.
-- Easy for users: A unified UI makes it easier for users to use, otherwise each extension might be set up differently.
-- It is convenient for developers to develop and test i18n functions by switching the extension language.
-
 ### Integrate other i18n implementations (frameworks) with browser.i18n
 
 This proposal can also be integrated with third-party i18n frameworks, as these frameworks typically allow developers to specify the language to be used.
@@ -179,14 +169,7 @@ N/A
 
 ### Manifest File Changes
 
-If the browser would like to support the built-in language selection menu for extensions, the manifest file should add a new key for developers to opt-in this feature, like the following:
-```
-{
-  "name": "__MSG_extName__",
-  "default_locale": "en",
-  "builtin_languages_menu": true | false(default)
-}
-```
+N/A
 
 ## Security and Privacy
 
@@ -231,4 +214,21 @@ When the extension is upgraded, if the new version removes the language that was
 
 ## Future Work
 
-N/A
+### Browser built-in language selection menu
+
+Based on the capabilities provided by this proposal, browsers could provide a unified built-in language selection menu for extensions, like Android and iOS.
+Whether to provide the built-in language selection menu and how to implement it is up to the browser to decide.
+
+The built-in language selection menu has the following benefits:
+- Easy for developers: Developers do not need to implement it themselves, they only need to adapt to this proposal.
+- Easy for users: A unified UI makes it easier for users to use, otherwise each extension might be set up differently.
+- It is convenient for developers to develop and test i18n functions by switching the extension language.
+
+If the browser would like to support the built-in language selection menu for extensions, the manifest file should add a new key for developers to opt-in this feature, like the following:
+```
+{
+  "name": "__MSG_extName__",
+  "default_locale": "en",
+  "builtin_languages_menu": true | false(default)
+}
+```
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -85,8 +85,8 @@ This method return the language that the extension is displayed in.
 
 - If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return `undefined`.
 - If the preferred language is not set by `i18n.setCurrentLanguage()`, returns the current language used by `i18n.getMessage()`, assuming that all languages support all possible keys.
-- If the preferred language is set by `i18n.setCurrentLanguage()`, and the extension supports this language, then return this language.
-- If the preferred language is set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language (no message file for this language), then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
+- If the preferred language has been set by `i18n.setCurrentLanguage()`, and the extension supports this language, then return this language.
+- If the preferred language has been set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language (no message file for this language), then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
 
 ##### Use Case 1: use the extension displayed language with other locale-related APIs.
 
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -107,7 +107,6 @@ const today_date = new Date().toLocaleString(); // format date string in 'en-US'
 
 This method sets the user preferred language for this extension.
 
-- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return a rejected Promise.
 - If the code is an invalid language tag or an unsupported language by this extension, return a rejected Promise.
 - If the code is null, revert to the unset state.
 - else do the following:
--- a/proposals/per-extension-language-preferences.md
+++ b/proposals/per-extension-language-preferences.md
@@ -79,6 +79,17 @@ i18n.onLanguageChanged.addListener(callback)
 
 ### Behavior
 
+#### Content Scripts Availability
+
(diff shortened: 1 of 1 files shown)


xPaw:
Looks good to me, although I would like this proposal to also cover a getMessage overload (or separate method) to get a message in a specific language.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+ * if code is not valid, return a rejected Promise.
+ * else return a Promise resolved when the operation is complete.
+ */
+i18n.setCurrentLanguage(code: string) : Promise<void>
+
+/**
+ * Get all languages that supported by this extension.
+ * return a Promise resolved with an array of language tags.
+ */
+i18n.getAllLanguages(): Promise<string[]>
```
oliverdunk, line 70: I'm not sure if this one makes sense, given that extensions should already know the languages they support. I think this gets tricky in certain situations (partial strings, parent language tags) so I wonder if we should leave this out. I see that having it provides some convenience but I'm not sure it's essential for the MVP.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+### Behavior
+
+#### Behavior of `i18n.getCurrentLanguage()`
+
+This method return the language that the extension is displayed in.
+
+- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return `undefined`.
+- If the preferred language is not set by `i18n.setCurrentLanguage()`, returns the current language used by `i18n.getMessage()`, assuming that all languages support all possible keys.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, and the extension supports this language, then return this language.
+- If the preferred language is set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language (no message file for this language), then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
```
oliverdunk: Should we allow you to get into this state? I would've expected `setCurrentLanguage` to throw an error and abort updating the language.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+const today_date = new Date().toLocaleString(); // format date string in 'en-US'
+```
+
+##### Use Case 2: show the current language in the extension language selection menu.
+
+#### Behavior of `i18n.setCurrentLanguage(code)`
+
+This method sets the user preferred language for this extension.
+
+- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return a rejected Promise.
```
oliverdunk: Do we need this line? I think the line below covers it and I think that aligns more closely with how this would be implemented in practice.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+- If the code is an invalid language tag or an unsupported language by this extension, return a rejected Promise.
+- If the code is null, revert to the unset state.
+- else do the following:
+  1. save the language persistently, and make sure that the language is prioritized for use next time.
+  1. return a resolved Promise.
+  1. trigger `i18n.onLanguageChanged` event.
+
+In addition, when `i18n.setCurrentLanguage(code)` success:
+- the browser should update related browser UI and extension UI, because some values in manifest.json may be changed, like `name`, `short_name`, `description` and `action.default_title`.
```
oliverdunk: I want to confirm this with the team. I suspect we might not want to translate everything, e.g certain parts of chrome://extensions, since those should show in the browser language.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+let text = otherI18nFramework.getMessage(key);
+```
+
+### New Permissions
+
+N/A
+
+### Manifest File Changes
+
+If the browser would like to support the built-in language selection menu for extensions, the manifest file should add a new key for developers to opt-in this feature, like the following:
```
oliverdunk: I think we should chat about this with other browser vendors, and only keep this section if at least one vendor plans to implement it.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Summary**
+
+This proposal allows developers (users) to set a specific language for their extension (which may be different from the browser or system default language) and create a language selection menu for users.
+
+**Document Metadata**
+
+**Author:** hanguokai
+
+**Sponsoring Browser:** Chromium
```
oliverdunk, line 11: I forget if we had already agreed this - was there a discussion somewhere? If not I will take it to the team.

oliverdunk (commented):
Thanks Jackie! I just left some initial thoughts :)

proposals/per-extension-language-preferences.md
hanguokai: @oliverdunk [a meeting note](https://github.com/w3c/webextensions/blob/658a794ac59f014e77a932593950c4188e2ca3a4/_minutes/2024-05-09-wecg.md?plain=1#L135) said this.

oliverdunk: Cool, let me confirm then. I definitely like the change, I'm just being extra careful any time we add ourselves as the sponsoring browser since we agreed on some strict rules about that meaning we will implement in a timely fashion.

proposals/per-extension-language-preferences.md
hanguokai: > extensions should already know the languages they support

Yes. It is just a convenient method. And I think it is easy to implement.

> in certain situations (partial strings, parent language tags)

_Languages with regional subtags_ are not a problem for users. For example, a language selection menu that includes French, French (Canada), French (Belgium) and French (France). This method should return all languages if they are there.

For partial strings, from the platform's standpoint, the platform does not need to consider these, but assumes that all these languages are supported.

proposals/per-extension-language-preferences.md
hanguokai: It depends on the implementation.  Although `setCurrentLanguage()` throw an error, but as I said in the "Implementation Notes" section: When an extension is upgraded, the browser should check to see if the languages it supports has changed (especially in the case of deletion). If the browser does this check, it will not get into this state, otherwise it may happen.

From a specification perspective, I'm just listing this possibility. From an implementation perspective, browsers just need to make sure that this problem can be avoided.

proposals/per-extension-language-preferences.md
hanguokai: From an implementation perspective, this is a bit redundant. But, from a specification perspective, this situation is slightly different from the following situation.

proposals/per-extension-language-preferences.md
hanguokai: For browser UI (like chrome://extensions), it makes sense to use the browser UI language if the extension supports.
For extension UI, e.g. `action.default_title`, I think the browser should update it. I will update this sentence.

proposals/per-extension-language-preferences.md
carlosjeurissen: @oliverdunk That is to say the browser per definition supports all the language tags specified in the extension. I suggest we change the method name to `i18n.getAvailableLanguages` to let the browser return all the languages the extension can call `setCurrentLanguage` to.

It is not just out of convenience. For example, for the Whale Store I have to exclude language tags with script variations. (zh-Hant, sr-Latn). Thus those are removed from `_locales`. Having to specify a list of all supported languages for each browser and store in every area of the extension this is used is not as straight forward as it might sound.

carlosjeurissen, line 70: Awaiting resolution of https://github.com/w3c/webextensions/issues/642, we might want to consider returning an object with more than just the code. Think about something like:
```json
[{
  "code": "en-US",
  "path": "en_US",
  "current": true,
  "default": true
}]
```

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+ */
+i18n.getCurrentLanguage() : string
+
+/**
+ * Set the user preferred language.
+ * if code is null, revert to the unset state.
+ * if code is not valid, return a rejected Promise.
+ * else return a Promise resolved when the operation is complete.
+ */
+i18n.setCurrentLanguage(code: string) : Promise<void>
```
carlosjeurissen, line 64: Should this method accept codes with underscores? See also https://github.com/w3c/webextensions/issues/642 To align with the web Intl standards I suggest we throw an exception in this case.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+For example, `en-US`, `zh-CN`, `fr`. It is used by `browser.i18n`, `Date`, `Intl` and various other APIs.
+- **the user preferred language** is the language set by `i18n.setCurrentLanguage(code)`. The user may not have set it.
+- **the extension displayed language** is the language which the extension is displayed in. Note that the extension displayed language may be different from the browser UI language. For example, if the user preferred language is not set, the browser UI language is English, and the extension only supports French (default) and Japanese, then the extension displayed language is French because `i18n.getMessage()` returns French message at this situation.
+
+### Schema
+
+```ts
+/**
+ * Get the extension displayed language.
+ * return the language tag, e.g. en-US.
```
carlosjeurissen, line 54:
```suggestion
 * return a [BCP47 language tag](https://www.rfc-editor.org/bcp/bcp47.html), e.g. en-US.
```

proposals/per-extension-language-preferences.md
carlosjeurissen:
```suggestion
- If the preferred language has been set by `i18n.setCurrentLanguage()`, and the extension supports this language, then return this language.
- If the preferred language has been set by `i18n.setCurrentLanguage()`, but the extension doesn't support this language (no message file for this language), then treat as if the preferred language is not set. This is an edge case, for example, the language was removed when the extension was upgraded.
```

Changing "is set" to "has been set" to prevent the confusion mentioned here: https://github.com/w3c/webextensions/pull/641/files#r1644443944

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+### Behavior
+
+#### Behavior of `i18n.getCurrentLanguage()`
+
+This method return the language that the extension is displayed in.
+
+- If the extension doesn't use `browser.i18n` (there is no "_locales" directory), return `undefined`.
+- If the preferred language is not set by `i18n.setCurrentLanguage()`, returns the current language used by `i18n.getMessage()`, assuming that all languages support all possible keys.
```
carlosjeurissen, line 103:
```suggestion
- If the preferred language has not been set by `i18n.setCurrentLanguage()`, returns the first language for which a message file exists following the same fallback mechanism used by `i18n.getMessage()
```

As a sidenote, the fallback mechanism is not the same across browsers. See: https://github.com/w3c/webextensions/issues/296

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+
+The built-in language selection menu has the following benefits:
+- Easy for developers: Developers do not need to implement it themselves, they only need to adapt to this proposal.
+- Easy for users: A unified UI makes it easier for users to use, otherwise each extension might be set up differently.
+- It is convenient for developers to develop and test i18n functions by switching the extension language.
+
+### Integrate other i18n implementations (frameworks) with browser.i18n
+
+This proposal can also be integrated with third-party i18n frameworks, as these frameworks typically allow developers to specify the language to be used.
+For example, developers use `i18n.getCurrentLanguage()` and `i18n.setCurrentLanguage()` for the user preferred language, but don't use `i18n.getMessage()`.
```
carlosjeurissen, line 175: Does this mean a developer would not have to specify `messages.json` files to inform the browser what languages it supports? If so this would be inconsistent with the `i18n.getCurrentLanguage()` behaviour specified above. Would we want an alternative way to specify the supported languages by the extension in this case? A manifest key like `supported_locales` which specifies a list of BCP47 language tags stating all the languages it supports.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+N/A
+
+### Manifest File Changes
+
+If the browser would like to support the built-in language selection menu for extensions, the manifest file should add a new key for developers to opt-in this feature, like the following:
+```
+{
+  "name": "__MSG_extName__",
+  "default_locale": "en",
+  "builtin_languages_menu": true | false(default)
```
carlosjeurissen: Why would we want this to be an opt-in feature? The only downside I see here is changing the language using a built-in languages menu could lead to extension pages in a dual language setup.

carlosjeurissen (changes requested):
Thanks so much for putting this together @hanguokai!

> Looks good to me, although I would like this proposal to also cover a getMessage overload (or separate method) to get a message in a specific language.

This has been proposed here:
https://github.com/w3c/webextensions/issues/274

We can include some variation of one of the proposals in this PR. Another potential syntax would be accepting an object for setCurrentLanguage which accepts a tabId. Something like:
```js
await browser.i18n.setCurrentLanguage({
  code: 'en-US',
  tabId: 82938
});
```

Which would restrict the change of the language to a specific tab. After which all calls to i18n.getMessage would follow the language specified in this call.

However my personal favourite would still be something like this:
```js
browser.i18n.withLanguage('pt-BR').then((getMessage) => {
  let someMessage = getMessage('some_id');
});
```

proposals/per-extension-language-preferences.md
hanguokai: Because some extensions are not ready to adapt to it. For example, some extensions seem to support many languages, but these are just for adding listing descriptions in other languages in the extension store (Extensions Store only allows adding descriptions in languages that the extension supports),  they don't actually support these languages in the extension.

proposals/per-extension-language-preferences.md
hanguokai: No, they still need to add messages.json files for languages they support. Because the extension name, short name and description can only be localized with `browser.i18n` mechanism.

proposals/per-extension-language-preferences.md
hanguokai: In the "Definitions" section, I already defined it. So don't need add it everywhere else.

proposals/per-extension-language-preferences.md
hanguokai: No.

proposals/per-extension-language-preferences.md
hanguokai: I think it is unnecessary, it complicates the design.

hanguokai:
> Another potential syntax would be accepting an object for setCurrentLanguage which accepts a tabId.

There is no "per-tab" demand. It complicates the design.

proposals/per-extension-language-preferences.md
hanguokai: Thanks. Grammatical issues aren't the top priority, though.

proposals/per-extension-language-preferences.md
carlosjeurissen: @hanguokai Considering the different formatting used, as mentioned here, https://github.com/w3c/webextensions/issues/642, I believe it is good to be explicit here. Your call though.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Get the extension displayed language.
+ * return the language tag, e.g. en-US.
+ */
+i18n.getCurrentLanguage() : string
+
+/**
+ * Set the user preferred language.
+ * if code is null, revert to the unset state.
+ * if code is not valid, return a rejected Promise.
```
carlosjeurissen, line 61:
```suggestion
 * if code is not valid, for example when it uses an underscore, return a rejected Promise.
```

Documenting https://github.com/w3c/webextensions/pull/641#discussion_r1647699428

carlosjeurissen (changes requested):

proposals/per-extension-language-preferences.md
aphillips: It's important to know the difference between a _locale_ and _localization_ here. There can be many locales supported by a single localization. This API seems to want to return the list of available localizations, but users typically need to choose between the list of available locales. The locale can be used do to internationalized formatting/processing (such as in MessageFormat or using `Intl`), which provides a better-adapted, richer localized experience than merely translating the static messages.

For example, many applications come in just two varieties of English (US vs. UK/International English), but support many locales. The language negotiation/resource fallback (such as used by `getMessage`) takes care of filling in the localized strings for any requested locale (including using the default language when the requested locale has no available localization), but you still want the locale (in most cases) to provide processing/formatting. Note that many applications also tailor the fallback, so that, for example, the `es-419` (Spanish, Latin America) localization serves many seemingly-unrelated (from a tag perspective) Spanish locales (`es-CO`, `es-CR`, `es-AR`, etc.....)

proposals/per-extension-language-preferences.md
carlosjeurissen: @hanguokai @oliverdunk How about instead of specifying `builtin_languages_menu`, we add a `available_locales` property in the manifest listing all the locales supported by the extension. This would prevent the need for specifying the messages.json as sometimes (from my experience with some extensions) a language is supported to some extend without having messages present in messages.json. To overcome this one would need to copy the english messages.json but this seems very counter-intuitive. Specifying an empty messages.json is not sufficient to at least tell the Chrome Web Store a specific language is supported.

carlosjeurissen:
> > Another potential syntax would be accepting an object for setCurrentLanguage which accepts a tabId.
> 
> There is no "per-tab" demand. It complicates the design.

Seems @xPaw has potentially such use case. From my own experience in a content-script I sometimes want to match the language of the page instead of the extension when injecting elements to the DOM to prevent mixed-language situations.

proposals/per-extension-language-preferences.md
carlosjeurissen: @hanguokai The motivation is not improving the grammar but to improve the meaning of the sentences. In this case to prevent confusion found in https://github.com/w3c/webextensions/pull/641#discussion_r1644443944. Generally I very much agree with leaving the grammar and formatting to later in the process. For example, with the icon_variants proposal it was counter productive to early-on add line-breaks for formatting motivations as with every change this had to be updated.

proposals/per-extension-language-preferences.md
hanguokai: > There can be many locales supported by a single localization. This API seems to want to return the list of available localizations

Yes, this proposal and the current `browser.i18n.getMessage()` mechanism ( [MDN doc](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization) , [Chrome doc](https://developer.chrome.com/docs/extensions/reference/api/i18n) ) mainly focus on localization (language translations). In other words, it focuses on languages, not all possible regions.

Regarding expanding locales, I think this may be an area that needs to be improved in the future, or expand by developers themselves.

aphillips: > Regarding expanding locales, I think this may be an area that needs to be improved in the future, or expand by developers themselves.

I agree about alternate fallback paths (although `Intl` has a proposal for this and it is already the way that CLDR data--and thus `Intl`--works for some locale-based "best match" selection vs. BCP47 Lookup), but think that this document isn't putting enough thought into the separation of locales and localizations/translations. Trying to use the same mechanism for both will disappoint users. Trivial examples with only a couple of languages don't match the needs of those developers who need to support a fairly large set of locales.

`getMessage` tries to use a BCP47 Lookup type of fallback to select the correct message "like a resource bundle" (cf. Java or GNU gettext for examples of other resource bundle systems). Resource selection often had a dual fallback capability (the _messages_ files and the specific keys within each message file).

The resource files want to use the **_least_** specific locale as their identifier possible (the better to provide coverage for many locales). The user, however, wants to specify the **_most_** specific locale possible (including script, region, and various locale extensions where applicable), the better to tailor the runtime experience.

In this document, there are calls to things such as "selecting which _locale_" to use (i.e. a picker) as well as "selecting which _localization_" to use (i.e. which language file or files to download. These lists want to be different (the list of locales is long, while the list of localizations is typically shorter).

---

Note that there is on-going work at W3C related to manifest (they are adopting slightly different mechanisms for managing localization). Also, I am the editor of [Developing Localizable Manifests](https://www.w3.org/TR/localizable-manifests/), which tries to enumerate a lot of this material and which readers on this thread might find useful. Lastly, there is work at Unicode on [MessageFormat 2](https://github.com/unicode-org/message-format-wg/) and proposed for MessageResource (which is similar to the message files in `getMessage`)

xPaw:
> Seems @xPaw has potentially such use case.

I'd be just fine with an overload that accepts a language. Is it even possible to get current tab id without having the tabs permission?

carlosjeurissen:
@xPaw Can you describe your use-case in detail? Getting the current tab id can always be achieved without the tabs permission using chrome.tabs.getCurrent(). This would not work in an extension popup as it is technically not a tab.

xPaw:
> Can you describe your use-case in detail? 

To be specific, [I have this extension](https://github.com/SteamDatabase/BrowserExtension), which lists a bunch of content scripts for various pages on Steam Store/Community, it adds some extra buttons/blocks to the page like 'lowest recorded price' for example. Since I added localization support, people are getting mismatching languages on the page (like the page may be in English, but their browser is in Japanese, so they get messages added by the extension in the browser language). Users are asking to change the extension's language because of this.

If I was able to pass a specific language to getMessage, I would be able to match the messages to the current page language (I am able to figure out the page's current language).

The content scripts simply use `i18n.getMessage` (I have a helper function `_t`), other extensions solve this by loading the locale json to get messages manually.

proposals/per-extension-language-preferences.md
carlosjeurissen: @oliverdunk If no browser vendor wants to implement something like this, we can offer these APIs in `browser.management` so an extension can create this UI to customise the languages of other extensions.

carlosjeurissen:
@hanguokai and others. Let me know if you feel https://github.com/w3c/webextensions/issues/274 should fit in this proposal. As the use case of @xPaw and others are not covered by all the methods in this proposal. If so I will write up suggested additions. Otherwise I could write a formal proposal as promised during the San Diego meetings.

hanguokai:
> Let me know if you feel #274 should fit in this proposal.

I think #274 should be a separate proposal. These use cases are different from setting the user's preferred language for this extension, they are more **dynamic** scenarios. However, if users want to use the same language in various places, they should use this proposal's APIs.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,234 @@
...
+
+This is a generic i18n feature, which means that all types of extensions can use it. Some extensions already provide this functionality via workarounds or non-browser.i18n implementations.
+
+## Specification
+
+### Definitions
+
+In this document:
+- **locale** or **language code** or **language tag** is a string that represent a language, defined in [BCP 47](https://www.rfc-editor.org/info/bcp47).
+For example, `en-US`, `zh-CN`, `fr`. It is used by `browser.i18n`, `Date`, `Intl` and various other APIs.
```
xfq, line 45: `zh-CN` is presumably meant to indicate Simplified Chinese, which is also used in Singapore. It might be better to use `zh-Hans` as the language tag in the example.

See also https://www.w3.org/International/articles/language-tags/index.en.html#script

hanguokai: Perhaps due to historical reasons, browsers still uses 'zh-CN' instead of 'zh-Hans'. For example, `navigator.language` returns "zh-CN" in Chrome and Firefox on macOS. And Chrome only supports [a list of languages](https://developer.chrome.com/docs/extensions/reference/api/i18n#locales) in the `/_locales/` directory, which only contains a limited combination of languages and regions.

I think solving this problem seems to go beyond the core issue that this proposal aims to address. This proposal is intended to expand functionality on the basis of existing capabilities.

carlosjeurissen: Still would be good and is also relevant for this issue what locales a browser supports / can deal with. See also: https://github.com/w3c/webextensions/issues/131

For example, the Naver Whale store rejects any language tag which includes the script subtag (Hans, Latn, Cyrl, and others).

proposals/per-extension-language-preferences.md
oliverdunk: @aphillips, thanks for taking your time to share your expertize on this issue. I am certainly out of my depth so the extra context and pointers on what to consider are really appreciated. If you'd like to attend a public meeting at some point, we could likely carve out some time to make sure we can have an in depth discussion. Otherwise your time continuing to help with the conversation here is valued regardless.

Based on all of the above, I am still of the opinion that offering a `getAllLanguages` that is intended for rendering a menu is going to be very hard to get right. Given users want to choose a fairly specific locale, it would likely need to return a much longer list of anything you can `setCurrentLanguage` to, as @carlosjeurissen suggested. However, that seems like it would be a long list which would be hard to render with the right nesting / structure without additional work. You would also need a mapping from returned codes to strings which are human readable.

I can see the value in a function that returns "which of the folders in _locales were accepted". Perhaps we could lean into that with an API like `getParsedLocalesDirectories()`? Based on @aphillips' explanation it seems like locales may be the wrong word there but I think we are slightly backed into a corner by the existing usage of `_locales` as the canonical place for extensions to store messages.json files.

For `setCurrentLanguage`, I'm convinced after this discussion that you should be able to set it to any valid locale, however specific, and the browser should pick the appropriate language file with fallbacks if needed. That way you can offer a fairly detailed picker for users and rely on the browser to use its built-in fallback behavior.

proposals/per-extension-language-preferences.md
oliverdunk: As mentioned above, following the discussion with @aphillips I'm more in favour of allowing this now. Perhaps we should still throw in extreme cases (if you try to set the language to French, but you only have English message files) but beyond that allowing you to use browser fallbacks seems helpful.

proposals/per-extension-language-preferences.md
oliverdunk: I think I'd still be in favour of removing this. It doesn't change anything about the behaviour and having the spec guide implementors seems like a desirable thing.

proposals/per-extension-language-preferences.md
oliverdunk: What would a `browser.management` API look like? If it involves rendering new UI, I don't think Chrome would be supportive. I would expect other browsers to feel similarly if they don't plan on implementing a menu.

hanguokai: I moved the "Browser built-in language selection menu" to the "Future Work" section in this document. When browser vendors are interested in this, we can discuss it again.

proposals/per-extension-language-preferences.md
hanguokai: I rephrased text about updating browser UI and extension UI in a new commit.

proposals/per-extension-language-preferences.md
hanguokai: I removed this line.

proposals/per-extension-language-preferences.md
hanguokai: > rendering a menu is going to be very hard to get right …… 

Creating a language selection menu has always been a challenge in i18n work, but it is also a necessary work. It can be very simple or very complex. For example, the extension can only list a few languages without variations (like regions and scripts), or list languages with limited or all regions, or let users set languages, regions and extensions (like calendar, date format, etc) separately.

> You would also need a mapping from returned codes to strings which are human readable.

`Intl.DisplayNames` can help it. It also depends on how developers want to design it. For example, a language code can be displayed in two display names (e.g. "zh-CN" can be displayed by "Chinese Simplified (简体中文)"). Anyway, display names are not what this proposal is trying to solve.

hanguokai: > For setCurrentLanguage, I'm convinced after this discussion that you should be able to set it to any valid locale, however specific, and the browser should pick the appropriate language file with fallbacks if needed. That way you can offer a fairly detailed picker for users and rely on the browser to use its built-in fallback behavior.

@oliverdunk Do you know why Chrome only allow developers to use [a fixed list of languages](https://developer.chrome.com/docs/extensions/reference/api/i18n#locales)? It is limited by the browser or Chrome Web Store? If it is a hard limitation, that means the browser doesn't support some locales to use in `setCurrentLanguage()` and the `/_locales/` directory.

proposals/per-extension-language-preferences.md
carlosjeurissen: @oliverdunk the whole point of adding APIs to `browser.management` would be to prevent the browser having to render UI but let extensions figure this out. imagine something like:
```js
browser.management.setLanguage(extensionId, locale)
```

To let an extension change the language of another extension.

oliverdunk: Ah, I missed this. I could go either way on if that should be in scope here but that seems reasonable.

proposals/per-extension-language-preferences.md
oliverdunk: @hanguokai My understanding is that it is based on the locales supported by some of the internal libraries we use to handle i18n.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * After changing to a new language, the browser triggers a language changed event.
+ * The callback is function(code: string) : void.
+ * The code parameter in the callback is the new language.
+ */
+i18n.onLanguageChanged.addListener(callback)
+```
+
+### Behavior
```
Rob--W, line 80: The `i18n` API is one of the few APIs that are available to all extension contexts, including lower-privileged content scripts.

Could you specify the intended availability of the API?

I'd say, the default is to NOT expose the APIs to content scripts unless a good reason is given otherwise.

The `getCurrentLanguage` method sounds potentially useful to content scripts, but I'm not sure what the value of this would be. It is specified to be synchronous, which means that browser implementations are forced to make the language choice available to all content processes. It is likely for this info to already be available along with the messages though.

`setCurrentLanguage` - To prevent (compromised) content processes from affecting extensions, I would restrict its availability to extension contexts only (e.g. background script, popup, extension tabs, ...).

`getAllLanguages` - given that its primary intended use is to determine the language to change to, I'd also restrict this to non-content scripts too.

`i18n.onLanguageChanged` - sounds potentially useful to content scripts. But even without this, extensions could use extension messaging to notify content scripts upon detecting a language change.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Exposed Sensitive Data
+
+N/A
+
+### Abuse Mitigations
+
+N/A
+
+### Additional Security Considerations
```
Rob--W, line 201: This proposal has `N/A` at "Abuse Mitigations" and "Additional Security Considerations".

I see potential for abuse: An extension could rapidly change their name at runtime to make it more difficult for users to discover what extension they are interacting with. The proposal does not describe any mitigations for this abuse concern.

Rob--W (changes requested):
At the surface this looks like a reasonable API proposal. I am concerned about the ability for extensions to rapidly change the locale and its impact on UI surfaces in the browser and would like to see some mitigations for this abuse by design.

proposals/per-extension-language-preferences.md
carlosjeurissen: @Rob--W In general I believe it would be best to fetch the translations for the manifest file in the language of the containing UI. So if it will be used in the browser extension management page, it would match the language of said page. Same for the context menu items on the action button. This would prevent the issue you are describing.

proposals/per-extension-language-preferences.md
hanguokai: For content scripts, there are several possibilities:

1. Loosest: all methods are available in content scripts.
1. Strictest: all methods are not available.
1. Some are not available.

I'm leaning towards option 3. Specifically, `setCurrentLanguage`, `getCurrentLanguage` and `getAllLanguages` are not available in content scripts (at least `setCurrentLanguage` is not available). And `onLanguageChanged` is available.

proposals/per-extension-language-preferences.md
hanguokai: Although I don't think this poses a real security problem. However, the browser can limit the frequency of `setCurrentLanguage` method calls, which is up to the browser.

proposals/per-extension-language-preferences.md
carlosjeurissen: @hanguokai I would suggest `getCurrentLanguage` to be available in content scripts considering the extension might want to use the current language code outside getMessage situations. In addition, if we add some predefined i18n message, something like @@current_locale, this would be available using getMessage in contentScripts anyway.

oliverdunk: If `onLanguageChanged` is available, I definitely agree with @carlosjeurissen that the corresponding getter should be available too.

hanguokai: I add a new section of Content Scripts Availability

proposals/per-extension-language-preferences.md
hanguokai: I added  a supplementary note in the Abuse Mitigations section. let me know if it's Ok.

proposals/per-extension-language-preferences.md
oliverdunk: Hey Jackie - I spoke to the team about this. We're supportive of the idea, but it is very unlikely we would be able to implement this ourselves any time soon. With that in mind, we'd only be comfortable sponsoring this if there was an external contributor able to implement it.

@xeenon / @Rob--W With that in mind, would either Apple or Mozilla be more likely to implement this soon and be interested in sponsoring?

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+N/A
+
+### Abuse Mitigations
+
+A case of abuse is changing the extension name/description dynamically by `i18n.setCurrentLanguage()`. This is not a security threat because all metadata published in the extension store is audited. And for the extension management page (e.g. chrome://extensions/), the extension name should be displayed in the same language as the management page (i.e. the Browser UI language), so it is not affected by `i18n.setCurrentLanguage()`.
```
Rob--W: > "This is not a security threat because all metadata published in the extension store is audited."

Please drop this part. This claim about it not being a security threat is not true. A legitimate extension can have different names for branding purposes, and that would be accepted by the web store. This is indistinguishable from a malicious extension that intentionally chooses different names to maximize confusion, when the extension switches names dynamically.

Here is a rephrased suggestion that calls out the exact risk, and also language on the mitigation. I also added another example of a browser UI surface where this matters (permissions dialog/prompt).

```suggestion
A case of abuse is changing the extension name/description dynamically by `i18n.setCurrentLanguage()` in an attempt to confuse the user and hinder their ability to identify the extension. A mitigation to this issue is to ignore changes from `setCurrentLanguage` in the extension management parts of the browser UI, and always use the browser's configured UI language instead. This includes extension management pages (e.g. `chrome://extensions` in Chrome and `about:addons` in Firefox) and permission dialogs.
```

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,251 @@
...
+ * The code parameter in the callback is the new language.
+ */
+i18n.onLanguageChanged.addListener(callback)
+```
+
+### Behavior
+
+#### Content Scripts Availability
+
+In history, all `browser.i18n` APIs were available in content scripts. But some of methods in this proposal should not be available in content scripts because of their purpose.
```
Rob--W:
```suggestion
Unlike other APIs, all `browser.i18n` APIs are historically available to content scripts. To counter abuse, some of the features in this proposal should not be available in content scripts.
```

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,251 @@
...
+
+| Method  | Content Scripts Availability |
+| ------------- | ------------- |
+| `getCurrentLanguage()` | Yes |
+| `setCurrentLanguage()` | No |
+| `getAllLanguages()` | No |
+| `onLanguageChanged()` | Yes |
```
Rob--W:
```suggestion
| `onLanguageChanged` | Yes |
```

This is an event, not a method.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,251 @@
...
+i18n.onLanguageChanged.addListener(callback)
+```
+
+### Behavior
+
+#### Content Scripts Availability
+
+In history, all `browser.i18n` APIs were available in content scripts. But some of methods in this proposal should not be available in content scripts because of their purpose.
+
+| Method  | Content Scripts Availability |
```
Rob--W:
```suggestion
| API  | Content Scripts Availability |
```

Method -> API because onLanguageChanged is not a method.

proposals/per-extension-language-preferences.md
```diff
@@ -0,0 +1,251 @@
...
+| ------------- | ------------- |
+| `getCurrentLanguage()` | Yes |
+| `setCurrentLanguage()` | No |
+| `getAllLanguages()` | No |
+| `onLanguageChanged()` | Yes |
+
+#### A New Predefined Message `@@current_locale`
+There is an exsiting predefined message `@@ui_locale`, that reflects the value of `i18n.getUILanguage()`, but the value uses underscore (e.g. "en_US") as separator in Chrome and hyphen (e.g. "en-US") in Firefox.
+
+Relative to `@@ui_locale`, a new predefined message, ``@@current_locale``, should be added, which reflects the value of `i18n.getCurrentLanguage()`.
```
Rob--W, line 96: This is a late addition to the proposal. What use cases are served with the proposed `@current_locale`? This property does not automatically follow from the three listed use cases in the proposal document.

Additionally, because the stylesheet does not update automatically in response to UI language changes, the use of the proposed `@current_locale` in the stylesheet can result in UI getting out of sync with the actual language, unless the extension has additional code to deal with language changes (with `i18n.onLanguageChanged`). If an extension already needs to account for this, then they may as well do all language-dependent CSS from JavaScript. These days, they can use CSS variables to control stylesheet behavior in a more powerful way than `@`-macros can. For example, the only way for an extension to reload the stylesheet to read the new `@` values is to replace the style sheet element with a new one that has the same URL and a random extra addition (query string) to bust the style cache.

Because of this, I am inclined to vote for removing the proposed `@current_locale` from the proposal.

hanguokai: Personally, I don't pursue this predefined message very much, but only add it for completeness reason. @carlosjeurissen Would you mind deleting this part? Because you suggested it.

carlosjeurissen: Also not feeling too strongly about it. One of the motivations is described here:
https://github.com/w3c/webextensions/issues/642#issuecomment-2181397837

As a way to get the path of the language files in JavaScript for example.

In general I agree with not using the predefined messages in CSS. Personally not been using them in any shape or form in the past few years in any projects in CSS. It very much feels like a hack and we might want to deprecate some of them moving forward.

proposals/per-extension-language-preferences.md
carlosjeurissen: As far as I know both the chrome web store and chrome browser support more languages than specified in that fixed list of languages. The Chrome Web Store also supports and recognises more languages currently than this list contains.

@oliverdunk are there any current internals which would restrict the languages one could use in `setCurrentLanguage()` to this fixed list?

For `i18n.getAllLanguages()` specifically I have opened a separate PR to specify and discuss this method as it is useful outside the `i18n.setCurrentLanguage()` use cases and could be implemented separately. See https://github.com/w3c/webextensions/pull/942

dara012752991-web (approved):

dara012752991-web (commented):
proposals/per-extension-language-preferences.md

carlosjeurissen:
@dara012752991-web thanks for reviewing the proposal! Do you have any remarks or suggestions?

Issue #601, opened 2024-04-25
Proposal: `StorageArea.getKeys()`


polywock:
Taken from #505, as an alternative proposal suggested by @tophf and @Rob--W. It's significantly different from my original proposal, so I'm creating a separate proposal for it. 

## Background

I regularly need to use storage keys that involve dynamic elements like flag:[tabId]. The only way to get all keys starting with the prefix `flag:` is using StorageArea.get() to get all the storage items and then to filter.

This isn't efficient. The storage areas support 10MB of data and even more with the unlimitedStorage permission. That's potentially a huge amount of storage data that needs to be serialized and sent over. There's also a special concern when using the Sync storage area.

## Proposal 

A method to get all keys without the corresponding values. 

```ts
browser.storage.local.getAllKeys(): Promise<string[]> 
```

## Resources 

https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex/getAllKeys

oliverdunk:
@polywock I plan to implement this in Chromium as part of some work I am doing on the storage API. I’ve started a proposal PR using the template - would you like to work together on finishing that, or are you happy for me to open it? Thanks!

polywock:
@oliverdunk I appreciate the offer to be involved. But, I'll probably slow down the process. 

Please go ahead without me. Thank you!

oliverdunk:
I have opened a proposal for this in https://github.com/w3c/webextensions/pull/649.

polywock:
Pull request has been approved and merged.  🚀

Rob--W:
Re-opening this ticket because we usually keep tickets open until there is nothing left to be done. In this case, we can use this ticket to signal the cross-browser implementation status of the feature (as issue labels here).

I have opened https://bugzilla.mozilla.org/show_bug.cgi?id=1910669 to track the implementation in Firefox.

oliverdunk:
I've just landed the CL for this in Chromium ([link](https://chromiumdash.appspot.com/commit/753b4707b454b3b99f40509aa8a7608feb8ed5c0)). It will be available in Canary later today and more generally Chrome 130 as that rolls out across channels ([schedule](https://chromiumdash.appspot.com/schedule)).

fregante:
Safari landed this in the their technology preview release: 🎉


- https://developer.apple.com/documentation/safari-technology-preview-release-notes/stp-release-206
[...]

PR #649, opened 2024-06-28
Add getKeys() proposal

oliverdunk:
Proposal for the getKeys() method from #601, originally proposed by @polywock.

I am planning to implement this in Chromium as part of some work on the `browser.storage` API.

### DIFF
--- a/proposals/storage-get-keys.md
+++ b/proposals/storage-get-keys.md
@@ -1,0 +1,115 @@
+# Proposal: browser.storage.&lt;area>.getKeys
+
+**Summary**
+
+API to retrieve all keys for a given storage area in the `browser.storage` API.
+
+**Document Metadata**
+
+**Author:** [Oliver Dunk (Google)](https://github.com/oliverdunk)
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** [polywock](https://github.com/polywock)
+
+**Created:** 2024-06-28
+
+**Related Issues:** [#601](https://github.com/w3c/webextensions/issues/601)
+
+## Motivation
+
+### Objective
+
+This API enables developers to retrieve the keys in storage without needing to
+receive all of the data stored alongside those keys. Depending on the
+implementation, the impact of this optimization may differ - for example
+Chromium has only a limited ability to access data from storage without
+fetching the associated data. Regardless, this reduces the amount of data
+returned to the caller and allows for further optimizations in the future.
+
+#### Use Cases
+
+There are existing use cases for enumerating the data in extension storage. For
+example, extensions for Chrome that add extension storage areas to DevTools.
+While these also need a way to show the associated values, they may not want to
+load these upfront.
+
+Additionally, in some cases developers may want to access keys with a given
+prefix, for example if using a key like `foo:tabid`. This API proposal does not
+allow matching by prefix (see "Future Work") but developers would be able to do
+this themselves after obtaining the list of keys.
+
+### Known Consumers
+
+There is developer interest for this API in
+[#601](https://github.com/w3c/webextensions/issues/601). While we don't expect
+significant immediate usage, it is a small, well-scoped API and should not
+represent significant engineering effort.
+
+## Specification
+
+### Schema
+
+```ts
+interface StorageArea {
+    getKeys(): Promise<string[]>;
+}
+```
+
+The original issue proposed the `getAllKeys()` name, to match IndexedDB. This
+proposal suggests the slightly shorter `getKeys()`, since that seems equally
+clear and less verbose.
+
+### Behavior
+
+This API will throw / populate browser.runtime.lastError in the case of
+failure, for example if access to the storage area is not allowed in the
+given context.
+
+### New Permissions
+
+This API will use the existing `storage` permission.
+
+### Manifest File Changes
+
+There are no changes to the manifest.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+This API does not expose any data which is not already accessible to the
+extension.
+
+### Abuse Mitigations
+
+This does not expose any new data so there are no new abuse vectors.
+
+### Additional Security Considerations
+
+N/A
+
+## Alternatives
+
+### Existing Workarounds
+
+Developers can retrieve all data using the `get()` API with `null` as the keys
+parameter, which will return everything in storage. This is very similar but
+requires all of the data stored for a given key to be returned, which has
+performance implications.
+
+### Open Web API
+
+In the future, using web storage APIs in WebExtensions would certainly be
+desirable. However, the storage API already has significant adoption.
+Therefore, there is value in making it more ergonomic for developers.
+
+## Implementation Notes
+
+N/A
+
+## Future Work
+
+It may be desirable to add a parameter to allow filtering of the keys that are
+returned. This is not an optimization we think is neccessary in an initial
+implementation.
--- a/proposals/storage-get-keys.md
+++ b/proposals/storage-get-keys.md
@@ -111,5 +111,5 @@ N/A
 ## Future Work
 
 It may be desirable to add a parameter to allow filtering of the keys that are
-returned. This is not an optimization we think is neccessary in an initial
+returned. This is not an optimization we think is necessary in an initial
 implementation.
--- a/proposals/storage-get-keys.md
+++ b/proposals/storage-get-keys.md
@@ -62,7 +62,7 @@ clear and less verbose.
 
 ### Behavior
 
-This API will throw / populate browser.runtime.lastError in the case of
+This API will reject with an Error / populate browser.runtime.lastError in the case of
 failure, for example if access to the storage area is not allowed in the
 given context.
 
--- a/proposals/storage-get-keys.md
+++ b/proposals/storage-get-keys.md
@@ -49,7 +49,7 @@ represent significant engineering effort.
 ## Specification
 
 ### Schema
-
+This proposal expands the existing [StorageArea type](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea) with a new `getKeys` method:
 ```ts
 interface StorageArea {
     getKeys(): Promise<string[]>;


proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+
+```ts
+interface StorageArea {
+    getKeys(): Promise<string[]>;
+}
+```
+
+The original issue proposed the `getAllKeys()` name, to match IndexedDB. This
+proposal suggests the slightly shorter `getKeys()`, since that seems equally
+clear and less verbose.
```
hanguokai, line 61: I'm not against this name. Just think of many other APIs that use the name `keys()`, such as: `Object.keys()`, `Map.keys()`, `Set.keys()` `Array.keys()`, `URLSearchParams.keys()`, `CacheStorage.keys()`, `Cache.keys()`, `FormData.keys()`, `Headers.keys()`, `NodeList.keys()`.

xeenon: In your examples, `CacheStorage.keys()` and `Cache.keys()` return promises, the others are synchronous results (many being iterators).

The extensions API has historically used `get` prefixes due to the async nature of the APIs with `callback` functions. However, with our APIs now using promises, I think it does make sense to drop `get` in our APIs going forward.

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Behavior
+
+This API will throw / populate browser.runtime.lastError in the case of
```
xeenon: "This API will reject in the case of…" since there is no callback version, right?

(However, in Safari / WebKit, everything that returns a promise can take a final `callback` parameter instead, even in v3 for v2 compatibility. Then `lastError` is relevant still. Maybe we should change that…)

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+failure, for example if access to the storage area is not allowed in the
```
xeenon, line 66: Wouldn't `browser.storage` be `undefined` if "access to the storage area is not allowed"? (This might not be the best failure example to give.)

proposals/storage-get-keys.md
oliverdunk: I think there's also an argument that we should continue with the existing naming convention until we can update the API surface as a whole. I don't feel too strongly though.

proposals/storage-get-keys.md
oliverdunk: That's a good question - I intentionally didn't focus too much on callbacks for the proposal, since as you mentioned they are mostly for backwards compatibility. Perhaps a brief mention is still worthwhile though - we should definitely support them.

proposals/storage-get-keys.md
oliverdunk: In Chrome, it is possible to limit access to the `storage.session` area using the `storage.session.setAccessLevel` API. The API is still exposed but throws if it is called.

oliverdunk (commented):
Thanks for doing a pass @xeenon, added some replies to your comments.

proposals/storage-get-keys.md
Rob--W: I'm okay with `getKeys` and `getAllKeys`. I'd rather not use `keys()`.

proposals/storage-get-keys.md
Rob--W: Chrome's API doesn't throw, but reject in this case.

proposals/storage-get-keys.md
Rob--W:
```suggestion
This API will reject with an Error / populate browser.runtime.lastError in the case of
```
I don't expect any error to be thrown, unless invalid parameters are passed. The normal mode of error handling here is by rejecting the Promise.

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+Therefore, there is value in making it more ergonomic for developers.
+
+## Implementation Notes
+
+N/A
+
+## Future Work
+
+It may be desirable to add a parameter to allow filtering of the keys that are
+returned. This is not an optimization we think is neccessary in an initial
```
Rob--W:
```suggestion
returned. This is not an optimization we think is necessary in an initial
```

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+
+There is developer interest for this API in
+[#601](https://github.com/w3c/webextensions/issues/601). While we don't expect
+significant immediate usage, it is a small, well-scoped API and should not
+represent significant engineering effort.
+
+## Specification
+
+### Schema
+
```
Rob--W:
```suggestion
This proposal expands the existing [StorageArea type](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea) with a new `getKeys` method:
```

For clarity, added a line explaining what "StorageArea" means. I linked to MDN because that's the only documentation that lists "StorageArea" as a generic type description.

Rob--W (approved):

xeenon (approved):

proposals/storage-get-keys.md
oliverdunk: Good spot, thanks. That's what I was trying to say.

proposals/storage-get-keys.md
oliverdunk: Adjusted this wording above.

proposals/storage-get-keys.md
oliverdunk: I removed the locale from your URL, but otherwise this looks good. Thanks!

proposals/storage-get-keys.md
oliverdunk: Given there is a Chrome and Firefox preference for `getKeys()` or `getAllKeys()`, it seems like we should pick between those. @hanguokai / @polywock, do you have preferences between those? I would prefer `getKeys()` if we can all align, just to keep it short.

polywock: I also prefer getKeys() as it's more flexible. The proposal's Future Work section mentions a potential filter parameter. getAllKeys() might not be compatible with such parameter. 

getKeys(filter)  // makes sense
getAllKeys(filter) // makes less sense

hanguokai: The name `x.keys()` is widely used in the JavaScript ecosystem, whether synchronous or asynchronous methods. It is used much more than the name `x.getKeys()` or `x.getAllKeys()`.

Also, in the `browser.storage` namespace, it uses only the names `get()` and `set()`, rather than `getValues()` or `getItems()`. So `get` is not used as a prefix here. In other words, the get prefix is not an existing naming convention in this namespace.

So I still suggest using the name `keys()`. But it doesn't really matter what it's called.

polywock: > get() and set(), rather than getValues() or getItems() ... So get is not used as a prefix here. In other words, the get prefix is not an existing naming convention in this namespace.


Many namespaces in the Web Extension API have a primary `get()` and `set()`, like `browser.tabs.get()` or `browser.cookies.set()`. Additional methods that retrieve values have a get prefix like `tabs.getCurrent()`, `tabs.getSelected`, `storage.getBytesInUse()`, etc. 

I've done a cursory glance at all the namespaces and couldn't find any example of a method that retrieves data that doesn't start with `get`. 

Note: I'm fine with either `keys()` or `getKeys()`. Leaning towards `getKeys()`, but `keys()` will be easier to autocomplete as no other methods start with K.

hanguokai: Here, it depends on whether you think of "keys" as a verb or a noun. As a verb, `keys()` makes sense, like "open" in `sidePanel.open()` and "capture" in `tabCapture.capture()`. As a noun or data, `getKeys()` makes more sense.

Especially in key-value data structures or databases, "keys" are more often used. And `browser.storage` falls into this category.

xeenon: I agree with @polywock that `getKeys()` makes more sense if `filter` is ever added in the future.

hanguokai: Both names ("getKeys" or "keys") are fine, simple and clear, I have no objection to either.

In fact, developers may primarily use it as an intermediate method, and wrap it as:
```js
async function getValues(filter) {
    let targetKeys = (await browser.storage.local.getKeys()).filter(filter);
    return browser.storage.local.get(targetKeys);
}
```

oliverdunk: Thanks Jackie! Going to resolve this since there seems to be alignment on getKeys(), which we discussed in the last public meeting.

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+Additionally, in some cases developers may want to access keys with a given
+prefix, for example if using a key like `foo:tabid`. This API proposal does not
+allow matching by prefix (see "Future Work") but developers would be able to do
+this themselves after obtaining the list of keys.
+
+### Known Consumers
+
+There is developer interest for this API in
+[#601](https://github.com/w3c/webextensions/issues/601). While we don't expect
+significant immediate usage, it is a small, well-scoped API and should not
```
travislove766Lovesbarber, line 46: wow

proposals/storage-get-keys.md
```diff
@@ -0,0 +1,115 @@
...
+
+There are existing use cases for enumerating the data in extension storage. For
+example, extensions for Chrome that add extension storage areas to DevTools.
+While these also need a way to show the associated values, they may not want to
+load these upfront.
+
+Additionally, in some cases developers may want to access keys with a given
+prefix, for example if using a key like `foo:tabid`. This API proposal does not
+allow matching by prefix (see "Future Work") but developers would be able to do
+this themselves after obtaining the list of keys.
```
J3ekir, line 40: Maybe it's gonna be too specific but would you consider adding a parameter for prefix(es)? Null or undefined or a string or an array of strings, just like browser.storage.&lt;area>.get and continue to return an array of strings (or an empty array).

I'm in this group (who wants to access keys with a given prefix, actually prefixes). I didn't know this feature was added. I was using browser.storage.local.getBytesInUse and to give that as parameter I was calling the whole storage and extracting only the keys I needed by calling Object.keys.filter. When the storage was taking up space (~25MB) I realized this process was sloving down the page and I googled for a more practical way. Then I found this feature in [here](https://stackoverflow.com/a/79179793/13630257).

I have a browser extension designed to be used on a few specific websites. The part that interests this feature is like this. The websites don't offer GIF avatars because of both server and client side load. But they allow uploads. The server takes up the first frame and resize it as JPEG. And the original files are accessible to everyone in specific URL format. The extension fetches the GIFS and resize them to provide smaller, non-burdeningclient-side. I keep those avatars as [siteId]_[userId] in the storage.

Even though they are small, they accumulate and take up space because I keep them in Base64. Because of this, I decided to put a size display and a reset option. Firefox doesn't currently offer browser.storage.local.getBytesInUse and requires manual (slower) calculation. Although I don't use very large storage except for avatars, it would be much more convenient to give the prefix(es) directly and get the keys that way.

oliverdunk: Thanks for sharing your use case! As mentioned in this proposal, we're definitely open to some way of filtering the keys. I think there's still an open question about what that would look like. In theory, prefixes would allow for optimization and be the simplest approach. In practice, however, I doubt any of the existing implementations could query by prefix - they would instead be retrieving all of the keys and filtering. At that point, there's some (but less) benefit to using a prefix and more advanced matching might make sense.

I think opening a new issue here would be reasonable if you're interested. The hardest thing will likely be finding a sponsoring browser - I suspect we won't have the capacity to prioritize this on the Chrome side in the short term. We would always accept patches once we've agreed on an API though.

Issue #670, opened 2024-08-14
Exposing developer tools contexts in runtime.getContexts() API


sohailrajdev97:
# Exposing developer tools contexts in [runtime.getContexts()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getContexts) API

## WECG proposal for `runtime.getContexts()`
<https://github.com/w3c/webextensions/blob/main/proposals/runtime_get_contexts.md>

## Background
Extensions can customize the browser developer tools. When an extension customises the developer tools, it can run code in the following contexts:
1. **Developer tools page**: This is a background page which is spawned as soon as developer tools is opened. Its lifetime is tied to the developer tools.
2. **Developer tools panel**: This is a panel which the extension can add to the developer tools by calling [`devtools.panels.create()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/panels/create) from the developer tools page (1).
3. **Developer tools sidebar pane**: This is a side pane which the extension can add to a particular panel (2) by calling [`devtools.panels.<panel>.createSidebarPane()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/panels/ElementsPanel/createSidebarPane) from the developer tools page (1).

<img src="https://github.com/user-attachments/assets/39a9e3b9-e0f5-4e38-9e63-034e9337c232" alt="drawing" width="500"/>

![image](https://github.com/user-attachments/assets/5c419102-5fdd-44ab-bd2a-a6500badb773)

![image](https://github.com/user-attachments/assets/a9395517-33df-458c-a091-c3829b3b306d)

## Proposal
All the above contexts host extension content (HTML and JavaScript). These should be available in the results of `runtime.getContexts() ` API. `runtime.getContexts()` asynchronously returns a list of [`ExtensionContext`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getContexts#return_value) objects. Following is a description of how `ExtensionContext`’s fields would be populated for developer tools contexts.

- **contextType**: Three new context types would be added (one for each context described above)
  1. `DEVTOOLS_PAGE` for developer tools page.
  2. `DEVTOOLS_PANEL` for developer tools panel.
  3. `DEVTOOLS_SIDEBAR_PANE` for developer tools sidebar pane.
- **tabId**: -1 as these contexts are not associated with a browser tab.
[...]

Issue #316, opened 2022-11-03
tabs.getCurrent() returns undefined when being called from the popup


kiaraarose:
I see that the MDN and Chrome documentation states that tabs.getCurrent() should return undefined when being called from an extension's popup. However, I think it's reasonable for a developer to want to know which tab the popup is currently open on. I'm curious to know what other developers think of this as I've seen some confusion about this online.

tophf:
It's the name which is misleading, a better name might be `getOwnTab`, but the behavior is correct: it should return undefined in a non-tab context such as the popup.

As for a new method, it would be essentially the same as `(await chrome.tabs.query({active: true, currentWindow: true}))[0]`, however there are some problems with naming:

* getActiveTab? but there's an active tab for each window
* getCurrentTab? but it's a different thing
* getPopupTab? but it can be used outside of the popup with the same result
* getActiveTabForCurrentWindow? it'll be almost as long as the `query` code

boontham33345:
Good

xeenon:
Safari currently returns the active tab in the active window when this is called from the background page. And it returns the active tab in the window showing the popup when called from the popup page.

carlosjeurissen:
Use case:
When wanting to use `executeScript` in mv3, you are required to specify a tabId. This is problematic for extension popups as one would have to rely on `tabs.query({currentWindow: true, active: true})`. Which could be different from the tab the extension popup was opened on.

Proposal:
Let `tabs.getCurrent()` return the tab the extension popup was invoked upon. This makes sure `executeScript` is applied to the correct tab.

Since Safari has weird behaviour currently with `tabs.getCurrent()` (active tab in current window). It might make sense to use a different name. like `tabs.getAssociated()`.

rdcronin:
I mentioned in #431, I mentioned I'd be supportive of changing this behavior to support returning the associated tab in an extension popup.  In Chrome, extension popups are strictly associated with a single tab (switching tabs in the window closes the popup), so there is no ambiguity about which tab is associated.  It's also inline (ish) with the motivation of the API, which is "get the tab this thing is running on".  While it's not returning the document the script is running in, I think the intention of the developer is clear from this context.

As @carlosjeurissen mentioned, it may make sense to explore using a different name, not least because "tabs.getCurrent()" is endlessly confused with getting the currently-active tab; we'll discuss that more.

Rob--W:
I am in favor of returning the tab that was clicked when the popup was opened.

[...]

PR #675, opened 2024-08-19
Add developer tools contexts in runtime.getContexts() proposal

sohailrajdev97:
Proposal for #670.

### DIFF
--- a/proposals/runtime_get_contexts.md
+++ b/proposals/runtime_get_contexts.md
@@ -242,8 +242,8 @@ for newly-introduced fields, we use the more intuitive `undefined` state.
 ### Developer Tools Contexts
 
 Extensions can use the
-[Dev Tools API](https://developer.chrome.com/docs/extensions/mv3/devtools/) to
-extend the browser's developer tools. When doing so, these extensions can run
+[Dev Tools API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools)
+to extend the browser's developer tools. When doing so, these extensions can run
 different types of contexts (page, panel, sidebar). `runtime.getContexts()`
 returns these contexts with the `DEVELOPER_TOOLS` context type.
 
--- a/proposals/runtime_get_contexts.md
+++ b/proposals/runtime_get_contexts.md
@@ -247,8 +247,8 @@ to extend the browser's developer tools. When doing so, these extensions can run
 different types of contexts (page, panel, sidebar). `runtime.getContexts()`
 returns these contexts with the `DEVELOPER_TOOLS` context type.
 
-- These contexts have a `tabId` of -1 as these contexts are not hosted inside
-a tab.
+- These contexts have `tabId` and `frameId` as -1 as these contexts are not
+hosted inside a tab.
 - If developer tools are docked to a tab, the `windowId` will be the ID of the
 window containing the tab.
 - If developer tools are undocked, the `windowId` will the ID of the developer
--- a/proposals/runtime_get_contexts.md
+++ b/proposals/runtime_get_contexts.md
@@ -105,6 +105,8 @@ extension.ContextType = {
   OFFSCREEN_DOCUMENT: 'OFFSCREEN_DOCUMENT',
   // A side panel or sidebar context.
   SIDE_PANEL: 'SIDE_PANEL',
+  // A developer tools context.
+  DEVELOPER_TOOLS: 'DEVELOPER_TOOLS',
 };
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/runtime_get_contexts.md
```diff
@@ -236,6 +238,22 @@ This is an artifact of existing APIs and precedence. Since many existing APIs
 use the constant integer values, we want to be consistent with those. However,
 for newly-introduced fields, we use the more intuitive `undefined` state.
 
+
+### Developer Tools Contexts
+
+Extensions can use the
+[Dev Tools API](https://developer.chrome.com/docs/extensions/mv3/devtools/) to
```
oliverdunk: Using https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/devtools might be slightly more browser neutral, but I don't have strong feelings.

proposals/runtime_get_contexts.md
```diff
@@ -236,6 +238,22 @@ This is an artifact of existing APIs and precedence. Since many existing APIs
...
+extend the browser's developer tools. When doing so, these extensions can run
+different types of contexts (page, panel, sidebar). `runtime.getContexts()`
+returns these contexts with the `DEVELOPER_TOOLS` context type.
+
+- These contexts have a `tabId` of -1 as these contexts are not hosted inside
```
oliverdunk: Is `frameId` set? I don't have strong feelings here.

oliverdunk (commented):
Thanks Sohail, this generally looks good to me! I'll chat to the team before adding an approval.

proposals/runtime_get_contexts.md
rdcronin: +1 to the question about frame ID.

Also, would it make sense to have tab ID be non-null if the devtools are docked to the tab?

(Similarly, no very strong feelings)

rdcronin (commented):
Thanks, Sohail!  This looks pretty good.  Same question as Oliver, but overall, no concerns.

proposals/runtime_get_contexts.md
```diff
@@ -236,6 +238,22 @@ This is an artifact of existing APIs and precedence. Since many existing APIs
...
+a tab.
+- If developer tools are docked to a tab, the `windowId` will be the ID of the
+window containing the tab.
+- If developer tools are undocked, the `windowId` will the ID of the developer
```
Rob--W, line 254: I thought that we didn't expose windowId of non-browser windows.

proposals/runtime_get_contexts.md
sohailrajdev97: Agreed. Changed the link.

proposals/runtime_get_contexts.md
sohailrajdev97: I too do not have very strong feelings here. I would want to know what other's think about this. Here are my thoughts:

1. From what I understand, `tabId` is set for contexts which are running inside a tab. Developer tools contexts do not run in a tab. They run in the developer tools window. 
2. Developer tools contexts can be background contexts too. Setting a `tabId` for such contexts sounds unnatural. 

Since `frameId` is used to identify a frame within a tab, we should set it to -1 too.

carlosjeurissen: Agree with @sohailrajdev97. In addition, getting the TabId is already possible using `devtools.inspectedWindow.tabId` https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/devtools/inspectedWindow/tabId from the devtools panel itself. Which can be requested using messaging between extension contexts.

See also: https://github.com/w3c/webextensions/issues/316

@dotproto having this info available from within getContexts could be nice additional information. Having this information for extension popups would be a nice addition. Currently there is no convincing way to even tho this from within extension popups and also option pages currently.

proposals/runtime_get_contexts.md
sohailrajdev97: Window ID is exposed for Dev Tools in Chromium and Firefox.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/WindowType

In that case, it should be good exposing it here too. What do you think? @oliverdunk, @xeenon, @Rob--W?

proposals/runtime_get_contexts.md
oliverdunk: Given the discussion yesterday, I would prefer to set both `tabId` and `frameId` to -1. That prevents confusion and avoids implying executing a script in those tabs / frames would execute it in the DevTools context. We can then talk about exposing the ID of the hosting tab as additional data in the future.

proposals/runtime_get_contexts.md
oliverdunk: I don't think there is any significant risk to this so I'm comfortable setting windows ID for docked DevTools windows. For undocked windows, I think it should be -1.

I'll check with @rdcronin to make sure we're all aligned on the Chrome side.

sohailrajdev97: > For undocked windows, I think it should be -1.

For my understanding, what is the risk if we expose the `windowId` for the undocked case too? It is already exposed via the windows API (i.e. `chrome.windows.getAll({windowTypes: ['devtools']})`).

oliverdunk: Good point. I just spoke to Devlin and based on this we both agreed we would prefer to set window IDs in all cases. That way we're consistent between various APIs.

xeenon (approved):

proposals/runtime_get_contexts.md
Rob--W: Firefox does not expose the devtools window at the moment.
Even Chrome hides the devtools window from most APIs, such as `windows.getCurrent()` and `windows.getLastFocused()`.

Nevertheless, I am not opposed to the windowId potentially be non-`-1`.

Rob--W (approved):

oliverdunk (approved):
This looks good to me and matches the behavior implemented in Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/5849634

Thanks for the work on this!

PR #678, opened 2024-08-22
Proposal: dom.execute()

rdcronin:
This adds a proposal for a new dom.execute() method, which will allow synchronous execution in another "world" from a content script.

### DIFF
--- a/proposals/dom_execute.md
+++ b/proposals/dom_execute.md
@@ -10,7 +10,7 @@ its injected in.
 
 **Author:** rdcronin
 
-**Sponsoring Browser:** Chromee
+**Sponsoring Browser:** Chrome
 
 **Contributors:** Rob--W, ...
 
@@ -38,9 +38,11 @@ Today, extensions can execute script in the main world by:
 
 The former is heavily asynchronous or requires knowledge of whether to inject
 beforehand (to register a script).  The latter is very visible to the site and
-has more potential side effects.  While there's no way to fully "hide" a script
-from a site, it is desirable to avoid adding DOM elements (which could
-interfere with sites with certain properties).
+has more potential side effects, and is also asynchronous since inline scripts
+are forbidden by the CSP applied in MV3 (to restrict remotely-hosted code
+injection).  While there's no way to fully "hide" a script from a site, it is
+desirable to avoid adding DOM elements (which could interfere with sites with
+certain properties).
 
 ### Known Consumers
 
@@ -96,15 +98,24 @@ key.
 Like arguments, the return value of the execution is serialized and
 deserialized using the Structured Cloning algorithm.
 
+##### Errors
+
+Thrown errors are also serialized using the Structured Cloning algorithm.  When
+an input parameter or return value cannot be serialized, an error is thrown
+synchronously.  Notably, `Promise` instances such as returned by `async
+function` cannot be serialized.
+
 ##### Injected script privileges
 
-The injected script has the same privileges has other script in the main world.
+The injected script has the same privileges as other scripts in the main world.
 This goes for API access, Content Security Policy, origin access, and more.
 
 ##### CSP Application
 
 The injected script is not considered inline and does not have a corresponding
-source; as such, CSP restrictions on script-src (and similar) do not apply.
+source; as such, CSP restrictions on script-src (including `script-src 'none'`)
+and similar directives do not apply.
+
 However, since the script executes in the main world, other CSP restrictions
 (including connect-src and which sources may be added to the document, among
 many others) *may* apply to the injected script.
@@ -123,6 +134,11 @@ scripting.executeScript() method.
 There are no necessary manifest changes.
 
 ## Security and Privacy
+
+### Availability
+
+`dom.execute()` will only be available in isolated worlds (content script worlds
+and user script worlds).
 
 ### Exposed Sensitive Data
 
@@ -172,3 +188,16 @@ For now, this API is intended to let scripts running in content or user script
 worlds inject in the main world.  In the future, we will expand this to allow a
 script to inject in additional other worlds, such as a content script injecting
 in a user script world.
+
+### Transferables
+
+It would be nice to consider supporting [transferables](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects)
+in the future. However, this has significant cost associated with it (and some
+potential other considerations), and is left out of the initial version of the
+API.
+
+### Promise support
+
+An injected function may evaluate to a promise. These cannot be serialized,
+which would (with the current proposed behavior) throw an error. We could
+instead add handling to wait for the promise to settle and return the result.
--- a/proposals/dom_execute.md
+++ b/proposals/dom_execute.md
@@ -1,0 +1,174 @@
+# Proposal: browser.dom.execute()
+
+**Summary**
+
+This API allows an isolated world (such as a content script world or user
+script world) to synchronously execute code in the main world of the document
+its injected in.
+
+**Document Metadata**
+
+**Author:** rdcronin
+
+**Sponsoring Browser:** Chromee
+
+**Contributors:** Rob--W, ...
+
+**Created:** 2024-06-24
+
+**Related Issues:** <TODO>
+
+## Motivation
+
+### Objective
+
+This allows isolated worlds to synchronously inject in the main world.  This is
+both beneficial by itself and also a prerequisite for our current plan to
+implement inter-world communication (the latter of which will be a separate
+proposal).
+
+#### Use Cases
+
+##### Executing script in the main world
+
+Today, extensions can execute script in the main world by:
+* Leveraging scripting.executeScript() or registering content scripts or user
+  scripts, or
+* Appending a new script tag to the document
+
+The former is heavily asynchronous or requires knowledge of whether to inject
+beforehand (to register a script).  The latter is very visible to the site and
+has more potential side effects.  While there's no way to fully "hide" a script
+from a site, it is desirable to avoid adding DOM elements (which could
+interfere with sites with certain properties).
+
+### Known Consumers
+
+User script managers have expressed an interest in this API, but this is also
+useful to any extension that wishes to execute script in the main world while
+leveraging information from the isolated world.
+
+## Specification
+
+### Schema
+
+```
+declare namespace dom {
+  interface ScriptInjection {
+    func: () => any;
+    args: any[];
+  };
+
+  export function execute(
+      injection: ScriptInjection
+  ): any
+}
+```
+
+### Behavior
+
+#### Function execution
+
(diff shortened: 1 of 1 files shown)


proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+
+##### Executing script in the main world
+
+Today, extensions can execute script in the main world by:
+* Leveraging scripting.executeScript() or registering content scripts or user
+  scripts, or
+* Appending a new script tag to the document
+
+The former is heavily asynchronous or requires knowledge of whether to inject
+beforehand (to register a script).  The latter is very visible to the site and
```
Rob--W, line 40: The latter is also asynchronous, because the default CSP of content scripts in MV3 forbids inline script elements, leaving the only option to be `<script src>` with extension URL as `src`.

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+
+This API allows an isolated world (such as a content script world or user
+script world) to synchronously execute code in the main world of the document
+its injected in.
+
+**Document Metadata**
+
+**Author:** rdcronin
+
+**Sponsoring Browser:** Chromee
```
carlosjeurissen:
```suggestion
**Sponsoring Browser:** Chrome
```

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+the main world of the document through either appending script tags or directly
+injecting with registered content or user scripts, or the
+scripting.executeScript() method.
+
+### Manifest File Changes
+
+There are no necessary manifest changes.
+
+## Security and Privacy
+
```
carlosjeurissen, line 137: Having this API in it's current form would not make sense outside isolated worlds.

```suggestion
### Availability

Initially `dom.execute()` will only be available in isolated worlds (content scripts and user scripts).

```

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+N/A.
+
+## Future Work
+
+### Inter-world communication
+
+This API is a necessary first step before introducing inter-world communication
+channels.  We plan to do that alongside the implementation for this API.
+
+### Specifying a target world
```
carlosjeurissen, line 185: Which means not specifying the target world in the future would default to the main world. If that is a downside or if explicitly specifying a target world right from the start is preferred specifying the target world right now world be good.

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+key.
+
+##### Return value serialization
+
+Like arguments, the return value of the execution is serialized and
+deserialized using the Structured Cloning algorithm.
+
+##### Injected script privileges
+
+The injected script has the same privileges has other script in the main world.
```
carlosjeurissen:
```suggestion
The injected script has the same privileges as other scripts in the main world.
```

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+This goes for API access, Content Security Policy, origin access, and more.
+
+##### CSP Application
+
+The injected script is not considered inline and does not have a corresponding
```
carlosjeurissen, line 115: Would this also apply if JavaScript in general is disabled? With `script-src` set to "none" for example.

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
+# Proposal: browser.dom.execute()
```
carlosjeurissen, line 1: What is the motivation to use the `dom` namespace instead of the `scripting` namespace? Considering it's similarities with `scripting.executeScript` and the seemingly lack of connection with the DOM. Considering `execute()`, `createPort()` and also `openOrClosedShadowRoot()`, would it not make sense to use `browser.document` instead?

carlosjeurissen (commented):
@rdcronin @Rob--W 
Thanks for putting this together and continue where we left off during the San Diego meeting!

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+##### Function serialization
+
+The injected function is serialized to and parsed from a string.  Bound
+parameters are not supported.  Arguments must be curried in through the `args`
+key.
+
+##### Return value serialization
+
+Like arguments, the return value of the execution is serialized and
+deserialized using the Structured Cloning algorithm.
```
Rob--W, line 99:
```suggestion
deserialized using the Structured Cloning algorithm.

Thrown errors are also serialized using the Structured Cloning algorithm.
When an input parameter or return value cannot be serialized, an error is thrown synchronously.
Notably, `Promise` instanced such as returned by `async function` cannot be serialized.
```

I elaborated on the serialization semantics. Although it would be nice to support `Promise`, I also recognize the complexity associated with supporting it, so I'm explicitly calling out non-support for `Promise`. I'm willing to consider support in the future, but note that a future expansion (`dom.createPort` proposal) also allows for async communication.

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
+# Proposal: browser.dom.execute()
+
+**Summary**
+
+This API allows an isolated world (such as a content script world or user
+script world) to synchronously execute code in the main world of the document
```
Rob--W, line 6: Is this proposal limited to the main world? I was imagining this API to also be used from content scripts to execute in a user script world.

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+other scripts running in the main world.
+
+### Open Web API
+
+The open web is unaware of the concept of multiple Javascript worlds, so this
+wouldn't belong as an open web API.
+
+## Implementation Notes
+
+N/A.
```
Rob--W, line 176: The whole proposal sketches this as a feature designed for isolated worlds. While the dom namespace is available to content scripts, it is not restricted to content scripts - regular extension documents can also access the API.

Should we restrict the availability of this new functionality to content scripts only, or do you want it to have the same availability as the current `dom` namespace?

proposals/dom_execute.md
```diff
@@ -0,0 +1,174 @@
...
+To prevent any "piercing" of the different worlds, all pieces are serialized in
+their original world, then a copy is deserialized in the target world.
+Otherwise, there would be a risk of sharing variables between worlds (even when
+non-obvious, such as by accessing the prototype of an item).
+
+##### Argument serialization
+
+Arguments are serialized and deserialized using the Structured Cloning
+algorithm.  This allows for more flexibility than simply JSON conversion.  In
+the future, some arguments may have custom serialization.
```
dotproto, line 88: Have the proposal authors considered whether arguments should support [transferable objects](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects)?

proposals/dom_execute.md
rdcronin: I think the separation between browser.dom and browser.scripting is valuable -- even though they have similar surfaces and serve conceptually-similar purposes, their use cases, restrictions, and implementations are quite different.  One is used from the extension's primary contexts (e.g., background script) to configure scripts that will be running in completely different contexts (e.g., tabs), and the other is used to interact with the same document the extension is currently running in.  IMO, this distinction more easily separates these APIs from developers, and also allows a cleaner isolation of APIs for browsers: scripting.execute() is not allowed in isolated worlds at all, since it is a privileged API.  By contrast, this is, and it is (kinda) just adding a script to the DOM (though it's closer to eval than adding a child element), and so is available in these (relatively) untrusted contexts.

> Considering execute(), createPort() and also openOrClosedShadowRoot(), would it not make sense to use browser.document instead?

I don't have strong preferences for browser.dom vs browser.document _except_ that we already have browser.dom : )  It doesn't seem worth the churn to migrate to browser.document at this stage to me, given the impact on the ecosystem and the delay it would entail to introduce new APIs (since I *do* think we'd then want to migrate fully away from dom before moving to document to avoid just having two APIs to maintain).

proposals/dom_execute.md
rdcronin: This is explicitly called out as future work - let's continue the discussion there (since Carlos added a comment on it, too).

proposals/dom_execute.md
rdcronin: Done

proposals/dom_execute.md
rdcronin: Good call; added.

proposals/dom_execute.md
rdcronin: We did consider it, but opted for structured cloning in order to reduce the implementation cost (allowing us to get this out sooner).  I also suspect transferables may have some interesting considerations with cross-world boundaries.  I've added this as potential future work.

proposals/dom_execute.md
rdcronin: Incorporated.  I think it's safe to say that we would never serialize a promise as a result, and would instead add handling to support waiting for the promise to be settled.  I've added a brief section to future work.

proposals/dom_execute.md
rdcronin: Whoops!  Fixed : )

proposals/dom_execute.md
rdcronin: Yes, I think this should bypass a script-src "none" setting from the site.  I've added that here.

proposals/dom_execute.md
rdcronin: Added!  (Though I dropped the "initially", since we have no plans to change that, and all proposals are just "initial" : ))

proposals/dom_execute.md
rdcronin: This is related to Carlos's comment above.  I think restricting this to isolated worlds (content and user script) makes.  I can't see a compelling use case for exposing it in main world pages, even for an extension document.

proposals/dom_execute.md
rdcronin: That's correct.  I think defaulting to the main world makes sense.  However, I don't feel strongly here, so if folks would prefer to have a target world specified in all cases, I'm okay with that.  @Rob--W or others, any preference?

proposals/dom_execute.md
```diff
@@ -0,0 +1,203 @@
...
```
tophf, line 88: If `dom.execute` doesn't support DOM nodes, it will cause problems for extension developers because dealing with DOM nodes is what content scripts usually do and `chrome.dom` namespace strongly implies that DOM is supported.

To be precise, it should be `EventTarget`, the ancestor of `Node`, which additionally allows transferring things like `iframeElem.contentWindow`, `window.parent` or `event.source` from a postMessage listener.

Although technically this is already possible by sending a `MouseEvent` with `relatedTarget` set to the EventTarget instance, but this trick is largely unknown. World isolation is maintained automatically: the event keeps only the spec'd part of the object i.e. without this JS world's expando properties.

Issue #284, opened 2022-09-29
Main world Content Script shared params


jonathanKingston:
Passing state/parameters to content scripts that are injected into the main world currently requires message passing that is visible to the page and also delays the functionality that the script can provide.

A use case DuckDuckGo need this for is:
- Overriding/removing page prototypes (for privacy reasons) but allowing breakage to be remotely / user configured to disable such protections.

_chrome.scripting.globalParams_ was added to solve similar use-cases however it's not ever exposed to the main world scripts. I'd instead propose a way to allow these variables to be passed into the script (perhaps cloned so that modification isn't possible) some avenues to inject these would be:

- document.currentScript.params
- import.meta.params
- Inject in a script local variable prior to execution of the script. (I believe all JS engines have the ability to easily register script local properties programatically)

These shouldn't be exposed on globalThis/window as the page should have no visibility of these parameters; most importantly the page should have no ability to read the contents.

Moved out from: https://github.com/w3c/webextensions/issues/279#issuecomment-1256989025
- This would be much more appealing than asking for another permission / user consent etc.
- Dynamic scripts pose a large security risk; if this is implemented correctly it has limited security issues.

cc @dotproto see also: https://bugs.chromium.org/p/chromium/issues/detail?id=1054624#c47

ghostwords:
Could this also be used to facilitate secure `document_start` main-world-to-content-script communication (https://github.com/w3c/webextensions/issues/78#issuecomment-1243309960, https://bugs.chromium.org/p/chromium/issues/detail?id=1054624#c61)?

cc @hackademix

ghostwords:
Previously: #103

jonathanKingston:
> Could this also be used to facilitate secure document_start main-world-to-content-script communication (https://github.com/w3c/webextensions/issues/78#issuecomment-1243309960, https://bugs.chromium.org/p/chromium/issues/detail?id=1054624#c61)?

Yes 👍🏻 kzar is actually on my team so was documenting our workaround.

jonathanKingston:
In the latest meeting I think there was some confusion with this being for _executeScript_ this would be additionally required for _chrome.scripting.registerContentScripts_ executed prior to the page load.

One suggestion was to use func; perhaps this could be resolved with allowing func to registerContentScripts such that the following is possible:

```js
let someSharedGlobal = 0
browser.alarms.onAlarm.addListener(() => {
  someSharedGlobal++
});

chrome.scripting.registerContentScripts([{
[...]

PR #679, opened 2024-08-22
Proposal: dom.createPort()

rdcronin:
This adds a proposal for a new `dom.createPort()` method, which will allow communication between different JavaScript "worlds". It leverages the proposed `dom.execute()` method (#678).

### DIFF
--- a/proposals/dom_create_port.md
+++ b/proposals/dom_create_port.md
@@ -1,0 +1,206 @@
+# Proposal: Inter (JS) World Communication
+
+**Summary**
+
+This introduces a special message port that can be passed and shared between different JS "worlds" (contexts) in the same document.
+
+**Document Metadata**
+
+**Author:** rdcronin
+
+**Sponsoring Browser:** Chrome
+
+**Contributors:** Rob--W, ...
+
+**Created:** 2024-06-24
+
+**Related Issues:** <TODO>
+
+## Motivation
+
+### Objective
+
+This allows the ability to establish a communication channel between different
+active JS worlds they may have.  This enables coordination between e.g. a main
+content script and any scripts that may inject in the main world, or between a
+content script and other user scripts.
+
+#### Use Cases
+
+The primary use case is coordinating work between different JS worlds in a way
+that is more difficult for any other script in the main world to intercept.
+This helps extensions avoid interacting with the existing page script, when
+doing so is undesirable.  This also allows executing code based on data within
+a content script without leaking the details of that content script to the main
+world.  (Though we don't consider the isolated world boundary a security
+boundary, it is useful isolation and can serve as a "first line of defense").
+
+### Known Consumers
+
+User script managers have expressed an interest in this API.  More broadly,
+this would help any extension that needs to communicate between the main world
+of a web page with a more trusted script.
+
+## Specification
+
+### Schema
+
+```
+declare namespace dom {
+  interface PortProperties {
+    world: ExecutionWorld;
+    worldId?: string;
+  };
+
+  interface MessagePort {
+    sendMessage(args?: any): void
+    onMessage(args?: any): void
+  };
+
+  export function createPort(
+      properties: PortProperties
+  ): MessagePort;
+}
+```
+
+### Behavior
+
+#### Creating a Port
+
+An extension can create a new message port using `dom.createPort()`.  This will
+create and return a new message port according to the specifies `properties`
+(see below).
+
+#### Port Properties and World ID
+
+A message port can only be used to communicate with a single other world.  This:
+* Avoids the "many-to-one" opener-listener behavior that existing message ports
+  (created by `runtime.connect()`) in extensions have.  This behavior has caused
+  increased complexity and developer confusion.  A given port will have only one
+  channel.
+* Protects against accidentally sharing the port with another world.
+
+The caller indicates which world it would like the port to be associated with
+by specifying the `world` property to the appropriate type of execution world
+and, if appropriate, the `worldId` (e.g. for specific user script worlds).
+
+#### Passing a Message Port
+
+A message port can be passed to another world (in order to establish a
+connection) by leveraging the new `dom.execute()` API.  `dom.execute()` will be
+expanded to have special serialization logic for the MessagePort type.
+
+#### Example Usage
+
+Script that executes in the main world:
+```js
+function getFoo() { ... }
+function getBar() { ... }
+// Other code dispatches 'foochanged' and 'barchanged' events.
+```
+
+Content script:
+```js
+// This function will execute in the main world.
+function setUpMainWorld(port) {
+  // This responds to a message received from the content script.
+  port.onMessage = (message) => {
+    if (message == 'getFoo') {
+      const foo = getFoo();
+      port.sendMessage({foo});
+    } else if (message == 'getBar') {
+      const bar = getBar();
+      port.sendMessage({bar});
+    }
+  };
+
+  // These notify the content script of changes.
+  addEventListener('foochanged', () => {
+    port.sendMessage({foo: getFoo()});
+  });
+  addEventListener('barchanged', () => {
+    port.sendMessage({bar: getBar()});
+  });
+}
+
+// The rest of this code executes in the content script world.
+const mainWorldPort = browser.dom.createPort({world: 'MAIN'});
+
+mainWorldPort.onMessage = (message) => {
+  if (message.foo) {
+    updateFoo(message.foo);
+  } else if (message.bar) {
+    updateBar(message.bar);
+  }
+};
+
+browser.dom.execute(
+    {
+      func: setUpMainWorld,
+      args: mainWorldPort,
+    });
+
+function fetchFoo() {
+  mainWorldPort.sendMessage('getFoo');
+}
+
+function fetchBar() {
+  mainWorldPort.sendMessage('getBar');
+}
+```
+
+##### Message serialization
+
+Message arguments are serialized and deserialized using the Structured Cloning
+algorithm.  This allows for more flexibility than simply JSON conversion.
+Message ports cannot pass other message ports in arguments.
+
+### New Permissions
+
+There are no new permissions for this capability.  This does not allow any new
+data access, since it is only accessible once the extension has already
+injected a script into the document.  Extensions can also already interact with
+the main world of the document through either appending script tags or directly
+injecting with registered content or user scripts, or the
+scripting.executeScript() method.
+
+### Manifest File Changes
+
+There are no necessary manifest changes.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+This does not result in any new data being exposed.
+
+### Abuse Mitigations
+
+This doesn't enable any new data access.  To reduce risk of cross-world
+contamination, extensions must specify the world with which they want to
+communicate, and all arguments are copied rather than directly shared.
+
+### Additional Security Considerations
+
+N/A.
+
+## Alternatives
+
+### Existing Workarounds
+
+Today, to communicate between JS worlds, extensions can use custom events as
+described in the [content script
+documentation](https://developer.chrome.com/docs/extensions/develop/concepts/content-scripts#host-page-communication).
+This is fragile and hacky, and can lean to leaking more data to the embedding
+page.
+
+### Open Web API
+
+The open web is unaware of the concept of multiple Javascript worlds, so this
+wouldn't belong as an open web API.
+
+## Implementation Notes
+
+N/A.
+
+## Future Work


proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+// This function will execute in the main world.
+function setUpMainWorld(port) {
+  // This responds to a message received from the content script.
+  port.onMessage = (message) => {
+    if (message == 'getFoo') {
+      const foo = getFoo();
+      port.sendMessage({foo});
+    } else if (message == 'getBar') {
+      const bar = getBar();
+      port.sendMessage({bar});
```
carlosjeurissen, line 113: Aligning with the code below to make the examples more readable / prevent perceived differences where they do not apply.

```suggestion
      port.sendMessage({foo:  getFoo()});
    } else if (message == 'getBar') {
      port.sendMessage({bar: getBar()});
```

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+  addEventListener('foochanged', () => {
+    port.sendMessage({foo: getFoo()});
+  });
+  addEventListener('barchanged', () => {
+    port.sendMessage({bar: getBar()});
+  });
+}
+
+// The rest of this code executes in the content script world.
+const mainWorldPort = browser.dom.createPort({world: 'MAIN'});
```
carlosjeurissen, line 127: Considering the port needs to be passed to the main world anyway. What is the reason to require explicitly specifying the world?

carlosjeurissen (commented):
@rdcronin @Rob--W Thanks for putting this together and continue where we left off during the San Diego meeting!

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+    updateFoo(message.foo);
+  } else if (message.bar) {
+    updateBar(message.bar);
+  }
+};
+
+browser.dom.execute(
+    {
+      func: setUpMainWorld,
+      args: mainWorldPort,
```
Rob--W, line 140:
```suggestion
      args: [mainWorldPort],
```

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+```
+declare namespace dom {
+  interface PortProperties {
+    world: ExecutionWorld;
+    worldId?: string;
+  };
+
+  interface MessagePort {
+    sendMessage(args?: any): void
+    onMessage(args?: any): void
```
Rob--W, line 57: Just to confirm that you are intentionally using a new design here:

There is no precedent in extension APIs for using properties to control API behavior - API interaction is always through methods. Some APIs can accept function members on objects passed to functions (e.g. `contextMenus.update(id, { onclick })`), but there is no namespace whose behavior changes through properties.

If this new pattern is not intentional, `setMessageListener( MessageListenerCallback? );` would fit the usual conventions (I don't mention `getMessageListener` because there is presumably no need to change the listener later).

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+
+Today, to communicate between JS worlds, extensions can use custom events as
+described in the [content script
+documentation](https://developer.chrome.com/docs/extensions/develop/concepts/content-scripts#host-page-communication).
+This is fragile and hacky, and can lean to leaking more data to the embedding
+page.
+
+### Open Web API
+
+The open web is unaware of the concept of multiple Javascript worlds, so this
```
Rob--W, line 199: While the open web does not support the concept of multiple worlds, it does already have a concept of a `MessagePort` (https://developer.mozilla.org/en-US/docs/Web/API/MessagePort). It's worth calling this out explicitly, and also mention that we cannot use this because (1) it is async and (2) the use of web prototypes would enable web pages to intercept messages (e.g. through `MessagePort.prototype` or `Event.prototype`).

Here is an example:

```javascript
mc = new MessageChannel();
mc.port1.onmessage = e => console.log(e.data);
mc.port2.postMessage(2);
console.log(3);

Result:
3
2
// ( 2 logged after 3 means that the message delivery was async )
```

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+wouldn't belong as an open web API.
+
+## Implementation Notes
+
+N/A.
```
Rob--W, line 204: Expand with notes on the prototype (maybe at the behavior section?):

```suggestion
### Prototype
To avoid interception by web pages, the prototype chain of the `MessagePort`
instance should terminate at `null` instead of `Object.prototype`.

Concretely, this means that an instance of `MessagePort` could either inherit
from a (hidden) MessagePort class \* (which itself has a `null` Prototype), or
be a plain object with a null Prototype and all specified members as own
property descriptors.

\* hidden means: not exposed as a global class. If anyone (e.g. the web page)
ever gets access to a `MessagePort` instance, it may see the hidden prototype
via `port.__proto__` or `port.constructor.prototype` and through that intercept
& interfere with all future port communications.
```

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+    if (message == 'getFoo') {
+      const foo = getFoo();
+      port.sendMessage({foo});
+    } else if (message == 'getBar') {
+      const bar = getBar();
+      port.sendMessage({bar});
+    }
+  };
+
+  // These notify the content script of changes.
```
Rob--W, line 117:
```suggestion
  // These notify the content script of changes.
  // NOTE: addEventListener used in this example is not part of the API.
```

Adding emphasis on the fact that `addEventListener` is NOT part of the API. Because if it were, it would easily be intercepted by the web page.

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+### Abuse Mitigations
+
+This doesn't enable any new data access.  To reduce risk of cross-world
+contamination, extensions must specify the world with which they want to
+communicate, and all arguments are copied rather than directly shared.
+
+### Additional Security Considerations
+
+N/A.
+
```
Rob--W, line 186: I'm suggesting to include a short write-up of the brittle safety offered and not-offered by the API. This clarifies what the API does and does not do to those who are not experts in the subject.

````suggestion

#### Secrecy of data

This API is designed to offer a secure communication channel between one world
and an untrusted world. It is the responsibility of the extension to maintain
further secrecy beyond that. In general, this requires avoiding the use of
prototype members.

For example, imagine a func that reads `message.foo` in the main world. This
is safe if `message` is always an object with the `foo` member. If this is not
the case, then it is a language feature of JavaScript to look up the `foo`
member in the prototype chain. Most objects ultimately inherit from
`Object.prototype`, which can be manipulated by other (untrusted) scripts to
intercept the message.

Example:

```javascript
// Attack:
Object.defineProperty(Object.prototype, "foo", {
  get() { console.log("Intercepted via foo:", this); return "intercepted" }
});

// Safe use: foo always defined, web page cannot intercept access.
message = { secret: 1, foo: true };
console.log("message.foo=", message.foo);
// Logs "message.foo= true"

// Unsafe use: foo not set; web page can intercept access attempt.
message = { secret: 1 };
console.log("message.foo=", message.foo);
// Logs "Intercepted via foo: { secret: 1 }
// Logs "message.foo= intercepted"
```

````

File: proposals/dom_create_port.md (hunk 4 above)
Comment (Rob--W) [inline proposals/dom_create_port.md:57]:
```suggestion
    sendMessage(message?: any): void
    onMessage(message?: any): void
```

To avoid confusion with `args` in `dom.execute`, suggest renaming to `message`.

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+The caller indicates which world it would like the port to be associated with
+by specifying the `world` property to the appropriate type of execution world
+and, if appropriate, the `worldId` (e.g. for specific user script worlds).
+
+#### Passing a Message Port
+
+A message port can be passed to another world (in order to establish a
+connection) by leveraging the new `dom.execute()` API.  `dom.execute()` will be
+expanded to have special serialization logic for the MessagePort type.
+
```
Rob--W, line 92: Here is a more explicit description of the message dispatch semantics. This defines synchronicity, error handling and non-errors.

```suggestion

### Message dispatch

Upon calling `sendMessage`, `onMessage` is synchronously called before
`sendMessage` returns. `sendMessage` returns successfully if `onMessage` has
been called, even if the function itself threw an error.

`sendMessage` may throw if the input message cannot be serialized.

`sendMessage` may throw if the return value cannot be serialized.

`sendMessage` may throw when the recipient is unavailable. For example, if the
recipient did not assign an `onMessage` listener. Or if the port was never
delivered to the other world.

```

I am also willing to use the phrase `throws` instead of `may throw` if you agree with that.

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+The caller indicates which world it would like the port to be associated with
+by specifying the `world` property to the appropriate type of execution world
+and, if appropriate, the `worldId` (e.g. for specific user script worlds).
+
+#### Passing a Message Port
+
+A message port can be passed to another world (in order to establish a
+connection) by leveraging the new `dom.execute()` API.  `dom.execute()` will be
+expanded to have special serialization logic for the MessagePort type.
```
Rob--W, line 91:
```suggestion
expanded to have special serialization logic for the MessagePort type.

Once a message port has been passed to a `dom.execute` call, it cannot be passed
again - an error will be thrown because it is not structurally cloneable.

```

I'm explicitly noting that the port cannot be reused. And also emphasizing that the port is not structurally cloneable, because making the type cloneable could be a trivial way to make it fit in the `dom.execute` requirement of `args` being strucurally cloneable (which would trigger a bunch of new issues, such as the "clone" potentially inverting roles without `dom.execute`, e.g. when calling `structuredClone(port)`).

Rob--W (changes requested):
I've left some comments but also suggestions to specify underspecified behavior. For ease of editing, you can commit these suggestions with Github's UI (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes) and then use `git pull origin refs/pull/679/head` to pull the new changes locally.

I'll use "Squash & Merge" before committing the PR, so the number of intermediate commits do not matter.

proposals/dom_create_port.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Specification
+
+### Schema
+
+```
+declare namespace dom {
+  interface PortProperties {
+    world: ExecutionWorld;
+    worldId?: string;
```
oliverdunk, line 52: Both @carlosjeurissen and @xeenon had mentioned that this might be redundant, and I can't immediately see a strong need for this (it is only useful if an extension sends the port to multiple worlds, which seems unlikely - and we might be able to just make it so the port only works in the first world it is sent to). Any thoughts on that?

proposals/dom_create_port.md
dotproto: I tried to find an existing web specification that has interfaces that define an interface or object that inherits from `null`. The closest thing I was able to find was the [create an exports object](https://www.w3.org/TR/wasm-js-api-2/#create-an-exports-object) algorithm from the WebAssembly JavaScript Interface working draft.

proposals/dom_create_port.md
dotproto: Should we rename `MessagePort` to `SyncMessagePort`? That would both address the naming collision called out in the previous comment and describe what is different about this interface.

proposals/dom_create_port.md
rdcronin: This wasn't to support multiple worlds (this wouldn't do that, since this still only supports one), but rather to ensure that:
a) a developer is aware that a port is bound to a _single_ world.  This is distinct from the existing uses of ports, which are a many-to-one relationship between receivers and senders.
b) a developer can't accidentally sending a port to an incorrect world.

Technically, this is redundant, since it could be derived from the first world it's sent to (and, since dom.executeScript() is proposed to be synchronous, there in theory aren't any TOCTOU issues here?), but that still seems a bit less obvious to me than binding the port at the start.  (And makes browser implementation slightly more complex, though we can always work around that.)

proposals/dom_create_port.md
rdcronin: Sorry, I meant to update this earlier.  This should be an event, not a function.  In our yet-to-be-finalized new syntax, this would be something like:

```
callback OnMessageListener = void (any message);

interface OnMessageEvent : ExtensionEvent {
  static void addListener(OnMessageListener listener);
  static void removeListener(OnMessageListener listener);
  static boolean hasListener(OnMessageListener listener);
}

interface MessagePort {
  static attribute OnMessageEvent onMessage;
  ...
}
```

Not sure if we want to use that here, or a more typescript-centric approach; lemme know if you have a preference

proposals/dom_create_port.md
rdcronin: Agreed; will incorporate

proposals/dom_create_port.md
rdcronin: I'd like to touch on the synchronocity aspect a bit more here (and I think return values are a natural byproduct of that).  I know we discussed this briefly before, but I'd like some more clarity on what this enables.  Given we have a synchronous dom.execute(), what the use cases in which synchronous message passing (as opposed to just FIFO) to a port is critical?  Is it for cases where the main world needs synchronous execution with the extension world?  If so, why?  Or for some other use case?

A not-necessarily-synchronous design leaves this more open -- we could potentially have a single world blocked while another is unblocked.  I'm honestly not sure if this exists today in Chromium (I'm not sure about other browsers), but I could see definite situations in which it could (e.g., we might pause the main world, but allow extensions to continue running).  If we guarantee this synchronocity of message passing, that means we also need to block execution in the extension world on any execution that may be happening in the main world (preventing extensions from doing work they may otherwise be able to).

proposals/dom_create_port.md
rdcronin: Ack; will incorporate

proposals/dom_create_port.md
rdcronin: See above from Oliver's comment.  This is to make it explicit and obvious to both the browser and the developer that this port is implicitly bound to one world, and disallow any accidental passing to another world or multiple worlds.

proposals/dom_create_port.md
rdcronin: Ack, will incorporate

rdcronin (commented):
Thanks, folks!  Just responding to open comments to seed some more discussion; haven't made changes to the commit just yet.

proposals/dom_create_port.md
dotproto: I'm not quite sure what you mean by "this might be redundant". Are you saying that `worldId` is not necessary, that `world` should not be required if `worldId` is present, or something else? 

The [Allow multiple user script worlds](https://github.com/w3c/webextensions/blob/main/proposals/multiple_user_script_worlds.md) proposal introduced the concept of `worldId` (along side `world`) as a property of `RegisteredUserScript`. Right now the presence of both properties on `PortProperties` follows that pattern. The main difference between how these properties are used is that this proposal defines `world` as a required property while in the other prooposal its optional. That allows developers to omit the `world` property when targeting a specific user script world or to omit `worldId` when targeting `MAIN` or `USER_SCRIPTS` worlds.

The only scenario I can think of where requiring `world` might be useful is if we anticipate allowing other types of script injection to occur in separate, isolated worlds. Or, at least want to reserve the ability to introduce such a concept in the future.

oliverdunk: > Are you saying that worldId is not necessary

This was the intended meaning of my original comment. Thinking about it some more (and as Devlin touched on below) the `world` property is also not strictly necessary.

I can see the argument for requiring these, but right now I'm not convinced the benefit is worth it. If a developer tries to send a port to a second world, we can easily throw a "Port cannot be sent to multiple worlds" error.

That seems like enough for developers to discover the right path without making the API more verbose.

proposals/dom_create_port.md
Rob--W: Generally, synchronous message passing enables the already injected main world script to reach back to the isolated world to look up relevant state that the script in the main world needs.

Other extension authors have already commented on some use cases at https://github.com/w3c/webextensions/issues/284

As an example of a concrete use case from personal experience: I have an extension called Don't Track Me Google, which is just a content script that ensures that <a href="//google.com/url=realurl">` becomes `<a href="realurl">`. Among other things, the extension intercepts the setter of `HTMLAnchorElement.prototype.href`, to hook in on calls from the web page. It sanitizes it and calls the original setter.

To minimize interference from the web page, the sanitization logic should run in an isolated world. And because the original caller expects a synchronous API, the messaging API has to be synchronous.

This specific use case also shows why sometimes it is desirable to be able to receive DOM objects, so that the page and caller can pass the DOM elements (that they already share!) without requiring appending to the DOM.

Note: this extension was originally a MV2 extension. To convert to MV3, I had to duplicate the sanitization logic in the main world and use a shared DOM element for communicating specific state (preferences from the options page) from the content script to the main world script: https://github.com/Rob--W/dont-track-me-google/blob/a73ab8b3bfa9794c75cb4e111c6424e31e6c019e/main_world_script.js. The shared DOM element can be detected by the web page. In my case it's a reasonable trade-off, especially because I don't expect Google web properties to intentionally break the extension. This assumption is not necessarily true for extensions that are privacy-focused and designed for the whole web.

> A not-necessarily-synchronous design leaves this more open -- we could potentially have a single world blocked while another is unblocked. I'm honestly not sure if this exists today in Chromium (I'm not sure about other browsers), but I could see definite situations in which it could (e.g., we might pause the main world, but allow extensions to continue running). If we guarantee this synchronocity of message passing, that means we also need to block execution in the extension world on any execution that may be happening in the main world (preventing extensions from doing work they may otherwise be able to).

There is already a synchronous communication between worlds - they share the same DOM, and therefore DOM events.

proposals/dom_create_port.md
tophf: > Should we rename MessagePort to SyncMessagePort?

Rather `MessagePortSync` because `Sync` at the end is the established pattern in the web platform (FileEntrySync) and nodejs (fs.readFileSync), whereas `Sync` at the beginning indicates it being the primary subject e.g. SyncEvent is for synchronization events.

tophf: > we cannot use MessagePort because it is async

Indeed, so maybe just make a new simpler API that doesn't use the concept of ports, but just establishes one secure synchronous channel.

proposals/dom_create_port.md
tophf: > Given we have a synchronous dom.execute()

Abusing it to pass messages synchronously would be very inefficient as running code requires a lot more internal work on top of passing the arguments. Userscripts may exchange thousands of messages between the worlds.

> use cases in which synchronous message passing [...] is critical

It allows implementing simple super-performant API via object getters exposed to the userscript or main worlds that receive the actual info from the secure world synchronously.

Tampermonkey, Violentmonkey, Firemonkey would like to keep supporting the legacy GM_xxx API, which is synchronous. This alone seems critical as there are thousands of userscripts that still work even though they are abandoned and no longer developed.

> we could potentially have a single world blocked while another is unblocked

They share the same JS/DOM thread, so what you describe is impossible.

> If we guarantee this synchronocity of message passing, that means we also need to block execution in the extension world on any execution that may be happening in the main world (preventing extensions from doing work they may otherwise be able to).

No, we just abandon the idea of MessagePort altogether and use CustomEvent mechanism internally, which already does it synchronously and which is what userscript extensions have been always using. The side benefit would be the API may be simplified as there's no need for a port.

tophf:
> A message port can be passed to another world [...] by leveraging the new dom.execute() API

But this won't support userscripts registered via userScripts.register() API in the target world (e.g. "main"), so we will have register them in the primary userscript world and then run them via dom.execute(), which is pretty wasteful due to the additional serialization/deserialization and various other JS checks.

Possible solutions:
* specify the userscript id in dom.execute() and allow us to register() this id with `runAt: 'manual'`
* introduce a new synchronous event available in the userscript/isolated worlds prior to running a userscript in the main world e.g. chrome.dom.onBeforeExecute, which would allow passing the port as `window.randomId` in the main world using the specified random id, the same random id would be embedded in the code of the main-world userscript so it'll read the value synchronously, delete window.randomId, and use it to establish the secure channel.

tophf:
It's also imperative to have the ability to pass DOM nodes (live and detached, from either a shadow or the main tree) and a `window` object. The recipient receives its own world's version of the object, of course. Currently ***monkey extensions are doing it by dispatching two messages: CustomEvent with the data and a MouseEvent with relatedTarget set to the node/window (the actual supported type is `EventTarget`).

❌ It means that a MessagePort cannot be used.

* An ideal solution would be a CustomEvent modified to allow DOM nodes and windows anywhere inside the object.
* An acceptable solution would be to have a dedicated `sendNode` method in addition to the main `send` method.

tophf:
> Today, to communicate between JS worlds, extensions can use custom events [...]
This is fragile and hacky, and can lean to leaking more data to the embedding page.

To clarify, it's not that using `CustomEvent` per se is fragile and hacky, but using any built-in global or prototype is unsafe (e.g. MessagePort) due to a bug in all browsers that allows the embedder to poison the prototypes of a same-origin iframe before document_start content script runs inside. There's also the problem of passing the random event id between worlds to avoid eavesdropping by other extensions or the web page scripts, although it's trivially solved by the upcoming dom.execute() in #678, which only leaves the problem of poisoning.

There is a way to use CustomEvent mechanism safely for a communication channel - just don't depend on current prototypes, i.e. addEventListener, dispatchEvent, CustomEvent constructor, CustomEvent.prototype.detail getter all should be used from a safe source, not from the current JS environment.

Current workaround for extensions is to extract these safe original functions from a temporary iframe that they create in their secure world using a function extractor that runs synchronously in the target world ("main").

Browsers should be able to do it without creating an iframe:
* by saving these original functions at the initialization of the world (should extensions specify some flag in the registration of the content script so browsers don't do this work unnecessarily?)
* by requesting the originals from JS engine on demand (not sure this is possible though).

In ether case the object produced by dom.createPort() will use these functions internally.

---

One inconvenience of CustomEvent is the need to have a target such as `window`, which means that the event listener can be seen in devtools inspector's event listeners panel. While not a safety concern, but an ideal solution would be to use the CustomEvent mechanism directly without attaching a visible event or mark this event listener as invisible for devtools and debugger protocol.

tophf:
Compiling the reasons why MessagePort cannot be used as the internal mechanism:
* pausing JS forcibly to dispatch MessagePort messages synchronously is arguably too expensive as userscripts may exchange thousands of messages per second; it'll also always rub every browser developer the wrong way when they look at this part of the code, just like the synchronous XMLHttpRequest does. Such response is natural as making an inherently asynchronous API work synchronously is an abuse.
* MessagePort has an arguably higher overhead as it supports cross-process dispatch, which is inherently complicated and may become noticeable when exchanging hundreds or even thousands of messages per second.
* MessagePort has an arguably higher overhead as it allows cross-origin use, which implies various security checks.
* MessagePort cannot send DOM nodes, which requires a synchronous same-origin operation due to the nature of DOM, which is inherently at odds with the purpose, implementation, and design of this API. Adding support for DOM nodes to MessagePort seems unrealistic, because such an inherently antagonistic functionality would arguably require a lot of weird ifs in the source code, which reduces maintainability of the code, thus leading to [security] bugs.

Rob--W:
Note: "MessagePort" in this proposal refers to a new type. It is not the MessagePort from the web platform. I explicitly called out before that this concept already exists and that it cannot be used to serve the use case of extensions, at https://github.com/w3c/webextensions/pull/679#discussion_r1754969359

Regarding the prototypes: I'd like the proposed API to be safe against prototype tampering. This was called out at https://github.com/w3c/webextensions/pull/679#discussion_r1754990101 . There is also a special note on (un)safety of received values, at   https://github.com/w3c/webextensions/pull/679#discussion_r1755098971

tophf:
> "MessagePort" in this proposal refers to a new type

Then it must be renamed, otherwise it'll be the "Local Storage" disaster all over again, when new developers confuse chrome.storage.local with localStorage en masse and mix up their usage patterns, limitations, availability, permissions, and inspectability in devtools.

It's unclear to me that it should contain "port" in its name.

What is clear to me is that it should be based on the existing robust synchronous mechanisms such as CustomEvent (for objects) and MouseEvent (for DOM nodes), which aren't "fragile or hacky" per se as I explained in my comments above in detail.

tophf:
* It's also unclear if dom.createPort() should be a separate thing, because it seems to be only useful as a part of dom.execute().
* The addListener/removeListener/hasListener part seems an overkill.
* Exceptions/errors:
   * when `args` can't be serialized before execution
   * when an exception occurs inside the function

   Currently the API would have to throw a hard exception in both cases, meaning that the API can't be used without wrapping it in try/catch, which is not obvious and is rather inconvenient.

* A pre-existing port makes it possible to send messages into nowhere before it's connected to the world and this should be reported somehow e.g. by throwing a synchronous exception or via chrome.runtime.lastError, both are very inconvenient.

A simpler solution might be to incorporate it all in dom.execute.

<details><summary>Example 1, <code>port</code> parameter and result</summary>

```js
const {result, error, port} = dom.execute({func, worldId, args, port: true});
port.onmessage = msg => console.log('from', worldId, msg);
port.send(1);
```
`port: true` adds an implicit argument to `func`:
```js
function (arg1, ..., argN, port) {
  port.onmessage = msg => {
    console.log(msg);
    if (msg === 1) port.send(2);
  };
}
```

OTOH, it may frustrate new developers because `port.send` cannot be used during the initial execution of the function.

</details>

<details><summary>Example 2, <code>onmessage</code> parameter</summary>

```js
const {result, error, send} = dom.execute({
  func, worldId, args,
  onmessage: (msg, respond) => { console.log('from', worldId, msg); respond('ok'); },
});
send(123);
```
`onmessage` adds an implicit argument to `func`:
```js
function (arg1, ..., argN, port) {
  port.onmessage = msg => console.log(msg);
  port.send(1);
}
```

</details>
<details><summary>Example 3, <code>onmessage</code> parameter + <code>port</code> result</summary>

```js
const {result, error, port} = dom.execute({
  func, worldId, args,
  onmessage: (msg, port) => { console.log('from', worldId, msg); port.send('ok'); },
});
port.send(123);
port.onmessage = null;
```
`onmessage` adds an implicit argument to `func`:
```js
function (arg1, ..., argN, port) {
  port.onmessage = (msg, port) => console.log(msg);
  port.send(1);
}
```

</details>

PR #690, opened 2024-09-18
Proposal: add cookies.removeAll() method

aselya:
Proposal introducing a new API cookies.removeAll()

### DIFF
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -120,7 +120,7 @@ Host permissions are required for this API.
 #### Special cases:
 ```cookies.removeAll({})```
 
->An empty details object will result in all cookies being removed. 
+>An empty details object will result in an error. 
 
 ```
 cookies.removeAll({
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -9,6 +9,8 @@ Adds a new API `cookies.removeAll()` to allow for the removal of multiple cookie
 **Sponsoring Browser:**  Chrome
 
 **Contributors:**  
+- [oliverdunk](https://github.com/oliverdunk)
+- [Rob--W](https://github.com/Rob--W)
 
 **Created:**  2024-09-18
 
@@ -63,15 +65,11 @@ An `object` containing information to identify the cookie(s) to remove. It conta
 >
 >An object representing the storage partition containing the cookie.
 >
->[`topLevelSite`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#toplevelsite)  optional:
->
->A  `string`  representing the first-party URL of the top-level site storage partition containing the cookie.
->
 >[`storeId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#storeid)  optional:
 >
 >A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
 >
->[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url) optional:
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url)
 >
 >A  `string`  representing the URL associated with the cookie.
 >
@@ -123,12 +121,16 @@ The cookies API is specific to extensions
 
 Host permissions are required for this API.
 #### Special cases:
-```cookies.removeAll({})```
 
->An empty details object will result in an error. 
+```
+cookies.removeAll({})
+```
+
+>An empty details object will result in an error.
 
 ```
 cookies.removeAll({
+    url: "https://example.com",
     partitionKey:{}})
 ```
 
@@ -136,19 +138,18 @@ cookies.removeAll({
 
 ```
 cookies.removeAll({
-    name: “example”,
-    url: “https://example.com”,
-    partitionKey:{}
-})
+    url: "https://example.com",
+    partitionKey:{hasCrossSiteAncestor : true}})
 ```
 
->If `topLevelSite` is the only argument, it will result in all cookies that have that value as the `topLevelSite` in their `partitionKey` being removed.
+>A `partitionKey` object in the details object, that does not have a `topLevelSite` and has a value for `hasCrossSiteAncestor` will result in an error.
 
 ```
 cookies.removeAll({
-    topLevelSite: “https://example.com”,
-    partitionKey:{topLevelSite: “https://foo.com”}
+    name: “example”,
+    url: “https://example.com”,
+    partitionKey:{topLevelSite: “https://example.com”}
 })
 ```
+>A `partitionKey` object in the details object, with no `hasCrossSiteAncestor` will result in cookies with both `true` and `false` values for `hasCrossSiteAncestor` being removed, that also match any other values in the details object.
 
->If the `topLevelSite` differs from the `topLevelSite`, if present, in the `partitionKey` an error will be returned.
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -69,7 +69,7 @@ An `object` containing information to identify the cookie(s) to remove. It conta
 >
 >A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
 >
->[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url)
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url) optional
 >
 >A  `string`  representing the URL associated with the cookie.
 >
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -73,7 +73,7 @@ An `object` containing information to identify the cookie(s) to remove. It conta
 >
 >A  `string`  representing the URL associated with the cookie.
 >
->A details object must contain a `url` and/or a `topLevelSite`.
+>A details object must contain at least a `url` a `partitionKey` with a valid `topLevelSite`.
 
 ##### Return
 
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -50,7 +50,7 @@ details // object
 )
 ```
 ##### Parameters
-An `object` containing information to identify the cookie(s) to remove. It contains the following properties:
+An `object` containing information to identify the cookie(s) to remove. It contains the following properties, based on the properties recognized by `cookies.remove`:
 
 >[`name`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#name) optional:
 >
--- a/proposals/Proposal: add cookies.removeAll() method.md
+++ b/proposals/Proposal: add cookies.removeAll() method.md
@@ -1,0 +1,147 @@
+**Summary**
+
+Adds a new API `cookies.removeAll()` to allow for the removal of multiple cookies with a single call.
+
+**Document Metadata**
+
+**Author:**  [aselya](https://github.com/aselya)
+
+**Sponsoring Browser:**  Chrome
+
+**Contributors:**  
+
+**Created:**  2024-09-18
+
+**Related Issues:**  [PrivacyCG/CHIPS issue 6 - How do Partitioned cookies interact with browser extensions?](https://github.com/privacycg/CHIPS/issues/6)
+
+## Motivation
+
+Prior to the introduction of [partitioned cookies](https://github.com/privacycg/CHIPS), the required parameters (`url` and `name`) used by the existing `cookies.remove()` api would be unique to a single cookie. The inclusion of an optional `partitionKey` means that there can be more than one cookie that matches the currently required parameters for `cookies.remove()`. Because of this, handling partitioned cookies with `cookies.remove()` requires additional steps which can be simplified by the inclusion of this new API.
+
+### Objective
+
+This new API addresses two workflows for developers:
+##### Removal of cookies with the same url and name combination but different partitionKey values:
+Since the introduction of partitioned cookies, to remove all the cookies with the same `url` and `name` a developer must first call `cookies.getAll()` to get all of the partitioned and unpartitioned cookies associated with the values. Then use the results of that call to make individual calls to `cookies.remove()` to delete each cookie. 
+
+##### Removal of cookies associated with a topLevelSite:
+To remove all the cookies associated with a `topLevelSite`, a developer must first call `cookies.getAll({})` to retrieve all partitioned and unpartitioned cookies. Then filter the results to identify  cookies that have a value for `topLevelSite` of their `partitionKey` match the desired `topLevelSite`. Finally the developer will need to make individual calls to `cookies.remove()` to delete each cookie. 
+
+#### Use Cases
+
+##### Cookie Manager:
+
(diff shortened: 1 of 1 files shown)


proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+
+The cookies API is specific to extensions
+
+## Implementation Notes
+
+Host permissions are required for this API.
+#### Special cases:
+```cookies.removeAll({})```
+
+>An empty details object will result in all cookies being removed.
```
aselya: I'd like some input on this. My thought was to keep the behavior the same between `cookies.getAll` and this implementation but it might make sense to block it or direct developers to use the already existing browserData API

oliverdunk: Open to thoughts, but I'd prefer to make this an error. As you mentioned, there is already the browsingData API for this, and even just during development it would be frustrating to accidentally clear all cookies. This would also allow us to match the CookieDetails object from the remove() API where `name` and `url` are currently required.

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+>A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
+>
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url) optional:
+>
+>A  `string`  representing the URL associated with the cookie.
+
+### Behavior
+
+The API will remove all cookies that match the `details` object parameter with the exception of the cases outlined in the implementation details.
+If the extension does not have  [host permissions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#host_permissions)  for this URL, the API call will fail.
```
oliverdunk: If we allow calling this API without a `url` (see below) we should add a note here about the behavior if you only have some host permissions. Presumably, we would just remove the cookies for the sites you have access to.

oliverdunk (changes requested):
This generally looks good, thanks! I left a couple of comments.

proposals/Proposal: add cookies.removeAll() method.md
xeenon: I would also like you make `{}` be an error. In Safari, we only return cookies you have host permissions for in `getAll({})`, so that is what would be removed if we supported it with `removeAll({})`. However, it does seem to be too much of a foot gun.

I think requiring one key is still useful and safe.

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Contributors:**  
+
+**Created:**  2024-09-18
+
+**Related Issues:**  [PrivacyCG/CHIPS issue 6 - How do Partitioned cookies interact with browser extensions?](https://github.com/privacycg/CHIPS/issues/6)
+
+## Motivation
+
+Prior to the introduction of [partitioned cookies](https://github.com/privacycg/CHIPS), the required parameters (`url` and `name`) used by the existing `cookies.remove()` api would be unique to a single cookie. The inclusion of an optional `partitionKey` means that there can be more than one cookie that matches the currently required parameters for `cookies.remove()`. Because of this, handling partitioned cookies with `cookies.remove()` requires additional steps which can be simplified by the inclusion of this new API.
```
Rob--W: > Prior to the introduction of [partitioned cookies](https://github.com/privacycg/CHIPS), the required parameters (`url` and `name`) used by the existing `cookies.remove()` api would be unique to a single cookie

This claim is not true. It was already possible for a given URL + name to potentially match more than one cookie. For context, see https://issues.chromium.org/issues/40572551 and https://bugzilla.mozilla.org/show_bug.cgi?id=1387957#c9

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Author:**  [aselya](https://github.com/aselya)
+
+**Sponsoring Browser:**  Chrome
+
+**Contributors:**  
+
+**Created:**  2024-09-18
+
+**Related Issues:**  [PrivacyCG/CHIPS issue 6 - How do Partitioned cookies interact with browser extensions?](https://github.com/privacycg/CHIPS/issues/6)
```
Rob--W: I suggest to also include https://issues.chromium.org/issues/40572551 and https://bugzilla.mozilla.org/show_bug.cgi?id=1387957

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Motivation
+
+Prior to the introduction of [partitioned cookies](https://github.com/privacycg/CHIPS), the required parameters (`url` and `name`) used by the existing `cookies.remove()` api would be unique to a single cookie. The inclusion of an optional `partitionKey` means that there can be more than one cookie that matches the currently required parameters for `cookies.remove()`. Because of this, handling partitioned cookies with `cookies.remove()` requires additional steps which can be simplified by the inclusion of this new API.
+
+### Objective
+
+This new API addresses two workflows for developers:
```
Rob--W: It seems that the examples you provided is how extensions would have to achieve the functionality these days. Can you rephrase to emphasize that? To someone not deeply aware of the details, it can be read as part of the proposal for something net, not of existing behavior.

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+
+##### Syntax
+
+```
+let removed = await browser.cookies.removeAll(
+details // object
+)
+```
+##### Parameters
+An `object` containing information to identify the cookie(s) to remove. It contains the following properties:
```
Rob--W:
```suggestion
An `object` containing information to identify the cookie(s) to remove. It contains the following properties, based on the properties recognized by `cookies.remove`:
```

Added more context, to offer the reasoning for the choice of parameters. Note that the `remove` method does not contain enough properties to uniquely describe one cookie, but since `cookies.removeAll` may match multiple by design, I wouldn't expansion of the property set to be a hard requirement for this proposal.

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,147 @@
...
+>A  `string`  representing the first-party URL of the top-level site storage partition containing the cookie.
+>
+>[`storeId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#storeid)  optional:
+>
+>A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
+>
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url) optional:
+>
+>A  `string`  representing the URL associated with the cookie.
+
```
Rob--W, line 84: "Return value" is not  documented in this proposal. What are you going to return?

The `remove` method returns the deleted cookie, and that could be reasonable to do the same in `cookies.removeAll`. A concrete use case is to allow for extensions to undo the removal. Currently extensions have to call `cookies.get` or `cookies.getAll` and keep track of the cookie. This tracking requires extensions to accurately predict which cookie would be removed for a given `cookies.removeAll` call, but it is not easy to determine which cookies would be removed.

Another option is to return nothing (undefined) and encourage extension developers to call `getAll` with the given parameters, under the assumption that the `cookies.getAll` call would return exactly the same set of cookies as matched by `cookies.removeAll`. In theory this is prone to TOCTOU issues, but I think that this is not that big of a deal in practice.

Yet another option is to add an option to the `cookies.removeCookies` call to flag whether deleted cookies should be returned.

aselya:
Wanted to add some context to why I'm proposing this. 

When working on updating the implementation of the cookies API to include the `hasCrossSiteAncestor` value in the `partitionKey`, the unit testing I added as part of the change, revealed that my new implementation introduced a [bug](crbug.com/362480909) that caused multiple cookies to be removed with `cookies.remove()`. I caught the bug and fixed it before it was ever merged into chromium. 

Despite the behavior being a bug, I found the ability to delete multiple cookies with the same call to be quite useful and thought it made sense to add it to the cookies api.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Thank you for the historical context. I updated language in the motivation to be less `partitionKey` specific to reflect this in addition to the update to the related issues.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Updated the related issues

proposals/Proposal: add cookies.removeAll() method.md
aselya: Updated the language and I would appreciate feedback on if my rephrasing addresses the issue you brought up.

proposals/Proposal: add cookies.removeAll() method.md
aselya: I think returning a list of cookies that were removed makes sense here. In addition to addressing the use case you mentioned it also matches the behavior of `cookies.getAll()`.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Thank you for the feedback @xeenon and @oliverdunk , having `{}` be an error works for me.

I'm inclined to agree with @xeenon, that having at least one key is useful and safe. Would requiring the caller to provide at least a `url` or a `topLevelSite` sound good?

A `url` allows for a similar requirement as the existing `cookies.remove()` and it's inclusion allows for preserving the functionality. While a `topLevelSite` enables the new functionality of removing all cookies associated with a single `topLevelSite` introduced by the API. Keeping both of these parameters optional but requiring at least one of them, matches existing APIs such as [webNavigation.getFrame()](https://developer.chrome.com/docs/extensions/reference/api/webNavigation#method-getFrame) that already have the pattern of having all optional parameters but returning an error when no parameters or an invalid combination of parameters are provided.

oliverdunk: I'm comfortable with that. Thanks!

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,157 @@
...
+```cookies.removeAll({})```
+
+>An empty details object will result in an error. 
+
+```
+cookies.removeAll({
+    partitionKey:{}})
+```
+
+>An empty `partitionKey` object as the only value in the details object, will result in all partitioned cookies being removed.
```
Rob--W: `partitionKey: {}` in `cookies.getAll` matches all cookies, partitioned and unpartitioned: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/getAll#partitionkey

Can we match that behavior, and if not, why do we deviate from it?

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,157 @@
...
+    url: “https://example.com”,
+    partitionKey:{}
+})
+```
+
+>An empty `partitionKey` object with other values populated, will result in both unpartitioned and partitioned cookies that also match the other values provided will be removed.
+
+```
+cookies.removeAll({
+    topLevelSite: “https://example.com”
```
Rob--W: Why is topLevelSite a separate key? I would expect it to be a member of partitionKey.

proposals/Proposal: add cookies.removeAll() method.md
aselya: My intention was to allow a caller to easily choose between removing just partitioned cookies  or just unpartitioned cookies (by not including a partitionKey). But it keeping the semantics the same between this API and `cookies.getAll` is better for developers as it's one less quirk for them to remember.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Having `topLevelSite` be a separate (optional) parameter addresses the second motivation "Removal of cookies associated with a topLevelSite". 

This could be replaced by having the developer pass in a partitionKey that contains the `toplevelSite` and no value for `hasCrossSiteAncestor`, such as`partitionKey:{topLevelSite: "https/example.com"}`.  But I wanted to provide an easier option for developers who may not be as familiar with the intricacies of partitioned cookies.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Updated the API to require a `url`

proposals/Proposal: add cookies.removeAll() method.md
aselya: Based on feedback, I updated the api design, to no longer have `topLevelSite` as a separate key.

oliverdunk (approved):
This looks good to me! Adding @rdcronin to make sure he is happy before we land.

aselya:
@xeenon @Rob--W any concerns with this API?

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,155 @@
...
+>
+>[`partitionKey`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#partitionkey)  optional:
+>
+>An object representing the storage partition containing the cookie.
+>
+>[`storeId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#storeid)  optional:
+>
+>A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
+>
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url)
```
Rob--W:
```suggestion
>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url) optional
```

The sentence two lines below implies that `url` can be optional.

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,155 @@
...
+##### Return
+
+> `Promise<Cookie[]>`:
+> 
+> All the cookies that were removed. If list is empty then no cookies were removed.
+
+
+### Behavior
+
+The API will remove all cookies that match the `details` object parameter with the exception of the cases outlined in the implementation details.
```
Rob--W: "all cookies that match" is doing some heavy lifting. Would it be meaningful to refer to `cookies.getAll` as the primitive? Or are there differences?

Rob--W (approved):
Some examples with non-trivial cases would be nice, but

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,155 @@
...
+>
+>[`storeId`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#storeid)  optional:
+>
+>A  `string`  representing the ID of the cookie store to find the cookie in. If unspecified, the cookie is looked for by default in the current execution context's cookie store.
+>
+>[`url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/remove#url)
+>
+>A  `string`  representing the URL associated with the cookie.
+>
+>A details object must contain a `url` and/or a `topLevelSite`.
```
rdcronin: nit: topLevelSite is not a (direct) property here, so maybe tweak to:

```
>A details object must contain at least a `url` a `partitionKey` with a valid `topLevelSite`.
```

proposals/Proposal: add cookies.removeAll() method.md
```diff
@@ -0,0 +1,155 @@
...
+
+##### Return
+
+> `Promise<Cookie[]>`:
```
rdcronin, line 80: /cc @Rob--W since I know he normally has thoughts on performance-related aspects.

The existing remove() method doesn't return a full cookie, but rather a subset of its properties: name, partitionKey, storeId, and url.  I'm curious if we should perhaps lean on that instead, which avoids serializing quite as much data -- in particular, it avoids passing all the values to the extension (which may be quite large in practice).

WDYT?

(If we go that route, we could introduce a common type for it -- `CookieSummary`, `CookieMetadata`, or similar?)

rdcronin (approved):
Thanks, @aselya !  Generally LG, but a few comments.

proposals/Proposal: add cookies.removeAll() method.md
aselya: Made update to language that references `cookies.getAll`

proposals/Proposal: add cookies.removeAll() method.md
aselya: Done

xeenon (approved):

PR #704, opened 2024-09-29
Publish minutes of 2024-09-23 - 2024-09-27 TPAC meetings

Rob--W:
Generated from https://docs.google.com/document/d/1Cg1hHvR7lAYUk4MPYKZPdlfpkMg6eLsB_DJxiC2jxSM/edit using the tool and process from https://github.com/w3c/webextensions/pull/105. I manually fixed up (code) formatting issues.

Last week, the WECG met at TPAC 2024 (#659). This PR contains the minutes of the meetings.

&bull; During the meetings on Monday, 23 september 2024, we discussed or mentioned issues #77, #78, #284, #659, #274, #296 and PRs #678, #679, #569.
&bull; During the meetings on Tuesday, 24 september 2024, we discussed or mentioned issues #702, #532, #369, #229, #636, #353, #673, #338, #619, #601, #517, #521, #680, #406, #693, #282, #252, #274, #689, #694, #646, #533, #534, #570, #507, #115, #12, #8, #1, #483, #688.
&bull; During the meetings on Thursday, 26 september 2024, we discussed or mentioned issues #362, #698, #19, #647, #416, #501, #353, #703, #519, #687.
&bull; During the meetings on Friday, 27 september 2024, we discussed or mentioned issues #683, #639, #698, #640, #615, #624, #658, #657, #645, #503, #653, #617.

For readability, the list of issue numbers along with their titles is included below.

<details>
<summary><strong>List of covered topics, issue numbers + titles</strong></summary>

Monday 23 september 2024

- #77
- #78
- #284
- #659
- #274
- #296
- PR #678
- PR #679
- PR #569

Tuesday 24 september 2024

- #702
- #532
- #369
- #229
- #636
- #353
- #673
- #338
- #619
- #601
- #517
- #521
- #680
- #406
- #693
- #282
- #252
- #274
- #689
- #694
- #646
- #533
- #534
- #570
- #507
- #115
- #12
- #8
- #1
- #483
- #688

Thursday 26 september 2024

- #362
- #698
- #19
- #647
- #416
- #501
- #353
- #703
- #519
- #687

Friday 27 september 2024

- #683
- #639
- #698
- #640
- #615
- #624
- #658
- #657
- #645
- #503
- #653
- #617

</details>

PR #599, opened 2024-04-18
Publish minutes of 2024-03-18 - 2024-03-20 meetings in San Diego

Rob--W:
Generated from https://docs.google.com/document/d/1iwpw4ljRLxcCLKInxsgvQ-6kedIXX57nBJ84oUsgK-8/edit using the tool and process from https://github.com/w3c/webextensions/pull/105. I manually fixed up the formatting of a code that appeared on the third day.

The WECG met in-person for three days on 2024-03-18 (#525). We discussed 66 issues, and also topics that don't even have an issue. This PR adds the minutes of the discussed topics on these days.

During the meetings on Monday, 18 march 2024, we discussed or mentioned issues #525, #128, #1, #7, #9, #12, #14, #274, #336, #483, #252, #319, #231, #318, #44, #134, #162, #493, #24, #25, #21, #387, #38, #40, #541, #69, #98, #99.
During the meetings on Tuesday, 19 march 2024, we discussed or mentioned issues #525, #493, #468, #394, #362, #529, #123.
During the meetings on Wednesday, 20 march 2024, we discussed or mentioned issues #525, #477, #565, #77, #229, #570, #573, #12, #91, #90, #34, #26, #244, #272, #301, #382, #326, #279, #250, #101, #138, #466, #441, #58, #87, #467, #179, #449, PR #562, #440, #439, #342, PR #543, #191, #457.

For ease of readability, I have also listed the issue numbers along with their titles below.

<details>
<summary><strong>List of covered topics, issue numbers + titles</strong></summary>

Monday 18 march 2024

- #525
- #128
- #1
- #7
- #9
- #12
- #14
- #274
- #336
- #483
- #252
- #319
- #231
- #318
- #44
- #134
- #162
- #493
- #24
- #25
- #21
- #387
- #38
- #40
- #541
- #69
- #98
- #99

Tuesday 19 march 2024

- #525
- #493
- #468
- #394
- #362
- #529
- #123

Wednesday 20 march 2024

- #525
- #477
- #565
- #77
- #229
- #570
- #573
- #12
- #91
- #90
- #34
- #26
- #244
- #272
- #301
- #382
- #326
- #279
- #250
- #101
- #138
- #466
- #441
- #58
- #87
- #467
- #179
- #449
- PR #562
- #440
- #439
- #342
- PR #543
- #191
- #457

</details>

PR #105, opened 2021-10-15
Add tool to convert Google Docs to Markdown

Rob--W:
This tool was used to generate `_minutes/2021-10-14-wecg.md` in #104, and supersedes the manual process from #23.

To use this tool:
1. Open the HTML document in a web browser.
2. Select the text in Google Docs (https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit) and paste this in the tool.
3. Click on the button to convert to markdown.
4. Copy-paste the generated markdown on a comment/issue box on Github to preview it (to check for correctness).
   (Note that the issue/comment field also linkifies `@name`-handles, but that does not happen in markdown previews in `.md` documents)

If the notes look fine, push to Github as follows:

5. Create the `meeting-yyyy-mm-dd` branch locally branched off `origin/main`, create a file called `_minutes/yyyy-mm-dd-wecg.md` and paste the markdown there. Also update the "Upcoming meetings" and "Past meetings" entries in `_minutes/README.md`. Commit the code to the branch and push the branch to the remote repo here.
5. The tool also shows a list of referenced issues. Create a pull request that mentions the listed issues and pull requests, optionally with commentary.

See #104 for the most recent example, or the pull requests linked from #23 for more comments.

PR #705, opened 2024-09-29
Update minute export tool to support rendered code

Rob--W:
After two years, another update to the minutes exporter tool that was originally introduced in #105!

Until now, code formatting required backticks in the source document. But Google Docs can also contain inline code (greenish text) or code blocks, via Markdown mode (Tools > Preferences > "Enable Markdown"). Previously, this was converted to plain text. With the updated tool, inline code is wrapped in single backticks, and code blocks are wrapped in triple backticks. Note: a code block consisting 

This updated version was used to generate the minutes in #704. Especially the code blocks on the last day (sep 27) would otherwise have required too much manual fiddling. If this update existed before, it would also have saved me some effort with formatting code in #599. From now on, rendered inline code and code blocks have first-class support.

### DIFF
--- a/_minutes/export-minutes.html
+++ b/_minutes/export-minutes.html
@@ -22,7 +22,8 @@ body {
 }
 #extraInfoOutput {
   white-space: pre-wrap;
-  height: 7em;
+  height: 8em;
+  overflow-y: auto;
 }
 #input, #output {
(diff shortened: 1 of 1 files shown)


dotproto (approved):
LGTM

PR #712, opened 2024-10-25
GetPartitionKey updates

aselya:
When implementing the getPartitionKey api, there were two issues that I felt should be updated in the proposal.

The first is removing the ability to allow for just a `frameId` to be used. Allowing this requires the browser to make an assumption on the intended tab which is not explicitly stated in the proposal and could lead to inconsistent behavior between browsers.

In chromium there are cases where the CookiePartitionKey of a cookie can be [unserializable](https://source.chromium.org/chromium/chromium/src/+/main:net/cookies/cookie_partition_key.h;l=27?q=cookiepart&ss=chromium#:~:text=//%20Methods%20for%20serializing,in%20volatile%20memory.) (if it's got a nonce or the topLevelSite is opaque). The API should return an error in this case to ensure that we are only returning valid keys with this method key.


proposals/hasCrossSiteAncestor.md
```diff
@@ -115,6 +114,7 @@ A Promise that will be fulfilled with a `Cookie.partitionKey` object that matche
 ##### Error conditions
 - If host permissions are not granted for the document whose partitionkey is getting queried an error will be returned.
 - When the parameters passed do not correspond to an existing frame, an error will be returned.
+- If the partitionKey that would be associated with the frame can not be serialized an error will be returned. This can happen when the origin associated with the topLevelSite is opaque or if the underlying key associated with the frame has a nonce.
```
Rob--W: Is this equivalent to getting an error when `document.cookie` is read/set from the web page? If not, then I would still expect a way to manage the cookies for that document.

If it is, it may be worth calling out that there cookies cannot exist in such documents.

aselya: Yes, it would be equivalent to getting an error when `document.cookie` is read/set from a web page.

My thought is that having an unserializable `partitionKey` return an error will align us with the behavior in the [HTML spec](https://html.spec.whatwg.org/multipage/dom.html#cookie-averse-document-object) where it describes throwing an error is the origin of a document is opaque.

proposals/hasCrossSiteAncestor.md
```diff
@@ -104,17 +104,21 @@ string. A UUID of the document.
 
 | Required Combinations | Notes |
 |---|---|
-|frameId | frameId must be a non-zero value|
```
Rob--W, line 107: A non-zero frameId is globally unique.

In a different extension API we agreed to support frameId without tabId, which enables matching frames that are not in tabs. The relevant discussion is at https://github.com/w3c/webextensions/issues/91

proposals/hasCrossSiteAncestor.md
```diff
@@ -104,17 +104,21 @@ string. A UUID of the document.
...
 |tabId + frameId |frameId can be 0|
 |tabId | will use the top-level frame (frameId 0)|
 |documentId| tabId/frameId are not required but permitted|
 
 ##### Return value
 A Promise that will be fulfilled with a `Cookie.partitionKey` object that matches the properties given in the details parameter and contains the `hasCrossSiteAncestor` value associated with the current cross-site status of the frame.
 
+##### Permissions required to use Cookies.getPartitionKey()
+- `host`
+- `webNavigation`
```
Rob--W: Why do we now require the webNavigation permission?

Host permissions on its own already enable extensions to query the origin/partition of a frame, including details about the full ancestor tree (through `location.ancestorOrigins`).

I don't really see a point in requiring webNavigation too.

proposals/hasCrossSiteAncestor.md
```diff
@@ -104,17 +104,21 @@ string. A UUID of the document.
...
-|frameId | frameId must be a non-zero value|
 |tabId + frameId |frameId can be 0|
 |tabId | will use the top-level frame (frameId 0)|
 |documentId| tabId/frameId are not required but permitted|
 
 ##### Return value
 A Promise that will be fulfilled with a `Cookie.partitionKey` object that matches the properties given in the details parameter and contains the `hasCrossSiteAncestor` value associated with the current cross-site status of the frame.
 
+##### Permissions required to use Cookies.getPartitionKey()
+- `host`
```
Rob--W: Did you mean host permissions? Can you clarify what exactly? Here is an example.

```suggestion
- Host permissions for the frame's document.
```

proposals/hasCrossSiteAncestor.md
aselya: Removed the requirement for webNavigation

proposals/hasCrossSiteAncestor.md
aselya: Thank you for the additional context, I'm fine adding it back in.

proposals/hasCrossSiteAncestor.md
```diff
@@ -104,17 +104,21 @@ string. A UUID of the document.
 
 | Required Combinations | Notes |
 |---|---|
-|frameId | frameId must be a non-zero value|
 |tabId + frameId |frameId can be 0|
 |tabId | will use the top-level frame (frameId 0)|
+|frameId| value must not be 0|
```
Rob--W:
```suggestion
|frameId| frameId must not be 0, when specified without tabId|
```

proposals/hasCrossSiteAncestor.md
```diff
@@ -104,17 +104,21 @@ string. A UUID of the document.
...
 ##### Return value
 A Promise that will be fulfilled with a `Cookie.partitionKey` object that matches the properties given in the details parameter and contains the `hasCrossSiteAncestor` value associated with the current cross-site status of the frame.
 
+##### Permissions required to use Cookies.getPartitionKey()
+- Host permissions for the frame's document.
+  
 ##### Error conditions
 - If host permissions are not granted for the document whose partitionkey is getting queried an error will be returned.
 - When the parameters passed do not correspond to an existing frame, an error will be returned.
+- If the partitionKey that would be associated with the frame can not be serialized an error will be returned. This can happen when the origin associated with the topLevelSite is opaque or if the underlying key associated with the frame has a nonce.
```
Rob--W: Suggested edit in relation to the discussion from https://github.com/w3c/webextensions/pull/712#discussion_r1822980547

```suggestion
- If the partitionKey that would be associated with the frame can not be serialized an error will be returned. This can happen when the origin associated with the topLevelSite is opaque or if the underlying key associated with the frame has a nonce. This is comparable to situations where `document.cookie` would throw.
```

proposals/hasCrossSiteAncestor.md
Rob--W: When I visit `data:text/html,<script>alert(document.cookie)</script>`, it throws in Chrome but does not in Firefox. I'd therefore like to emphasize that the behavior matches the usual behavior on the web platform, so that the `cookies` extension API can always manage all cookies that can be created by websites, and not a subset of them.

aselya: Resolving based on this [commit](https://github.com/w3c/webextensions/pull/712/commits/eb67586a416514a041e0f2658f88e24746489f80).

proposals/hasCrossSiteAncestor.md
aselya: Resolving based on this [commit](https://github.com/w3c/webextensions/pull/712/commits/1c38656d9093979f71d44c1434341b367168fcb0)

Rob--W (approved):

Rob--W:
Can I squash and merge or would you like to make more changes?

aselya:
@Rob--W, feel free to squash and merge.

Issue #719, opened 2024-11-02
Adding a temporary listener just for the current run of the service worker or event page so it doesn't wake up next time for this event


tophf:
The problem I'm trying to solve is that `addListener` called from a "lazy" background script (an MV3 service worker or an event page) sets an internal flag in the browser to wake up the background script next time even if I add the listener asynchronously (i.e. not during the startup of the background script), which may be wasteful in case the extension chose to recreate its state fully, for example because it's faster or more reliable than reading the saved state from indexedDB or chrome.storage.session, applying the difference, and finally saving the result to storage yet again.

There's also no way to remove the listener reliably right before the service worker terminates.

The events I have in mind are `chrome.tabs.onRemoved` and `onReplaced`, which I use in a userstyle extension to track which tab ids should be removed from my data map that I use to avoid calling chrome.tabs.get for each iframe that loads in an existing tab or when broadcasting to tabs that had been injected with the userstyles. Ideally I wouldn't want to keep hundreds of ports open just to track hundreds of tabs and frames. Instead, each time the extension wakes up for a webNavigation or a webRequest event for this frame's document it calls chrome.tabs.query, which illustrates there's no need to wake up for those auxiliary onRemoved/onReplaced events.

My workaround for the reported problem is that... I stopped using these events altogether and switched to reacting to an API error when sending a message to a missing tab, which should be rare hopefully.

Here's my idea:

* `chrome.tabs.onRemoved.addEagerListener(fn)` as an antonym to the "lazy listener" concept used in Chromium's source code.

tophf:
OTOH, the current Chromium's behavior is conceptually a bug, because an asynchronously added listener cannot hear the event next time the SW wakes up.

polywock:
I don't know the specifics, but with adding a listener asynchronously, wouldn't it just wake up once in excess? As long as you don't keep adding the listener, it shouldn't be that wasteful.

tophf:
It's necessary every time the extension wakes up, so it'll be an eternal problem.

polywock:
Ah, that makes sense.

dotproto:
In abstract I'm in favor of providing extensions with a way to register an event listener in an evented background context that won't wake the background context. The behavior described in the original issue is Chromium-specific. In Firefox and Safari event listeners registered a short period after the background context starts will not wake the context in the future.

[...]

Issue #718, opened 2024-10-30
Agenda discussion for public meeting on 2024-11-07


Rob--W:
This issue is used to collect topic suggestions for our November 7, 2024 meeting. We will stop accepting suggestions ~48 hours before the meeting and the final agenda will be added to the [meeting notes doc](https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit) ~24 hours before the meeting.

The content below this line will be copy-pasted to the agenda section of the meeting.

## Agenda

The meeting will start at 3 minutes after the hour.
See [issue 531](https://github.com/w3c/webextensions/issues/531) for an explanation of this agenda format.

 * **Announcements** (2 minutes)
   * None
 * **Triage** (25 minutes)
   * [Issue 708](https://github.com/w3c/webextensions/issues/708): Cache Partitioning, HTTP Cache Management, and Conditional DNR Rule Application
   * [Issue 711](https://github.com/w3c/webextensions/issues/711): Proposal: Enable changing the permissions for only new user
   * [Issue 713](https://github.com/w3c/webextensions/issues/713): Should tabs.Tab.lastAccessed be optional?
   * [Issue 715](https://github.com/w3c/webextensions/issues/715): Feature request: Enhance the tab group api to support the new pinning functionality
   * [Issue 716](https://github.com/w3c/webextensions/issues/716): Inconsistent support for [context]menus.getTargetElement()
   * [Issue 717](https://github.com/w3c/webextensions/issues/717): Request: [ContextMenus] Include <canvas>s in the "page" ContextType (or it's own ContextType)
   * [Issue 719](https://github.com/w3c/webextensions/issues/719): Adding a temporary listener just for the current run of the service worker or event page so it doesn't wake up next time for this event
* **Timely issues** (10 minutes)
   * [Issue 527](https://github.com/w3c/webextensions/issues/527): chrome.scripting.executeScript doesn't resolve/reject for `frozen` state tabs
 * **Check-in on existing issues** (10 minutes)
   * [Issue 19](https://github.com/w3c/webextensions/issues/19): request: APIs and infrastructure to simplify cross-browser automated testing of extensions

Rob--W:
Note: I moved all topics from the previous agenda at https://github.com/w3c/webextensions/issues/709 to this one because we canceled the previous meeting.

oliverdunk:
[...]

Issue #717, opened 2024-10-30
Feature Request: [ContextMenus] Include <canvas>s in the "page" ContextType (or it's own ContextType)


hicallmeal:
Currently, the only way for a ContextMenu item to appear on a **\<canvas>** click, is with the ["all"] ContextType. 
(From testing on Firefox and Chromium).

This adds menu items to places they aren't needed (e.g. Popup), leading to potential user confusion.

An example use case might be a "Resize this image" menu item, which would ordinarily need ["image", "page" (SVGs)] ContextTypes, but "all" for canvases (for cases where a canvas has an image). 

My request/proposal is \<canvas> inclusion in the "page" context, or it's own (if there are security concerns).

maxnikulin:
I do not need namely `CANVAS: "canvas"` in `contextMenus.ContextType`, but related Firefox `menus` extensions look promising. Unfortunately my reading of #188 is that Chrome and Safari teams are against of features in that spirit.

A more general idea was posted to the chromium-extensions mailing list (likely by @robbi73): CSS selectors to specify HTML elements meaningful for context menu actions, see [Why not brush up the contextMenu API a little? Sun, 2 Jul 2023 14:52:27 -0700 (PDT)](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/DQTjTDhFunc/m/EH7E30SVBQAJ).

To avoid context menu item in extension action popups and side panels, there is a workaround, see [Simeon Vincent. Re: [crx] Context menu appears on the pages of other extensions. It's a normal behaviour? Thu, 6 Jul 2023 15:07:39 -0700](https://groups.google.com/a/chromium.org/d/msgid/chromium-extensions/CACA%3D7_ZQ_NcwXYmXC898_UU4%3Dh%2Bq4PrRw8Kisz5umTva06T8AQ%40mail.gmail.com):
> `documentUrlPatterns` value to match `*://*/*` ... will only match against HTTP and HTTPS pages

[...]

Issue #716, opened 2024-10-30
Inconsistent support for [context]menus.getTargetElement()


hicallmeal:
There are cases where an extension might need the element a contextMenu was opened (and actioned) on. 
A classic example case might be to block/remove said element
Another, might be "Copy this element".

For this, Firefox has implemented [getTargetElement()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/getTargetElement), paired with [targetElementId](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/OnClickData#targetelementid).

For browsers without this, the best current solution (from what I've found) is to inject a content script with a contextMenu listener and get the target, i.e. 

```Javascript
function contextMenuListener(evt) {
  let targetElement = evt.target
  copyElement(targetElement) ..
}
```

The inherent problem with this is the necessity for global content script injection, preventing dynamic injection based on user action.

Another solution is [injection _onClicked_, retrieving the activeElemet](https://github.com/tophf/copy-link-text/blob/741e60c3aed14b11dbd138407676fbaccd1f4397/bg.js#L28), but for the above cases, this wouldn't suffice (with non-focusable elements and such).

As for implementation, I believe Firefox's approach is great for this. 

Note - I'm not sure whether this falls under _inconsistency_ or _Feature request_, though I believe it's implementation is warranted, so I've gone with inconsistency.

fregante:
Last discussion related to `menus` inconsistencies:

- https://github.com/w3c/webextensions/issues/188

hicallmeal:
@fregante Thanks! I hadn't seen that one. I'm actually not sure how to word that in the post, so I'll say here (just generally speaking) - 

Given the examples listed, and the desire for in-page actions based on a menu item click to be (I'd think) common enough, I'd say its addition to the ContextMenu APIs would be largely beneficial. 

I definitely can't comment on other Menus v ContextMenus differences though!

Rob--W:
Note: I designed and implemented `menus.getTargetElement()` in Firefox. There were multiple reasons for restricting it to the `menus` namespace without exposing it in the `contextMenus` namespace:

1. The API would become available in content scripts. If exposed as `contextMenus`, it would have been more likely to break extensions if there is code like this:
   ```js
   // common.js
   if (chrome.contextMenus) {
     // Script assumes that it is in the background page.
     chrome.contextMenus.onClicked.addListener(...);
   }
   ```
[...]

PR #720, opened 2024-11-07
Publish minutes of 2024-11-07 meeting

Rob--W:

During this meeting we discussed or mentioned issues #718, #531, #708, #711, #713, #715, #716, #717, #719, #527, #532, #501 and PR #546.

In this PR I also included the brief meeting notes of the previous canceled meeting (#709), because I never created a PR for it.

### DIFF
--- a/_minutes/2024-10-24-wecg.md
+++ b/_minutes/2024-10-24-wecg.md
@@ -1,0 +1,40 @@
+# WECG Meetings 2024, Public Notes, Oct 24
+
+ * Chair: Simeon Vincent
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=67198e00,384
+Call-in details: [WebExtensions CG, 24th October 2024](https://www.w3.org/events/meetings/a97adab8-e1ae-4a2b-85cf-e6b6d3d35f00/20241024T080000/)
+Zoom issues? Ping @zombie (Tomislav Jovanovic) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+
+## Agenda: [discussion in #709](https://github.com/w3c/webextensions/issues/709), [github issues](https://github.com/w3c/webextensions/issues)
+
+The meeting will start at 3 minutes after the hour.
+
+See [issue 531](https://github.com/w3c/webextensions/issues/531) for an explanation of this agenda format.
+
+ * **Announcements** (2 minutes)
+   * MEETING CANCELED, MOST FOLKS ARE OUT (AD-FILTERING DEV SUMMIT)
+ * **Triage** (15 minutes)
+   * (topics will move to the 7 nov meeting)
+   * [Issue 708](https://github.com/w3c/webextensions/issues/708): Cache Partitioning, HTTP Cache Management, and Conditional DNR Rule Application
+   * [Issue 711](https://github.com/w3c/webextensions/issues/711): Proposal: Enable changing the permissions for only new user
+ * **Timely issues** (10 minutes)
+ * **Check-in on existing issues** (20 minutes)
+
+
+## Attendees (sign yourself in)
+
+ 1. Rob Wu (Mozilla)
+ 2. Tomislav Jovanovic (Mozilla)
+
+
+## Meeting notes
+
+Meeting canceled
+
+ * [rob] Many of the usual Google and Mozilla representatives are currently absent because they are at the [Ad Filtering Dev Summit](https://adfilteringdevsummit.com/). At this exact time, Google folks are presenting “What's new in Chrome”. These sessions are recorded and will be published to Eyeo's YouTube channel.
+ * [tomislav] Canceled due to most folks being out.
+
+The next meeting will be on [Thursday, November 7th, 8 AM PST (4 PM UTC)](https://everytimezone.com/?t=672c0300,3c0). Warning: daylight saving time changes between today and the next meeting. Still 5 PM in Europe.
--- a/_minutes/2024-11-07-wecg.md
+++ b/_minutes/2024-11-07-wecg.md
@@ -1,0 +1,186 @@
+# WECG Meetings 2024, Public Notes, Nov 7
+
+ * Chair: Simeon Vincent
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=672c0300,3c0
+Call-in details: [WebExtensions CG, 7th November 2024](https://www.w3.org/events/meetings/a97adab8-e1ae-4a2b-85cf-e6b6d3d35f00/20241107T080000/)
+Zoom issues? Ping @zombie (Tomislav Jovanovic) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+
+## Agenda: [discussion in #718](https://github.com/w3c/webextensions/issues/718), [github issues](https://github.com/w3c/webextensions/issues)
+
+The meeting will start at 3 minutes after the hour.
+
+See [issue 531](https://github.com/w3c/webextensions/issues/531) for an explanation of this agenda format.
+
+ * **Announcements** (2 minutes)
+   * None
+ * **Triage** (25 minutes)
+   * [Issue 708](https://github.com/w3c/webextensions/issues/708): Cache Partitioning, HTTP Cache Management, and Conditional DNR Rule Application
+   * [Issue 711](https://github.com/w3c/webextensions/issues/711): Proposal: Enable changing the permissions for only new user
+   * [Issue 713](https://github.com/w3c/webextensions/issues/713): Should tabs.Tab.lastAccessed be optional?
+   * [Issue 715](https://github.com/w3c/webextensions/issues/715): Feature request: Enhance the tab group api to support the new pinning functionality
+   * [Issue 716](https://github.com/w3c/webextensions/issues/716): Inconsistent support for [context]menus.getTargetElement()
+   * [Issue 717](https://github.com/w3c/webextensions/issues/717): Request: [ContextMenus] Include `<canvas>`s in the "page" ContextType (or it's own ContextType)
+   * [Issue 719](https://github.com/w3c/webextensions/issues/719): Adding a temporary listener just for the current run of the service worker or event page so it doesn't wake up next time for this event
+ * **Timely issues** (10 minutes)
+   * [Issue 527](https://github.com/w3c/webextensions/issues/527): chrome.scripting.executeScript doesn't resolve/reject for frozen state tabs
+   * [PR 546](https://github.com/w3c/webextensions/pull/546): update window.browser spec per #532
+ * **Check-in on existing issues** (10 minutes)
+   * None
+
+
+## Attendees (sign yourself in)
+
+ 1. Simeon Vincent (Mozilla)
+ 2. Rob Wu (Mozilla)
+ 3. Aaron Selya (Google)
+ 4. Timothy Hatcher (Apple)
+ 5. Oliver Dunk (Google)
+ 6. Carlos Jeurissen (Jeurissen Apps)
+ 7. Patrick Kettner (Google)
+ 8. Steven McLintock (1Password)
+ 9. Cameron Eckelberry (Malwarebytes)
+ 10. Tomislav Jovanovic (Mozilla)
+ 11. Kiara Rose (Apple)
+ 12. Mukul Purohit (Microsoft)
+
+
+## Meeting notes
+
+[Issue 708](https://github.com/w3c/webextensions/issues/708): Cache Partitioning, HTTP Cache Management, and Conditional DNR Rule Application
+
+ * [oliver] Multiple questions here. Overall use case is wanting to have a custom PDF viewer. Rather than redirecting as some pages do, this extension injects an extension frame from a content script. Their approach faces challenges caused by cookie partitioning. Perhaps the issue should be broken down.
+ * [rob] Don't think we need to break down this issue. Root cause is that it's not possible to select which partition will be used to make a request. There are some not well documented mechanisms available in Chrome. It would be nice if there was a way to make a request from the right partition.
+ * [simeon] I believe we've discussed a way of selecting a partition to make a request before. Aaron may have some additional thoughts.
+ * [aaron] Specifically selecting the partition for a frame? Within the cookies API, we are defining the cookies.getPartitionKey method to retrieve the partition for a frame, including the topLevelSite and hasCrossSiteAncestor.
+ * [rob] That would provide a way to query the partition,but doesn't give us a way to specify the partition for a given request. I see two approaches: make sure the extension iframe is loaded in the right partition. Another is to provide a dedicated API for the extension to make a request from a given partition. Need to get feedback from browser vendors on preferred approach.
+   * Firefox bug that requests the capability to send requests associated with a specific partition: https://bugzilla.mozilla.org/show_bug.cgi?id=1670278
+ * [oliver] In this case, the third-party context is a frame. In the extension document there's an iframe with a third party site. You'd want to change which partition that embedded frame uses.
+ * [patrick] In this case it's specifically an embedded frame from the Adobe CDN.
+ * [rob] In that case I don't think we should allow it to select a different partition, because that iframe is regular web content.
+ * [aaron] Sounds like requestStorageAccess would be useful for this. Inside the frame the extension could call this. If the cookie is a third party unpartitioned cookie we could potentially provide access to that. This is an area I work with often. requestStorageAccess is a purely JS API. As long as the iframe can use JS it could use this. We can follow up.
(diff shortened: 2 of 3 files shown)


_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [rob] Don't think we need to break down this issue. Root cause is that it's not possible to select which partition will be used to make a request. There are some not well documented mechanisms available in Chrome. It would be nice if there was a way to make a request from the right partition.
+ * [simeon] I believe we've discussed a way of selecting a partition to make a request before. Aaron may have some additional thoughts.
+ * [aaron] Specifically selecting the partition for a frame? Within the cookies API, we are defining the cookies.getPartitionKey method to retrieve the partition for a frame, including the topLevelSite and hasCrossSiteAncestor.
+ * [rob] That would provide a way to query the partition,but doesn't give us a way to specify the partition for a given request. Two approaches: make sure the extension iframe is loaded in the right partition. Another is to provide a dedicated API for the extension to make a request from a given partition. Need to get feedback from browser vendors on preferred approach.
```
dotproto:
```suggestion
 * [rob] That would provide a way to query the partition,but doesn't give us a way to specify the partition for a given request. I see two approaches: make sure the extension iframe is loaded in the right partition. Another is to provide a dedicated API for the extension to make a request from a given partition. Need to get feedback from browser vendors on preferred approach.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+   * Firefox bug that requests the capability to send requests associated with a specific partition: https://bugzilla.mozilla.org/show_bug.cgi?id=1670278
+ * [oliver] In this case, the third-party context is a frame. In the extension document there's an iframe with a third party site. You'd want to change which partition that embedded frame uses.
+ * [patrick] In this case it's specifically an embedded frame from the Adobe CDN.
+ * [rob] In that case I don't think we should allow it to select a different partition, because that iframe is regular web content.
+ * [aaron] Sounds like requestStorageAccess would be useful for. Inside the frame the extension could call this. If the cookie is a third party unpartitioned cookie we could potentially provide access to that. This is an area I work with often. requestStorageAccess is a purely JS API. As long as the iframe can use JS it could use this. We can follow up.
```
dotproto:
```suggestion
 * [aaron] Sounds like requestStorageAccess would be useful for this. Inside the frame the extension could call this. If the cookie is a third party unpartitioned cookie we could potentially provide access to that. This is an area I work with often. requestStorageAccess is a purely JS API. As long as the iframe can use JS it could use this. We can follow up.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [oliver] Might be best to post on the issue. That said, not purely a cookies issue. They're looking to apply DNR rules. Doesn't seem like this is the ideal way to solve this for
```
dotproto:
```suggestion
 * [oliver] Might be best to post on the issue. That said, not purely a cookies issue. They're looking to apply DNR rules. Doesn't seem like this is the ideal way to solve this for the general case.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [rob] Agreed.
+ * [patrick] Ideally they would like it to not be optional, but don't want to disable the extension for existing users.
+ * [rob] For context, in Chrome new permissions cause the extension to be disabled. In Firefox the new version requires confirmation, but the old version continues working.
+ * [timothy] In Safari host permissions do not trigger anything, they are always optional in Safari. API permissions are often not warning, except in some cases, in which case the extension is disabled like Chrome (e.g. webNavigation).
+ * [tomislav] The result of the proposal would be a new kind of state where the extension in installed and working but doesn't have access to all the required permissions from the manifest. If we want to open that can of worms, that can be useful for other things as well. That would require extensions to code against that. So far the presumption has been that if it's a required permission it's always available. This would break that model.
+ * [timothy] I'm also concerned that developers won't checking permissions.contains() at runtime.
+ * [patrick] Nobody requested this as the default behavior, opt in is preferred.
+ * [simeon] …
```
dotproto:
```suggestion
 * [simeon] The current model assumes required permissions are granted. If they can sometimes be optional, developers will need to approach all permissions added after the first public release as potentially optional. Can no longer guarantee that a required permission has been granted.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [patrick] Impression I got was that this would be an upgraded permission. A fresh install would treat the permission as required. Permissions would only be treated as optional in an upgrade.
+ * [simeon] Extensions should develop as if all permissions are optional. (...) No guarantee that the contract with the platform won't change and the “required” permissions won't be treated as optional in the future.
```
dotproto:
```suggestion
 * [simeon] Extensions should develop as if all permissions are optional. Safest to assume required permissions may not be granted and code defensive. If this contract changes, no guarantee that the contract with the platform won't change further and the “required” permissions won't be treated as optional in the future.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [rob] I'm aware that this is an issue that extension developers are facing. What is the appetite from browser vendors to implement solutions to the issue here?
+ * [patrick] This is a frequently recurring issue.
+ * [carlos] I've also had this issue in the past: you introduce something for a subset of users and it causes challenges for all users.
+ * [tomislav] On the Firefox side we've been discussing some related permissions issues around what's required or optional.
```
dotproto:
```suggestion
 * [tomislav] On the Firefox side we've been discussing some related permissions issues around what's required or optional. Would be interested in further discussion.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [rob] Oliver approved the PR. I approved as well, now it just needs to be reviewed by Timothy (Apple) as well.
+ * [timothy] I'll approve today.
+
+[Issue 713](https://github.com/w3c/webextensions/issues/713): Should tabs.Tab.lastAccessed be optional?
+
+ * [oliver] There's a lastAccessed property on the Tab object. For a long time this was Firefox only, but we recently added support in Chrome. The semantics of how it's set is a bit inconsistent. My assumption was that when you first access it this field would be undefined, but Firefox sets it on tab creation.
+ * [rob] What is the impact of marking it as optional or not?
+ * [simeon] My understanding was this was more about a behavior change – that when a tab is created the value is undefined.
+ * [oliver] If Firefox's behavior to always set it is intentional, then Firefox could mark it as not optional.
+ * [timothy] We haven't implemented this in Safari yet.
```
dotproto:
```suggestion
 * [timothy] We haven't implemented this in Safari yet. Open to going with the consensus approach.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [simeon] I haven't found any specific reason for the current implementation. undefined/0 seemed like an intuitive default; the current behavior is to set the timestamp at creation.
```
dotproto:
```suggestion
 * [simeon] I haven't found any specific reason for the current implementation. When I first saw this I assumed that `undefined` would be used; seems like an intuitive default. The current behavior is to set the timestamp at creation surprised me.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [tomislav] There is no discussion about optionality at introduction; I guess that we marked is as optional because it was Firefox-only at that time and not available in all browsers. Willing to mark it non-optional.
+ * [oliver] Would you be willing to defer setting it until the tab has been accessed?
+ * [rob] Could you come up with a specification of what you want it to be in all relevant scenarios? Then we can evaluate the request on that basis instead of just wondering whether it is set or not.
+   * [oliver] I can follow up and leave a comment on the issue.
+ * [simeon] I'd suggest using 0 over undefined. It's falsey, it's a valid value for new Date(), it continues having the property be defined on the object. Most consistent with how this value may be used in the ecosystem.
```
dotproto:
```suggestion
 * [simeon] I'd suggest using `0` over `undefined`. It's falsey, it's a valid value for new Date(), it continues having the property be defined on the object. Most consistent with how this value may be used in the ecosystem.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [timothy] 0, NaN?
```
dotproto:
```suggestion
 * [timothy] What about NaN?
 * [tomislav] Produces an invalid date.
 * [timothy] Right, NaN is kind of the nuclear option.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+
+[Issue 715](https://github.com/w3c/webextensions/issues/715): Feature request: Enhance the tab group api to support the new pinning functionality
+
+ * [simeon] Feature request on tabGroups API to pin groups to the bookmarks.
```
dotproto:
```suggestion
 * [simeon] Feature request on tabGroups API to pin groups to the bookmarks. Oliver, can you speak to this capability?
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [oliver] I am not familiar with this.
+ * [rob] Does Safari support the tabGroups API?
+ * [timothy] We don't, but we have the feature and are interested in shoehorning the functionality in the API.
+ * [rob] In Firefox there's also interest in tab groups but we don't have the API yet.
+ * [oliver] The two requests that would be API level changes is to add a pinned property (Boolean) and tabs.query() include closed tab groups, which I'm not familiar with.
+ * [timothy] We have closed tab groups as well.
+ * [simeon] Would it make sense to have closed groups be a distinct type of tab to query?
+ * [timothy] Not necessarily.
+ * [rob] How does this functionality relate to sessions.restore? Based on the description it sounds similar.
```
dotproto:
```suggestion
 * [rob] How does this functionality relate to sessions.restore? Based on the description it sounds similar. Seems no one is familiar offhand.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [oliver] Pinned fields sounds non-controversial, querying/managing closed groups sounds like something needing more discussion.
+
+[Issue 716](https://github.com/w3c/webextensions/issues/716): Inconsistent support for [context]menus.getTargetElement()
+
+ * [rob] Feature request for the ability to retrieve the element that is clicked when a context menu was opened. Firefox solved this by introducing getTargetElement(). When contextMenus.onClicked fires you can get a reference (targetElementId) that you pass to the menus.getTargetElement function in a content script to retrieve the element. There are workarounds such as exposing the coordinates in the document, but they're lossy. Firefox's approach is a reliable solution.
```
dotproto:
```suggestion
 * [rob] Feature request for the ability to retrieve the element that is clicked when a context menu was opened. Firefox solved this by introducing `getTargetElement()`. When `contextMenus.onClicked` fires you can get a reference (`targetElementId`) that you pass to the `menus.getTargetElement` function in a content script to retrieve the element. There are workarounds such as exposing the coordinates in the document, but they're lossy. Firefox's approach is a reliable solution.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [timothy] Against coordinate suggestion as well; interested in implementing Firefox's getTargetElement API.
+ * [oliver] Treating this as opaque is fine, but I'm curious what this ID looks like in Firefox.
+ * [rob] It's generated in Firefox. I believe it's the timestamp. On click, a weak ref to the element is stored, timestamp is captured and passed to the extension.
```
dotproto:
```suggestion
 * [rob] It's a generated number in Firefox. I believe it's the timestamp. On click, a weak ref to the element is stored, timestamp is captured and passed to the extension.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [oliver] Thanks, just wanted to understand the general shape.
+ * [timothy] I'd want it to be opaque. UUID or some string.
+ * [rob] It's currently an integer. This enables extensions to pass the value to getTargetElement in a code string (tabs.executeScript) without serialization concerns. That concern is lesser in MV3 because scripting.executeScript does not accept code strings, and accepts args for func instead.
```
dotproto:
```suggestion
 * [rob] It's currently an integer. This enables extensions to pass the value to `getTargetElement` in a code string (`tabs.executeScript`) without serialization concerns. That concern is lesser in MV3 because `scripting.executeScript` does not accept code strings, and accepts `args` for `func` instead.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [rob] I don't think that it should be a string; the integer can be a random number, as it is only used to avoid race conditions where “the last clicked element” changed between the time of the invocation of the context menu and the invocation of the script.
+ * [tomislav] I don't think Timothy means a value you can't look into, just a value without meaning.
+ * [timothy] Exactly. Not a timestamp. No inherent meaning to the value.
+ * [rob] Thanks for the clarification - I thought that you meant a string or object by “opaque”. An integer sounds good to me.
+ * [rob] Safari seems on board. Does Chrome have any oppositions?
+ * [oliver] I can follow up.
+
+[Issue 717](https://github.com/w3c/webextensions/issues/717): Request: [ContextMenus] Include `<canvas>`s in the "page" ContextType (or it's own ContextType)
+
+ * [simeon] The only way to match &lt;canvas> elements is to specify “all”. The request is to offer a specific contextType to match &lt;canvas>.
```
dotproto:
```suggestion
 * [simeon] The only way to match `<canvas>` elements is to specify “all”. The request is to offer a specific `contextType` to match `<canvas>`.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [tomislav] I'm supportive of a canvas context type. No implementation concerns come to mind.
+ * [timothy] Commented on the issue, but I'm also supportive of a canvas type.
+ * [rob] What useful information would an extension get for a canvas? There is no URL or other relevant context.
+ * [timothy] I think this is related to the previous issue where you need to be able to work with the element to meaningfully interact with it. They lean on each other.
+ * [simeon] Sounds like we're aligned?
+ * [rob] I'd like to see more details on use cases. One of the comments was to support a CSS selector to match context menus by selector. That would be a more general way to support this capability. I'm generally in favor of CSS selector. I'm opposed to adding “canvas” ContextType when there is no special context associated with it.
+ * [timothy] We'd still have the issue of needing getTargetElement to meaningfully interact with the element
+ * [simeon] Agreed, I think we're discussing which solution to issue 717 is the right approach.
```
dotproto:
```suggestion
 * [simeon] Agreed, I think we're discussing which solution to issue #717 is the right approach.
```

_minutes/2024-11-07-wecg.md
```diff
@@ -0,0 +1,184 @@
...
+ * [timothy] I'm in favor of “canvas” context type because it is simpler than matching by CSS selector.
```
dotproto:
```suggestion
 * [timothy] I'm in favor of “canvas” context type because it is simpler than matching by CSS selector. Not worried about API bloat if this case later becomes addressable with selector matching.
```

dotproto (changes requested):

dotproto:
I have not applied these changes to the Google Doc.

_minutes/2024-10-24-wecg.md
```diff
@@ -0,0 +1,40 @@
+# WECG Meetings 2024, Public Notes, Oct 24
+
+ * Chair: Timothy Hatcher
```
xeenon:
```suggestion
 * Chair: Simeon Vincent
```

_minutes/2024-11-07-wecg.md
Rob--W: `&lt;` is a valid way to render `<` in markdown, at least on Github.

I have applied this suggestion since the title is also using backticks around the canvas; but in the future let's not wrap `<canvas>` unnecessarily. I try to minimize the amount of unnecessary formatting in the notes because it may be distracting from its content.

_minutes/2024-11-07-wecg.md
Rob--W: I usually linkify "issue" numbers (e.g. would be `[issue 717](https://github.com/w3c/webextensions/issue/717)`), but in this case the issue being discussed is the same as the header, so I did not link it.

Is 717 the issue you had in mind, or was it 716?

Rob--W:
I have applied the requested changes and updated the doc too. Thanks for the review!

_minutes/2024-11-07-wecg.md
dotproto: You're right, I meant 716. Please link-ify 😄

dotproto (approved):

Issue #353, opened 2023-02-16
Support onEnabled event when the extension from disabled state to enabled state


hanguokai:
References:
- There were many discussions about the lifecycle of extension in [this thread](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/QjuCv-LRK2c/).
- [Issue 388231](https://bugs.chromium.org/p/chromium/issues/detail?id=388231): chrome.runtime.onInstalled not run when extension is updated while its disabled

## Problem Description

Extensions that are disabled are very common. You can see what percentage of users have your extension enabled and disabled from the extension store dashboard.

At present, there are some events when an extension startup, i.e. `runtime.onInstalled` and `runtime.onStartup`. But there is no event when an extension from disabled state to enabled state, for example, users disable then re-enable an extension.

Similar to the purpose of using `onInstalled` and `onStartup` event, sometimes `onEnabled` event is also needed.

### 1. New API: `runtime.onEnabled `

```
browser.runtime.onEnabled.addListener(() => { ... })
```

### 2. Fire an event in `runtime.onInstalled`

```
browser.runtime.onInstalled.addListener(details => {
    if (details.reason == "enable") {
      ……
    }
})
```
Here, add a new reason in `runtime.OnInstalledReason`.

### 3. Allow to use `management.onEnabled` without the "management" permission

```
browser.management.onEnabled.addListener(info => { ... })
```
At present, `management.onEnabled` is required "management" permission to use it, even though you just listen yourself.

Rob--W:
I'm supportive of supporting `runtime.onInstalled` with a new reason.

tophf:
1 and 3 can work, 2 definitely won't because runtime.onInstalled clearly and unequivocally says it's about installation but none happens on re-enabling.

fastaddons:
From the practical point of view, I would like this to be a `runtime.onInstalled` event PLUS I would like to have also `previousVersion` field set to a correct value (the version of extension when it was disabled).

Since extensions are updated even when they are disabled, they are missing all `runtime.onInstalled` events with `update` reason. So this would allow to fix that issue.

tophf:
Having a misleading API is not really practical, so onInstalled should be about installations. 

>updated even when they are disabled, they are missing all runtime.onInstalled events 

[...]

PR #729, opened 2024-12-03
Add proposal for runtime.onEnabled and runtime.onExtensionLoaded events

sohailrajdev97:
Creating a proposal for #353. This is based on @justinlulejian's proposal document here:

https://docs.google.com/document/d/1Fpqg-cQRoNkQVL-hoQi5fqhVoOEdjOiGoQXWdw_yFxo/edit?tab=t.0

### DIFF
--- a/proposals/runtime_on_load_on_enabled_events.md
+++ b/proposals/runtime_on_load_on_enabled_events.md
@@ -1,0 +1,165 @@
+# Proposal: New runtime.onEnabled and runtime.onExtensionLoaded events
+
+**Summary**
+
+Creates two new extension state notification events that will help developers
+take action on the full “lifecycle” of their extensions.
+
+**Document Metadata**
+
+**Author:** jlulejian@chromium.org
+
+**Sponsoring Browser:** Google Chrome and Microsoft Edge
+
+**Contributors:** sorajdev@microsoft.com
+
+**Created:** 2024-12-03
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/353
+
+## Motivation
+
+### Objective
+
+Enable two new extension lifecycle events for the developer to add listeners for:
+
+#### chrome.runtime.onEnabled
+Enables a developer to listen to when an extension is re-enabled from a disabled
+state. There is currently no way for a developer to listen for the extension
+being disabled and then enabled.
+
+### chrome.runtime.onExtensionLoaded
+Enables a developer to listen to when an extension is added to the set of active
+extensions. This can occur from being installed, enabled, or when the browser /
+profile first starts. The listener will be given the cause so they can decide to
+take different actions based on the reason it was added to the set. There is
+currently no generic event that encapsulates all instances of extension install
+(and update), startup, and being enabled from a disabled state.
+
+#### Use Cases `chrome.runtime.onEnabled`
+A developer wants to run code once in an extension’s "lifecycle", without storing
+a flag to keep track of this.
+
+Before this API the developer would need to manually keep track of this state
+via a storage flag in persistent storage.
+
+With this new API a single listener can be registered and reduce the extension’s
+complexity.
+
+More use cases can be found [here](https://github.com/w3c/webextensions/issues/353#issuecomment-1490078217).
+
+#### Use Cases `chrome.runtime.onExtensionLoaded`
+An extension automatically does initial bootstrapping, such as creating an
+initial state used throughout the browser session, creating a WebSocket
+connection, checking for things you've missed while the browser was closed, and
+putting something in chrome.storage.session. 
+
+Before this API the developer would have to create multiple listeners for
+`chrome.runtime.onInstalled()`, `onStartup()`, and (above) `onEnabled()`.
+
+With this new API method the developer could write one method that performs the
+same action for all three, potentially with small differences based on the
+loaded cause.
+
+More use cases can be found [here](https://github.com/w3c/webextensions/issues/353#issuecomment-1582536300).
+
+### Known Consumers
+
+This is a broadly applicable API method that will be helpful to any developer
+wanting to execute code as the extension changes running states. But there
+appears to be strong interest per the [WECG discussion](https://github.com/w3c/webextensions/issues/353).
+
+## Specification
+
+### Schema
+#### `chrome.runtime.onEnabled`
+```typescript
+namespace runtime {
+  // Fired when an extension goes from being in a disabled state to an enabled
+  // state.
+  export interface onEnabled {
+    addListener(callback: () => void): void;
+    removeListener(callback: () => void): void;
+  }
+}
+```
+
+#### `chrome.runtime.onExtensionLoaded`
+```typescript
+namespace runtime {
+  export type OnLoadedReason = 'enabled' | 'installed' | 'updated' | 'profileStart';
+
+  export interface ExtensionLoadDetails {
+    // The reason that this event is being dispatched.
+    onLoadReason: OnLoadedReason;
+
+    // Indicates the previous version of the extension, which has just been
+    // updated. This is present only if 'reason' is 'updated' or ‘enabled’.
+    previousVersion?: string;
+  }
+
+  // Fired when the extension is added to the set of active extensions. This can
+  // occur in multiple scenarios: when an extension is first installed, updated
+  // to a new version, and enabled after being disabled.
+  export interface onExtensionLoaded {
+    addListener(callback: (details: ExtensionLoadDetails) => void): void;
+    removeListener(callback: (details: ExtensionLoadDetails) => void): void;
+  }
+}
+```
+
+### Behavior
+
+Described as code comments in schema description.
+
+### New Permissions
+
+No new permissions are needed as extensions can access the `browser.runtime`
+API without any permissions.
+
+### Manifest File Changes
+
+No new manifest changes are added.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+The new events expose some information that wasn’t present before. It does
+expose that an extension was (explicitly) disabled, which is something we don't
+expose today. Today, an extension cannot necessarily distinguish between being
+disabled and the user simply not opening the browser for &lt;n> amount of
+time.
+
+### Abuse Mitigations
+
+These APIs would be difficult to use in an abusive way since they do not give
+information or access to other extensions, and don’t give control into the
+extensions lifecycle – only knowledge as it transitions to different states.
+
+### Additional Security Considerations
+
+N/A.
+
+## Alternatives
+
+### Existing Workarounds
+
+There is currently `chrome.runtime.onInstalled`, and `chrome.runtime.onStartup`.
+These API have gaps in listening that are addressed by these two events. In
+short, `chrome.runtime.onEnabled`, covers the remaining status that developers
+care to know, and `chrome.runtime.onExtensionLoaded` encapsulates them all
+together with a single listener.
+
+### Open Web API
+
+These API methods are specific to extension lifetime / lifecycle and would not
+be appropriate to be in the open web.
+
+## Implementation Notes
+
+N/A.
+
+## Future Work
+
+N/A.
--- a/proposals/runtime_on_load_on_enabled_events.md
+++ b/proposals/runtime_on_load_on_enabled_events.md
@@ -87,7 +87,23 @@ namespace runtime {
 #### `chrome.runtime.onExtensionLoaded`
 ```typescript
 namespace runtime {
-  export type OnLoadedReason = 'enabled' | 'installed' | 'updated' | 'profileStart';
+  // The reason for which the event is being dispatched.
+  //
+  // 'enabled': The extension was re-enabled from a disabled state.
+  //
+  // 'installed': The extension was newly installed.
+  //
+  // 'updated': The extension was reloaded after an update.
+  //
+  // 'profileStart': The extension is being loaded during a profile startup.
+  //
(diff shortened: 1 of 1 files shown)


hanguokai:
Thanks for writing the formal proposal and sponsoring the implementation.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -0,0 +1,181 @@
...
+  // the user manually reloaded the extension).
+  export type OnLoadedReason = 'enabled' | 
+                               'installed' |
+                               'updated' |
+                               'profileStart' |
+                               'reload';
+
+  export interface ExtensionLoadDetails {
+    // The reason that this event is being dispatched.
+    onLoadReason: OnLoadedReason;
```
oliverdunk: In the original proposal (and `runtime.onEnabled`), this field is named `reason`. Was there a reason for changing it?

oliverdunk (changes requested):
Thanks so much for picking this up! I left one comment, but otherwise LGTM.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -0,0 +1,181 @@
...
+  // 'updated': The extension was reloaded after an update.
+  //
+  // 'profileStart': The extension is being loaded during a profile startup.
+  //
+  // 'reload': The extension was reloaded (e.g. via `chrome.runtime.reload() or`
+  // the user manually reloaded the extension).
+  export type OnLoadedReason = 'enabled' | 
+                               'installed' |
+                               'updated' |
+                               'profileStart' |
```
xeenon: I don’t think “profile” should be in the name here. I think this should be `started` (to be consistent with the other -ed words). Not all browsers have profiles, and many users only use a single profile, so referencing “profile” could create unnecessary confusion or imply a narrower scope than intended.

All the other reasons — `enabled`, `installed`, `updated`, and `reload` — also apply to profiles since these actions occur within a specific profile context, but they avoid explicitly mentioning profiles.

Similarly, using a name like `started` ensures consistency with the other past-tense reason names while keeping the terminology universal and clear.

proposals/runtime_on_load_on_enabled_events.md
xeenon: I agree, this should be named `reason`. All "on" prefixed properties in web extensions are event objects.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Alternatives
+
+### Existing Workarounds
+
+There is currently `chrome.runtime.onInstalled`, and `chrome.runtime.onStartup`.
+These API have gaps in listening that are addressed by these two events. In
+short, `chrome.runtime.onEnabled`, covers the remaining status that developers
+care to know, and `chrome.runtime.onExtensionLoaded` encapsulates them all
+together with a single listener.
```
xeenon, line 175: Should the other events be soft deprecated now and slated for removal in the next manifest version bump? (I think so, as they are redundant now.)

xeenon (changes requested):
See above comments.

proposals/runtime_on_load_on_enabled_events.md
oliverdunk: The impression I got from past discussions is that we see some value in having discrete events, for the cases where you do only want to listen to a particular part of the lifecycle. While I agree they are technically redundant, since we already have them I'm not sure if we should go out of our way to remove them.

proposals/runtime_on_load_on_enabled_events.md
sohailrajdev97: Makes sense. I have renamed it to started.

proposals/runtime_on_load_on_enabled_events.md
sohailrajdev97: This was a miss from my side. I have renamed this to reason.

proposals/runtime_on_load_on_enabled_events.md
sohailrajdev97: I agree that the old events are now technically redundant. If we want to soft-deprecate them, I think we can discuss that separately. What do you think?

sohailrajdev97:
@Rob--W - Can you also take a look at this one?

xeenon (approved):

proposals/runtime_on_load_on_enabled_events.md
oliverdunk: Thanks!

proposals/runtime_on_load_on_enabled_events.md
rdcronin: I don't think we should use "started", because it's ambiguous what "started" means -- without additional explanation, that could also mean the background context started, or the extension got an active process, or any number of other things.

Maybe "browserStart"?  Technically, that's not _strictly_ true, since in the case of browsers that support profiles, the browser may have already been running -- but given extensions are scoped to profiles, I think this abstraction is okay.

WDYT?

xeenon: @rdcronin `browserStart` also introduces ambiguity. It suggests the event is tied directly to the browser process starting, which as you say, isn’t _strictly_ true, especially with multiple profiles (the opposite problem of the original `profileStart`). This can mislead developers into thinking the event is narrower in scope than it is and does not apply to profiles.

My main concern with “profile” (besides the tense and camel case — we really need to be [consistent with value strings](https://github.com/w3c/webextensions/issues/563#issuecomment-2207298858)) is that this is, as far as I know, the first mention of profiles in the extension API. While the other enum values equally apply to profiles, they don’t mention profiles — and they shouldn’t. (If all enum values equally apply to the same context, explicitly mentioning the context in the values is redundant and unnecessary.) Hence, why I still think `started` is perfectly fine since it applies to the profile just like all the other reasons.

If `started` isn’t acceptable, another option would be `initialized`. It’s clear, concise, and consistent with past-tense naming conventions like `enabled`, `installed`, and `updated`. It has the connotation that it only happens once per profile load, while keeping the terminology flexible enough to apply across different scenarios, such as profile or browser-level startup.

rdcronin: @xeenon the reason I'm okay with the ambiguity of `browserStart` is that, from the extension's POV, it *is* the browser start -- no other existing windows or behavior is exposed to the extension, so it shouldn't know or care about them.  There could always be other browser instances running (multiple channels or running completely different browsers), so I don't think any extension should ever assume that, even with "browserStart", it means "there are definitely no other browser-like things running".  That said, if we wanted to avoid browserStart, how about `sessionStart`?  That has its own set of ambiguity ("session" on the web means something different), but is perhaps more clear?

Agreed we need to come to a conclusion on the value strings.  Let's plan on discussing that soon.  I don't think it's a scalable solution to say we'll only add one-word enum values : )

In the interest of parallelizing this work, WDYT about adding a specific note that this name is TBD, but otherwise allowing this to proceed (perhaps merge with an explicit TODO or similar)?  We can be sure to finalize the name before we launch the API in any browser.

xeenon: @rdcronin I'd be fine with `session-started` (or `sessionStarted`, if that is the case style going forward). That fits with the lifetime of `storage.session`. I wouldn't want to resolve to say this is TBD, as that just sets it in stone once one of the browsers ship it — API is permanent.

I think we should write an API style guide to cover things like this for easy referral to by all parties as we write API proposals. I'll file an issue for that or repose that other issue for making a style guide.

hanguokai: Personally, I like the word `init`, `startup` or `firstStart`. It is the starting point of the extension's lifecycle. We already use the name in `runtime.onStartup()`.

`browserStart` is OK for me. I don't like the word `session` , it's ambiguous for me.

hanguokai:
There is a special edge case where the two reason `enabled` and `updated` conceptually overlap. That is when the extension adds a new warning permission, then it is updated and disabled by the browser. Then the user enable the extension manually. At this point, what should `onExtensionLoaded`'s behavior be?

- A) Only fire once, and the reason is `enabled`.
- B) Only fire once, and the reason is `updated`.
- C) fire the event twice, one is `enabled` and the other is `updated`.

This special case needs to be clearly defined in the specification. Personally, I like A or B. The B is best for developers. The A is simple for implementation. The C will cause trouble for developers.

proposals/runtime_on_load_on_enabled_events.md
rdcronin: @xeenon To clarify, I'm suggesting we unblock implementation work, *not* that we ship this in any browser.  We would absolutely have it be a requirement to finalize all names before shipping anything.  But if the only remaining bit is bikeshedding the name, I think Sohail could get a start on the implementation while we discuss, rather than blocking that.  Does that sound reasonable?  (Though, maybe moot, depending on below)

@hanguokai I like `startup`, which is subtly different IMO from `started`, and as you pointed out, has precedence with `runtime.onStartup`.  I don't think `firstStart` makes sense, because it's ambiguous what "first" is related to ("firstStart" could also mean "install"), and I think "init" is generally unclear.

Anyone else have thoughts on `startup` vs `browser-start` vs `session-start`?

xeenon: I'm fine with `startup`.

rdcronin: :tada: Thanks for the suggestion, @hanguokai , and for the discussion, @xeenon !

sohailrajdev97: Yes, having `startup` makes sense as it is similar to `runtime.onStartup`. I have updated the proposal. Thank you for the inputs here @rdcronin, @xeenon and @hanguokai 😄

oliverdunk (approved):
LGTM - thanks! Seems like the discussion on enum values is resolved.

xeenon (approved):

Rob--W:
I reviewed the proposal, and it generally looks good. There is one unspecified aspect... What are the expected semantics when `incognito: "split"` is used? This is currently only implemented by Chromium browsers (e.g. Firefox does not support it).

When `"incognito": "split"` is specified in `manifest.json`, a new extension instance (background script) starts when the first Incognito window opens. When the last incognito window closes, the incognito session including extensions are torn down. In this model, my expectation is that Chrome may consider firing `onExtensionLoaded` whenever a new incognito window is opened, even if a non-incognito window has already been initialized.

The proposal currently suggests to use the `onExtensionLoaded` event to decide on whether to populate `storage.session`. That use case does not require a new event: an extension could unconditionally call `storage.session.get` at startup and check whether the value was already populated, and if not, start the initialization. Note that in Chrome, `storage.session` in incognito and non-incognito modes are separate.

oliverdunk:
> I reviewed the proposal, and it generally looks good. There is one unspecified aspect... What are the expected semantics when `incognito: "split"` is used?

This is a great callout. I think https://issues.chromium.org/41029550 which you opened is good context as well. This is an existing problem with the onInstalled event and it would be nice to agree on something as part of this work. My first thought is that firing the event probably makes sense but I'd be interested to hear any other arguments.

> The proposal currently suggests to use the `onExtensionLoaded` event to decide on whether to populate `storage.session`. That use case does not require a new event: an extension could unconditionally call `storage.session.get` at startup and check whether the value was already populated

I think we can probably just remove that example from the proposal. The other use cases are compelling and harder to do with existing workarounds.

sohailrajdev97:
> This is an existing problem with the onInstalled event and it would be nice to agree on something as part of this work. My first thought is that firing the event probably makes sense but I'd be interested to hear any other arguments.

I too agree that firing the event for split mode incognito workers makes sense. I have updated the proposal accordingly. Please take a look 😄.

FYI @Rob--W, @oliverdunk, @rdcronin, @xeenon

oliverdunk:
Looks good to me. I confirmed that Devlin was happy with the split mode behavior from a Chrome perspective.

proposals/runtime_on_load_on_enabled_events.md
xeenon: Yes, that can be a follow up discussion.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -0,0 +1,185 @@
...
+Before this API the developer would have to create multiple listeners for
+`chrome.runtime.onInstalled()`, `onStartup()`, and (above) `onEnabled()`.
+
+With this new API method the developer could write one method that performs the
+same action for all three, potentially with small differences based on the
+loaded cause.
+
+Some browsers (e.g. Chromium) support running a separate instance of the background
+script in incognito mode, by setting `incognito: "split"` in the extension manifest.
+This event will be fired for such instances separately.
```
Rob--W: The current text does not clarify that the event may fire repeatedly due to how split incognito works. I'm therefore offering an extra sentence to emphasize this. Double-check whether this is the intended design, and if not, then we need to discuss this further.

```suggestion
This event will be fired for such instances separately. The event may fire
repeatedly during the same browser session if the incognito session ends and
(re)starts, for example by closing the last incognito window and opening a new
incognito window.
```

Rob--W (approved):
I'm approving to signal that I am in favor of merging this with the suggested change below applied. If the intended behavior differs, please update the text with something else and request another review.

sohailrajdev97:
@oliverdunk - I think this is ready to merge now.

oliverdunk:
Thanks @sohailrajdev97! I'll go ahead and merge this.

Issue #565, opened 2024-03-13
Support multiple user script worlds in the user scripts API


rdcronin:
The user scripts API allows scripts to inject in a new "user script" world (in addition to the main world), which has slightly different properties than the isolated world content scripts inject in.  From early discussions (e.g. [this comment](https://github.com/w3c/webextensions/issues/279#issuecomment-1255843961) from issue #279 ), we have supported the idea of multiple user script worlds.

This issue tracks that work.

Rob--W:
I am currently working on the implementation of multiple user script worlds in Firefox, and while examining Chrome's implementation more closely, I found an issue with the currently specified/implemented behavior for `worldId`.

The current spec is at https://github.com/w3c/webextensions/blob/main/proposals/multiple_user_script_worlds.md and describes `worldId` as follows:

```
 /**
+     * If specified, specifies a specific user script world ID to execute in.
+     * Only valid if `world` is omitted or is `USER_SCRIPT`. If `worldId` is
+     * omitted, the script will execute in the default user script world.
+     * Values with leading underscores (`_`) are reserved.
+     */
+    worldId?: string;
```

This description (and the spec in general) does not specify the worldId to use for the "default user script world".

Chrome's current implementation seems to use `null` for that. This causes issues in cases where null is ambiguous: the `userScripts.update` method ought to replace some properties, but because null is interpreted as "not explicitly specified", it is not possible to change worldId back to the default world.

Example (requires starting Chrome with `--enable-features=ApiUserScriptsMultipleWorlds`):

```javascript
await chrome.userScripts.register([{
  id: "id",
  js: [{ code: "alert(0)" }],
  matches: ["https://example.com/*"],
  worldId: "x",
}]);
await chrome.userScripts.update([{
  id: "id",
  // Also weird: Chrome requires js to be set in update()
  js: [{ code: "alert(1)" }],
  worldId: null, // intent to change to default world
}]);
var scripts = await chrome.userScripts.getScripts({ ids: ["id"] });
console.log(scripts[0].worldId);
// Expected: null (default?)
// Actual: "x" (unchanged)
```

If we want to keep the default at null, it is technically possible to distinguish "property set" (=set to null) from "property not set" (= do not update) in Firefox's implementation. However, this is unusual and I would prefer a different solution that is easier to explain.

Last time we discussed the worldId format, I mentioned two options at https://github.com/w3c/webextensions/pull/560#discussion_r1517940673

[...]

PR #739, opened 2024-12-11
Clarify behavior of userScripts API

Rob--W:
This PR updates the existing userScripts API proposals to clarify some behavior.

- Use empty string instead of unspecified/null as the default worldId, following https://github.com/w3c/webextensions/issues/565#issuecomment-2473708131

- Declare "js" property as optional with remark that it is required in "register", to enable `userScripts.update()` without requiring "js".

- Expand on RegisteredUserScript validation, notably on validating matches+includeGlobs after an update.

- Update `resetWorldConfiguration()` signature to match Chrome's and Firefox's actual implementation: `worldId` is optional.

- Create a new section "World Configurations" and specify world configuration behavior more precisely. In particular, clarify fallback behavior, following https://github.com/w3c/webextensions/issues/565#issuecomment-2536646436 (in relation to https://issues.chromium.org/issues/40938749).

- Mention Firefox's optional-only "userScripts" permission design.

### DIFF
--- a/proposals/multiple_user_script_worlds.md
+++ b/proposals/multiple_user_script_worlds.md
@@ -11,7 +11,7 @@ Allow developers to configure and use multiple user script worlds in the
 
 **Sponsoring Browser:** Google Chrome
 
-**Contributors:** N/A
+**Contributors:** [Rob--W](https://github.com/Rob--W)
 
 **Created:** 2024-03-07
 
@@ -58,19 +58,24 @@ Relevant methods and types:
    export interface WorldProperties {
 +    /**
 +     * Specifies the ID of the specific user script world to update.
-+     * If not provided, updates the properties of the default user script
-+     * world.
++     * If not provided, defaults to the empty string (""), which
++     * updates the properties of the default user script world.
 +     * Values with leading underscores (`_`) are reserved.
 +     */
 +    worldId?: string;
 
      /**
-      * Specifies the world's CSP. The default is the `ISOLATED` world CSP.
+-     * Specifies the world's CSP. The default is the `ISOLATED` world CSP.
++     * Specifies the world's CSP. When not specified, falls back to the
++     * default world's `csp`. The default CSP of the default world is the
++     * `ISOLATED` world's CSP, i.e. `script-src 'self'`.
       */
      csp?: string;
 
      /**
-      * Specifies whether messaging APIs are exposed. The default is `false`.
+-     * Specifies whether messaging APIs are exposed. When not specified, falls
++     * back to the default world's `messaging`. The default is `false` for the
++     * default world.
       */
      messaging?: boolean;
    }
@@ -115,9 +120,10 @@ Relevant methods and types:
 
      /**
       * The list of ScriptSource objects defining sources of scripts to be
-      * injected into matching pages.
+      * injected into matching pages. This property must be specified for
+      * ${ref:register}
       */
-     js: ScriptSource[];
+     js?: ScriptSource[];
 
      /**
       * Specifies which pages this user script will be injected into. See
@@ -141,7 +147,8 @@ Relevant methods and types:
 +    /**
 +     * If specified, specifies a specific user script world ID to execute in.
 +     * Only valid if `world` is omitted or is `USER_SCRIPT`. If `worldId` is
-+     * omitted, the script will execute in the default user script world.
++     * omitted, the default value is an empty string ("") and the script will
++     * execute in the default user script world.
 +     * Values with leading underscores (`_`) are reserved.
 +     */
 +    worldId?: string;
@@ -164,8 +171,10 @@ Relevant methods and types:
 +    * the world with the specified ID will use the default world configuration.
 +    * Does nothing (but does not throw an error) if provided a `worldId` that
 +    * does not correspond to a current configuration.
++    * If omitted or the empty string ("") is used, it clears the configuration
++    * of the default world and all worlds without a separate configuration.
 +    */
-+   export function resetWorldConfiguration(worldId: string): Promise<void>;
++   export function resetWorldConfiguration(worldId?: string): Promise<void>;
 +
 +   /**
 +    * Returns a promise that resolves to an array of the the configurations
@@ -187,15 +196,9 @@ Worlds may be configured via `userScripts.configureWorld()` by indicating the
 given `worldId`. User scripts injected into a world with the given `worldId`
 will have the associated properties from the world configuration. If a world
 does not have a corresponding configuration, it uses the default user script
-world properties. Any existing worlds are not directly affected by
-`userScripts.configureWorld()` calls; however, the browser may revoke
-certain privileges (for instance, message calls from existing user script worlds
-may beging to fail if the extension sets `messaging` to false). This is in line
-with behavior extensions encounter when e.g. the extension is unloaded and the
-content script continues running.
-
-World configurations can be removed via the new
-`userScripts.resetWorldConfiguration()` method.
+world properties. World configurations can be removed via the new
+`userScripts.resetWorldConfiguration()` method. For additional behavioral
+notes, see the [World Configurations](#world-configurations) section.
 
 Additionally, `runtime.Port` and `runtime.MessageSender` will each be extended
 with a new, optional `userScriptWorldId` property that will be populated in the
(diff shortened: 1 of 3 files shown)


proposals/multiple_user_script_worlds.md
```diff
@@ -115,9 +120,10 @@ Relevant methods and types:
 
      /**
       * The list of ScriptSource objects defining sources of scripts to be
-      * injected into matching pages.
+      * injected into matching pages. This property must be specified for
+      * ${ref:register}
       */
-     js: ScriptSource[];
+     js?: ScriptSource[];
```
Rob--W, line 126: There is consensus among the browser vendors that this should be optional. Devlin also confirmed that on behalf of Google.

I filed an issue for Chromium to change that, at https://issues.chromium.org/issues/383712209

proposals/user-scripts-api.md
```diff
@@ -52,9 +52,12 @@ User scripting related features will be exposed in a new API namespace, tentativ
 #### Types
 
 ```
+// See RegisteredUserScript validation section, below.
 dictionary RegisteredUserScript {
   boolean? allFrames;
-  ScriptSource[] js;
+  // js is required in userScripts.register(), optional in userScripts.update().
+  // When specified, must be a non-empty array.
```
Rob--W, line 59: Side note: Chrome currently requires `js` to be a non-empty array, Firefox currently accepts an empty array.

proposals/user-scripts-api.md
```diff
@@ -157,11 +162,54 @@ As mentioned in requirement A, the user script world can communicate with differ
   - Scripts registered via [`scripting.registerContentScripts()`](https://developer.chrome.com/docs/extensions/reference/scripting/#method-registerContentScripts), following the order they were registered in. Updating a content script doesn't change its registration order.
   - Scripts registered via `userScripts.register()`, following the order they were registered in. Updating a user script doesn’t change its registration order.
 - User scripts are always persisted across sessions, since the opposite behavior would be uncommon. (We may explore providing an option to customize this in the future.)
+- Unlike regular content scripts, `matches` is allowed to be optional when `includeGlobs` is specified. A user script matches a document when its URL matches either `matches` or `includeGlobs`.
```
Rob--W, line 165: Chrome has currently not implemented this, and this is tracked at https://issues.chromium.org/issues/41483539

proposals/user-scripts-api.md
```diff
@@ -157,11 +162,54 @@ As mentioned in requirement A, the user script world can communicate with differ
...
+### RegisteredUserScript validation
+
+The `RegisteredUserScript` type is shared by `userScripts.register()` and
+`userScripts.update()`. All fields except `id` are declared as optional, to
+allow `userScripts.update()` to update individual properties.
+
+The following fields are required for a valid `RegisteredUserScript`:
+
+- `js` must be a non-empty array.
+- At least one of `matches` or `includeGlobs` must be a non-empty array.
```
Rob--W, line 178: As mentioned before, Chrome currently requires `matches`, but it should not be. When optional, "obviously" one of the two arrays should be non-empty, as mentioned before at https://issues.chromium.org/issues/41483539#comment16

proposals/user-scripts-api.md
```diff
@@ -157,11 +162,54 @@ As mentioned in requirement A, the user script world can communicate with differ
...
+
+The above requirements must be checked when `userScripts.register()` is called.
+
+When `userScripts.update()` is invoked, individual properties may be `null` or
```
Rob--W: Chrome currently resets properties when individual properties are missing, but there is consensus that it should not be changed. For details, see https://github.com/w3c/webextensions/issues/565#issuecomment-2536646436

... and this is tracked at https://issues.chromium.org/issues/40938749

Rob--W (commented):
@EmiliaPaz I've added links to Chromium bugs below in cases where Chrome's behavior does not match the specified behavior yet.

proposals/user-scripts-api.md
```diff
@@ -133,6 +136,8 @@ where
 
 In the future, if we allow multiple user script worlds (see section in Future Work below), this method can be expanded to allow for a user script world identifier to customize a single user script world.
 
+The proposal at [multiple_user_script_worlds.md](multiple_user_script_worlds.md) expands the behavior of `userScripts.configureWorld`.
```
EmiliaPaz, line 139: I would move this to `Separate USER_SCRIPT Worlds`  on the `(Potential) Future Enhancements. That way it will be easier to find the multiple user scripts world proposal on this doc

proposals/user-scripts-api.md
EmiliaPaz: Agree that matches should be optional, and with the other comments. However, it's not as clear when reading it on the updated proposal. "The following fields are required for a valid `RegisteredUserScript`" are only applicable when registering the script, right? Because updating a script doesn't require to provide at least one of `matches` or `includeGlobs`, or `js`.

What about:

```
The `RegisteredUserScript` type is shared by `userScripts.register()` and
`userScripts.update()`. All fields except `id` are declared as optional, to
allow `userScripts.update()` to update individual properties.

#### Requirements per method:
`userScripts.register()`:
- `js` must be present and a non-empty array.
- At least one of `matches` or `includeGlobs` must be a non-empty array.
 
`userScripts.update()` 
- individual properties may be `null` or omitted to leave the value 
unchanged. 
- To clear an array, an empty array can be passed. 
- The resulting script must be validated to make sure that the updated
script remains a valid script before it replaces a previous script.

#### Example
...
```

EmiliaPaz (commented):
Thanks for the updates, Rob! They make sense, just a small comment on how to better phrase one of them.

From Chrome's side, there are 3 main things missing:
1. `userScripts.register()` should require either matches or includeGlobs ([crbug.com/41483539](http://crbug.com/41483539))
2. `userScripts.update()` should not reset properties when individual properties are missing. e.g reset the user script world. ([crbug.com/40938749](http://crbug.com/40938749))
3.  `js` in RegisteredUserScript should be optional, only enforcing to be required for `userScripts.register()`. ([crbug.com/383712209](http://crbug.com/383712209))

And `userScripts.execute()`, which is still missing implementation

proposals/user-scripts-api.md
Rob--W: I have updated the proposal based on your suggestion. This is the difference:

```diff
diff --git a/proposals/user-scripts-api.md b/proposals/user-scripts-api.md
index 4dcb1b4..b584136 100644
--- a/proposals/user-scripts-api.md
+++ b/proposals/user-scripts-api.md
@@ -170,19 +170,21 @@ The `RegisteredUserScript` type is shared by `userScripts.register()` and
 `userScripts.update()`. All fields except `id` are declared as optional, to
 allow `userScripts.update()` to update individual properties.

-The following fields are required for a valid `RegisteredUserScript`:
+#### Requirements per method

-- `js` must be a non-empty array.
+`userScripts.register()`:
+
+- `js` must be present and a non-empty array.
 - At least one of `matches` or `includeGlobs` must be a non-empty array.

-The above requirements must be checked when `userScripts.register()` is called.
+`userScripts.update()`:

-When `userScripts.update()` is invoked, individual properties may be `null` or
-omitted to leave the value unchanged. To clear an array, an empty array can be
-passed. The resulting script must be validated to make sure that the updated
-script remains a valid script before it replaces a previous script.
+- Individual properties may be `null` or omitted to leave the value unchanged.
+- To clear an array, an empty array can be passed.
+- The resulting script must be validated to make sure that the updated
+  script remains a valid script before it replaces a previous script.

-Consider the following example:
+#### Example

 ```javascript
 // Valid registration:
```

proposals/user-scripts-execute-api.md
```diff
@@ -52,6 +52,10 @@ dictionary UserScriptInjection {
   target: InjectionTarget,
   // The JavaScript "world" to run the script in. The default is `USER_SCRIPT`.
   world?: ExecutionWorld,
+  // A specific user script world ID to execute in. Only valid if `world` is
+  // omitted or is `USER_SCRIPT`. If `worldId` is omitted, the default value is
+  // an empty string ("") and the script will execute in the default world.
+  worldId?: string,
```
Rob--W, line 58: @EmiliaPaz I noticed that `userScripts.execute` was missing the `worldId` property, so I decided to include it in the update here.

EmiliaPaz: Good call! Proposal was made before the introduction of multiple user script worlds. Makes sense to add `worldId`, thanks!

Rob--W:
@EmiliaPaz Is this PR ready to merge?

EmiliaPaz:
Yes, it is!

Issue #477, opened 2023-10-26
Proposal: userScripts.execute() method


newRoland:
The [user-scripts proposal](https://github.com/w3c/webextensions/blob/main/proposals/user-scripts-api.md#execute-user-scripts-one-time) mentions this as a potential future enhancement, nevertheless, I'm creating a formal proposal so that it can be discussed and tracked. 

# Context
The userScripts API addresses a significant gap that MV3 created for extensions that need to execute user-supplied JavaScript. Nonetheless, a gap persists for userscripts not driven by URLs. Many extensions activate userscripts through varied methods including gestures, keyboard shortcuts, context menus, voice, and more. 

Issue was first [mentioned](https://github.com/w3c/webextensions/issues/279#issuecomment-1369568559) by @Robbendebiene. 

# Proposal 

A method to execute user-supplied javascript on the fly. 

`browser.userScripts.execute({code: "", world, allFrames, etc})`

# Security Considerations 

The current User Scripts proposal already allows this capability. The extension can register the user script to run on all domains, and call a *wrapped* userscript on demand. For example, when the user triggers a gesture. This is inefficient and it would be harmful to users if developers had to resort to that. 

## A few extensions that would need this. 
Cr×Mouse Chrome™ Gestures
Vimium [\[1\]](https://github.com/philc/vimium/issues/4331)
AutoHotkey 
modern scroll
Script Runner Pro
Foxy Gestures
Gesturefy
smartUp Gestures
Tampermonkey, Violentmonkey (edge cases like RegExp) 
Surfingkeys [\[1\]](https://github.com/brookhong/Surfingkeys/issues/1445)
Automa [\[1\]](https://github.com/AutomaApp/automa/issues/1473)

erosman:
> A method to execute user scripts on the fly.
> browser.userScripts.execute({code: ""})

Firefox only [userScripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/userScripts) in MV2 is replaced by [scripting](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting) API in MV3 which has [scripting.executeScript()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript) method.

It is my understanding that the ability to pass a `string` to the `scripting` API is already agreed upon. Therefore, all its applicable methods should have the same ability.

newRoland:
[...]

PR #745, opened 2025-01-09
Update userScripts.execute() to receive multiple js sources

EmiliaPaz:
Updated as discussed in https://github.com/w3c/webextensions/issues/477#issuecomment-2569301401

### DIFF
--- a/proposals/user-scripts-execute-api.md
+++ b/proposals/user-scripts-execute-api.md
@@ -46,8 +46,8 @@ dictionary UserScriptInjection {
   // prior to page load, as the page may have already loaded by the time the
   // script reaches the target.
   injectImmediately?: boolean,
-  // The script source to inject into the target.
-  js: ScriptSource,
+  // The script sources to inject into the target.
+  js[]: ScriptSource,
   // Details specifying the target into which to inject the script.
   target: InjectionTarget,
   // The JavaScript "world" to run the script in. The default is `USER_SCRIPT`.


Rob--W (commented):
Can we specify what the behavior should be when a runtime error occurs while executing one of the `ScriptSource`s?

Options:
- Interrupt execution and propagate error to the caller.
- Log error and continue executing the next script.

Since `executeScript` is specified to propagate the result to the caller, it would make sense to propagate errors.

On the other hand, if registered declaratively, the expected behavior is to run them all (and uncaught errors be reported to the console).

There is something to say for either approach.

For comparison, where I checked Chrome's current [`scripting.executeScript` implementation](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/api/scripting/scripting_api.cc;l=358;drc=ddde0258ce85610aeea0e25923927777e8756dc0) will eventually be handled by the generic logic that is shared between declarative scripts and `executeScript` ([`ScriptInjection::InjectJs`](https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/script_injection.cc;l=315-321;drc=3b5e9130048e66ce39338d34f99751d0a2346e9c). This will ultimately just execute all scripts and swallow errors (in [`WebScriptExecutor::Execute`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/pausable_script_executor.cc;l=149-158;drc=3b5e9130048e66ce39338d34f99751d0a2346e9c)). Now that I see this code, I guess that it also explains why `scripting.executeScript` does not expose errors in Chrome (https://issues.chromium.org/issues/40205757).

ahyenz (approved):

xeenon (approved):

Rob--W (approved):
I'll merge this; the discussion on error handling can happen as a follow-up in the `userScripts.execute` issue itself.

PR #760, opened 2025-02-03
Specification for loading extensions in WebDriver Classic

kiaraarose:
Preview available [here](https://kiaraarose.github.io)

### DIFF
--- a/specification/webdriver-classic.bs
+++ b/specification/webdriver-classic.bs
@@ -1,0 +1,210 @@
+<pre class="metadata">
+Title: webdriver-classic
+Shortname: wecg-webdriver-classic
+Level: 1
+Group: wecg
+Status: w3c/CG-DRAFT
+URL: https://w3c.github.io/webextensions/specification/webdriver-classic.html
+Editor: Kiara Rose, Apple, kiara_rose@apple.com
+Abstract: This specification defines how web extensions should be loaded and unloaded in WebDriver Classic.
+Repository: w3c/webextensions
+</pre>
+
+  <section class="non-normative">
+    ## Introduction
+      <em>This section is non-normative.</em>
+
+      This specification is non-normative.The intention of this spec is to
+      document how extensions should be loaded/unloaded in WebDriver Classic,
+      for browser vendors who choose to implement it.
+  </section>
+  <section>
+    <h3> Endpoints</h3>
+
+    <table class=simple>
+     <tr>
+      <td>POST</td>
+      <td>/session/{<var>session id</var>}/webextension</td>
+      <td>Load WebExtension</td>
+     </tr>
+
+     <tr>
+      <td>DELETE</td>
+      <td>/session/{<var>session id</var>}/webextension/{<var>extension id</var>}</td>
+      <td>Unload WebExtension</td>
+     </tr>
+    </table>
+  </section>
+
+  <section>
+    <h3>Errors</h3>
+    <table class=simple>
+     <tr>
+      <th>Error Code
+      <th>HTTP Status
+      <th>JSON Error Code
+      <th>Description
+     </tr>
+
+     <tr>
+      <td>no such web extension
+      <td>404
+      <td><code>no such web extension</code>
+      <td>No web extension matching the given extension id was found
+          amongst the <var>session</var>&apos;s loaded extensions.
+     </tr>
+
+      <tr>
+       <td>unable to load extension
+       <td>500
+       <td><code>unable to load extension</code>
+       <td>A command to load an extension could not be satisfied.
+     </tr>
+
+     <tr>
+       <td>unable to unload extension
+       <td>500
+       <td><code>unable to unload extension</code>
+       <td>A command to unload an extension could not be satisfied.
+     </tr>
+    </table>
+  </section>
+
+  <section>
+    <h3>Commands</h3>
+    <section>
+      <h2>WebExtensions</h2>
+      <p>The WebExtensions API provides an interface that allows
+      extensions to modify and enhance the capability of the browser.
+      This section describes the interaction with WebExtensions.
+
+      <section>
+        <h3>Load WebExtension</h3>
+        <table class="simple jsoncommand">
+         <tr>
+          <th>HTTP Method</th>
+          <th>URI Template</th>
+         </tr>
+         <tr>
+          <td>POST</td>
+          <td>/session/{<var>session id</var>}/webextension</td>
+         </tr>
+        </table>
+
+        <p>Based on the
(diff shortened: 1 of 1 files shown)


specification/webdriver-classic.bs
```diff
@@ -0,0 +1,177 @@
...
+    <table class=simple>
+     <tr>
+      <td>POST</td>
+      <td>/session/{<var>session id</var>}/webextension</td>
+      <td>Load WebExtension</td>
+     </tr>
+
+     <tr>
+      <td>DELETE</td>
+      <td>/session/{<var>session id</var>}/webextension/{extension id}</td>
```
xeenon:
```suggestion
      <td>/session/{<var>session id</var>}/webextension/{<var>extension id</var>}</td>
```

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,177 @@
...
+      <section>
+        <h3>Unload WebExtension</h3>
+        <table class="simple jsoncommand">
+         <tr>
+          <th>HTTP Method</th>
+          <th>URI Template</th>
+         </tr>
+         <tr>
+          <td>DELETE</td>
+          <td>/session/{<var>session id</var>}/webextension/{extension id}</td>
```
xeenon:
```suggestion
          <td>/session/{<var>session id</var>}/webextension/{<var>extension id</var>}</td>
```

xeenon (approved):

Rob--W:
The spec proposed endpoints taking `{extension id}` in the URL.

What is the format of `extension id`? (if provided externally, how about URL encoding and such?)

How does the caller know the extension ID? That is unspecified.

xeenon:
> How does the caller know the extension ID? That is unspecified.

It is the result of the load command:

> 5. Let result be the identifier of the loaded extension.
> 6. Return success with result.

> What is the format of extension id? (if provided externally, how about URL encoding and such?)

I think it can be an opaque URL path compatible string, even just a monotonic increasing integer would do.

kiaraarose:
> I think it can be an opaque URL path compatible string, even just a monotonic increasing integer would do.

The extension identifier is the same format of what's returned in Safari. It matches`browser.runtime.id`

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,177 @@
+<pre class="metadata">
+Title: webdriver-classic
+Shortname: wecg-webdriver-classic
+Level: 1
+Group: wecg
+Status: w3c/CG-DRAFT
+URL: https://w3c.github.io/webextensions/specification/webdriver-classic.html
+Editor: Kiara Rose, Apple, kiara_rose@apple.com
+Abstract: This specification defines how web extensions should be loaded and unloaded in WebDriver Classic.
```
oliverdunk, line 9: Can we add a prominent note that this is non-normative and that not all browser vendors are expected to implement this?

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,177 @@
...
+     <tr>
+      <th>Error Code
+      <th>HTTP Status
+      <th>JSON Error Code
+      <th>Description
+     </tr>
+
+     <tr>
+      <td>no such web extension
+      <td>404
```
oliverdunk, line 51: Could we mention what error from the BiDi spec each of these refers to? It actually looks like if you click the errors in the BiDi spec, it links you to a page with an existing mapping to HTTP status codes: https://www.w3.org/TR/webdriver-bidi/#expand-a-web-extension-data-spec

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,177 @@
...
+          <th>HTTP Method</th>
+          <th>URI Template</th>
+         </tr>
+         <tr>
+          <td>POST</td>
+          <td>/session/{<var>session id</var>}/webextension</td>
+         </tr>
+        </table>
+
+        <p>The remote end steps, given <var>session</var>, <var>URL
```
oliverdunk: It would be nice to mention that this is based on the BiDi algorithm. We could maybe even say that we are mimicking the "remote end steps" from here: https://www.w3.org/TR/webdriver-bidi/#command-webExtension-install

oliverdunk (changes requested):
Generally this looks good. I've left a couple of comments following our call earlier to clarify the purpose of this document and the relationship to WebDriver BiDi :)

oliverdunk (approved):
Thanks Kiara!

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,210 @@
...
+
+          <li><p>If loading web extensions isn't supported, return
+          <a href="https://w3c.github.io/webdriver/#dfn-error">error</a>
+          with <a href="https://w3c.github.io/webdriver/#dfn-error">error code</a>
+           unsupported operation.
+          <li><p>Let <var>type hint</var> be the result of getting the
+          property "<code>type</code>" from <var>parameters</var>.
+            <ol>
+              <li type='a'><p> If <var>type hint</var> does not have the value of
+                      "path", "archivePath", or "base64", return
```
mukul-p, line 114: Based on description, it was not clear if "hint" is a required or optional field.
This param could be optional i.e. if the value is not specified; or assume a default (say path). The name "hint" also seems like an optional param.

specification/webdriver-classic.bs
```diff
@@ -0,0 +1,210 @@
...
+                      error with <a href="https://w3c.github.io/webdriver/#dfn-error">
+                      error code</a> unable to load extension.
+            </ol>
+          <li><p>If the extension fails to load, return
+          <a href="https://w3c.github.io/webdriver/#dfn-error">error</a> with
+              <a href="https://w3c.github.io/webdriver/#dfn-error">error code</a>
+              unable to load extension.
+          <li><p>Let <var>result</var> be the identifier of the loaded extension.
+
+         <li><p>Return success with <var>result</var>.
```
mukul-p, line 157: [suggestion] : some examples would be helpful.

specification/webdriver-classic.bs
kiaraarose: the parameter is required and there is no default value if one isn't supplied

zombie (approved):
LGTM

I had some nits, but realized they apply equally to the BiDi version of this proposal, so we can discuss them there.

whimboo:
@kiaraarose I would like to follow-up on this PR as landed and check back if you all would be ok when we better align the WebDriver classic integration as landed here with WebDriver BiDi in terms of wording and synchronized error codes. We would like to come up with a PR before we are going to start implementing it for Firefox.

CC @zombie, @oliverdunk, @Rob--W

PR #767, opened 2025-02-19
Update specification for loading web extensions in WebDriver Classic

kiaraarose:
As specified in the BiDi spec, the resource for "path" and "archivePath" should be held in parameters["path"] and "base64" should be in parameters["value"].

https://www.w3.org/TR/webdriver-bidi/#command-webExtension-install


xeenon (approved):

zombie (approved):
LGTM

PR #585, opened 2024-04-05
dark-mode: Add a proposal for dark mode extension icons

solomonkinard:
Proposal for https://github.com/w3c/webextensions/issues/229

PR #775, opened 2025-03-03
IconVariants: Correct spelling & address svg and empty icon_variants

solomonkinard:
> An edge case came up tho. What if an extension which has only SVG icons defined in icon_variants gets loaded into Chrome once Chrome implemented icon_variants. What would it do? Will it fall back to the legacy icons?

> https://github.com/w3c/webextensions/pull/585#issuecomment-2406633384

This PR aims to address three questions before merge:
1) How will browsers that don't support .svg handle extensions with only "any" defined as an .svg file?
2) Can "all" be set with a key other than an .svg file, such as .png?
3) Should icon_variants error if it's supplied in the manifest, but it's of the wrong type or it has the correct type but there isn't at least one valid icon_variant inside?

### DIFF
--- a/proposals/dark_mode_extension_icons.md
+++ b/proposals/dark_mode_extension_icons.md
@@ -109,7 +109,7 @@ menus.create(menusProperties);
 menus.update(id, menusProperties);
 ```
 
-A benefit of this new structure is that it's more resiliant to future changes,
+A benefit of this new structure is that it's more resilient to future changes,
 thus allowing for more keys such as density (e.g. 2dppx), purpose (e.g.
 monochrome), and etc.
 
@@ -166,10 +166,14 @@ effectively have a wild-card for `color_schemes`. For example, `["*"]` is valid.
 **Misc**
 1. If the top-level `icon_variants` key is provided, the top level `icons` key
 will be ignored.
-1. `icon_variants` will not cause an error in the event that it is invalid.
+1. `icon_variants` will not cause hard errors if no icon groups are found or
+specific icon groups have unknown properties. It may cause a hard error if icon
+groups are not of type object.
 Instead, only the faulty icon group(s) will be ignored, with an optional
 warning. Warnings are preferred over errors because they're more adaptable to
-changes in the future.
+changes in the future. If none of the icon groups are valid, agents can
+fallback to the `icons` property in `manifest.json`. If `icons` is
+missing, agents are expected not to give a hard error, but can yield a warning.
 1. `color-schemes`. Any icon group that does not contain a `color_schemes` key
 will apply to all available options, e.g. both "light" and "dark".
 1. **Group**. If only one icon group is supplied, it will be used as the


solomonkinard:
Thought:
1) AFAIK, if a browser doesn't currently have .svg support, but only "any" was supplied with it, then the extension should load without error. Perhaps a warning will suffice. Ideally there would be svg support. However, if that's not available the question is whether to a) fall back to "icons" or b) treat the extension as if it doesn't have any icons at all. I think that b) would be ideal for consistency. WDYT? Also, any thoughts on the other questions. Thanks.

oliverdunk:
If an extension specifies `icon_variants` but there are no supported variants, I could see an argument for either a warning or an error. That is something the developer can and should avoid (by specifying a lower priority variant with a commonly supported file type, like a PNG).

Without having thought too much about it, I don't think we need to fallback to the `icons` key. I'm also not against that though.

@xeenon, do you know how this is handled in Safari?

xeenon:
@oliverdunk We do record an warning / soft error if `icon_variants` is empty or the specified icons are not valid, file missing, etc.

Rob--W:
An extension without an icon is valid, so as the error handling mode I would prefer no icon over a hard error.

I wouldn't mind falling back to icons for compatibility, since that would match the behavior of the extension in browsers that don't recognize icon_variants. I don't feel too strongly about this though.

proposals/dark_mode_extension_icons.md
```diff
@@ -166,10 +166,10 @@ effectively have a wild-card for `color_schemes`. For example, `["*"]` is valid.
 **Misc**
 1. If the top-level `icon_variants` key is provided, the top level `icons` key
 will be ignored.
-1. `icon_variants` will not cause an error in the event that it is invalid.
+1. `icon_variants` will not cause an error if it's invalid.
```
carlosjeurissen:
```suggestion
1. `icon_variants` will not cause hard errors if no icon groups are found or specific icon groups have unknown properties. It may cause a hard error if icon groups are not of type object.
```

proposals/dark_mode_extension_icons.md
```diff
@@ -166,10 +166,10 @@ effectively have a wild-card for `color_schemes`. For example, `["*"]` is valid.
...
 Instead, only the faulty icon group(s) will be ignored, with an optional
 warning. Warnings are preferred over errors because they're more adaptable to
-changes in the future.
+changes in the future. If none of the icon groups are valid, should this error?
```
carlosjeurissen:
```suggestion
changes in the future. If none of the icon groups are valid, agents are expected to fallback to the `icons` property in `manifest.json`. If `icons` is missing, agents are expected to use a default error and not give a hard error.
```

This behaviour matches the current Safari implementation and reliefs extension authors from having to declare additional fallbacks in `icon_variants` for browsers which do not support SVG for example. We may want to revisit this in MV4+ when the `icons` property might be deprecated/discontinued.

proposals/dark_mode_extension_icons.md
```diff
@@ -200,7 +200,8 @@ starting with smaller sizes if available, retreating to the nearest larger size.
 1. The `"16"` is a size in `{"16": "icon.png"}` and any number can be used as a
 size, per the browser’s icon requirements. The word `"any"` can also be used in
 place of a number to represent an icon that can be shown at any size (usually
-vector images). The icon size used by the browser will be determined as follows:
+vector images). Does that mean that "any" will only have an svg and not a png?
```
carlosjeurissen:
```suggestion
vector images).
```

As currently with the `manifest.icons` key, you can already specify icons in a size not matching the size specified, I do not see a reason to not support the `any` keyword for raster images.

carlosjeurissen (changes requested):
For browsers not supporting SVG icons falling back to manifest.icons would be great for webcompat. However doing so would make the case for icon_variants as a replacement for icons weaker.

Safari currently falls back to `manifest.icons` in case `manifest.action.icon_variants` is malformed, however it skips `manifest.action.icons`. As @xeenon pointed out in https://github.com/w3c/webextensions/pull/775#discussion_r1982200238 introducing the fallback would be a behavioural change.

xeenon:
> Considering the current implementation in Safari does fall back to `icons`, and considering Google and Mozilla do not have any objections against falling back to `icons`. I believe we should fallback to `icons` for the convenience of the extension authors and the benefit for users who suffer from bad extension code.

Safari _does not_ fallback to `icons`. We have separate fallback to the containing app's icon if we can't get the icon from `icon_variants`.

proposals/dark_mode_extension_icons.md
xeenon: Safari _does not_ fallback to `icons`. We have separate fallback to the containing app's icon if we can't get the icon from `icon_variants` / `icons`. I would be fine adding fallback to `icons` but it would require a change in our current shipping behavior.

carlosjeurissen: @xeenon thanks for clarifying. You are right. In my tests, invalid `manifest.action.icon_variants` would fallback to `manifest.icons` if invalid. Which led me to the false conclusion this also happens for `manifest.icon_variants`. In that case this is up for discussion once again. For browsers not supporting SVG icons falling back to manifest.icons would be great for webcompat. However doing so would make the case for `icon_variants` as a replacement for `icons` weaker.

proposals/dark_mode_extension_icons.md
solomonkinard: Added suggested change.

proposals/dark_mode_extension_icons.md
solomonkinard: Incorporated change but made it more relaxed such that falling back to icons could be done, but isn't suggested. It's a middle of the road approach vs either requiring it or forbidding it.

proposals/dark_mode_extension_icons.md
solomonkinard: Thus, the previous question has been removed.

xeenon (approved):

carlosjeurissen (approved):

Rob--W (approved):

Issue #693, opened 2024-09-23
Add API to switch focus from `sidePanel` to main document to enable accessibility features


SpaceK33z:
**Description:**
Interacting with the extension’s side panel (`chrome.sidePanel`) shifts focus away from the main webpage, and there is no way to programmatically return focus to the main document. This limitation makes it impossible to implement some features related to accessibility.

**Context:**
Managing focus between the main document and the side panel's document within the same tab is essential for accessibility. 
With proper focus management, the user can select something from the side panel, and the side panel can simulate certain events that only work when the main document is in focus.

Currently, extensions can switch focus between tabs, but not within the main document and side panel document.

**Proposed Solution:**
Introduce a new API that allows extensions to programmatically shift focus from the side panel back to the main document.

hanguokai:
In contrast, when the side panel is opening, the focus remains on main webpage and does not move to the side panel, which makes some functions impossible. For example, a typical user requirement is to focus on the input box when the side panel opens, so that the user can input something immediately.

So I hope there is a way to solve the problem of bidirectional focus movement.

oliverdunk:
We discussed this during an in-person triage meeting at TPAC. We are aligned on the following:

- The behavior @hanguokai mentioned seems like a bug. We should look at ensuring the side panel takes focus from the page when it is first opened regardless of the mechanism used to open it.
- We are supportive of implementing a tabs.focus() method which would return focus to the web content. This would close the popup if open and move focus away from other parts of the UI such as the address bar.

hanguokai:
> We are supportive of implementing a tabs.focus() method which would return focus to the web content. This would close the popup if open and move focus away from other parts of the UI such as the address bar.

A typical case comes to mind. When opening a new tab page(by shortcut or mouse), the focus is on the address bar for user input. Now the new tab page extension would like to call tabs.focus() to move the focus into the NTP for its own searching or shortcut function. Do browsers allow this behavior?

There may be other risks of abuse.

Rob--W:
I raised that exact question and the answer is that the content area would be focused indeed. The meeting notes for the TPAC 2024 meeting will be published soon, here is the relevant excerpt:

* [rob] If the address bar was focused and you call this method, what happens?
* [devlin] It focuses the main content in the tab (document tree in Chrome, webview in Safari) - whatever was the active frame is focused.

jpc-ae:
Any movement on this?

Diego1149:
+1, this is blocking one of our launches.

[...]

PR #779, opened 2025-03-12
Proposal for sidePanel lifecycle events.

rishikramena-ms:
Creates a proposal for the extensions sidePanel API events which will cater to basic lifecycle events like onClosed and onOpened.

### DIFF
--- a/proposals/sidepanel_events.md
+++ b/proposals/sidepanel_events.md
@@ -135,7 +135,7 @@ None.
 
 ### Existing Workarounds
 
-Developers are using the `runtime.connect()` method to create a long lived channed from the side panel script and listening to the `onDisconnect` event from the service worker.
+Developers are using the `runtime.connect()` method to create a long lived channel from the side panel script and listening to the `onDisconnect` event from the service worker.
 References:
 - https://groups.google.com/a/chromium.org/g/chromium-extensions/c/o1_-Su6DkCI
 - https://medium.com/@latzikatz/chrome-side-panel-simulate-close-event-c76081f53b97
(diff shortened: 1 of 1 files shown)


akornmeier:
Related issue: #693

rishikramena-ms:
Gentle reminder for reviews on this @Rob--W, @oliverdunk, @xeenon.

proposals/sidepanel_events.md
```diff
@@ -0,0 +1,165 @@
...
+     // Reason for the side panel close trigger.
+     enum onClosedReason {
+         USER_ACTION,
+         PROGRAMMATIC,
+         NEW_SIDE_PANEL_OPENED,
+         TAB_CLOSED
+     }
+   }
+   ```
+- Would adding onHidden, onShown or onVisibilityChanged events for side panel extensions make sense too?
```
oliverdunk, line 161: As discussed in the last meeting, I am waiting until there is more discussion in the proposal about this before doing a full review.

hharshas:
Hey! I have two questions about the current proposal:
1. About the onClosed event:

>// Fired when a sidePanel hosted by an extension is triggered to close.

The spec says this event is fired when a side panel is closed. But it’s not clear who will receive this event.

Can you confirm if this event is only sent to the extension that originally opened the side panel? It would be helpful if this behavior could be made clear in the spec.

2. About the path property in the event:

>// The path to the side panel HTML file in use.
    DOMString path;

Is it necessary to include the path property in the event payload?

Since the extension itself sets the path when it opens the side panel, it can maintain the context internally and can track this information on its own. Also, there isn’t a standard pattern in other extension APIs where `path` is returned in event payloads, so this could introduce ambiguity.

hharshas:
Hi! Just a heads-up — we are currently working on a CL related to the SidePanel lifecycle.
Specifically, we have two CLs in progress: 
[onOpen](https://chromium-review.googlesource.com/c/chromium/src/+/6656092)
[onClose](https://chromium-review.googlesource.com/c/chromium/src/+/6651789)

Let me know if you see any potential suggestions. Thanks!

PR #598, opened 2024-04-18
Proposal: manifest key trial_tokens

bershanskiy:
This proposal formalizes #454 into a concise document.

PR #791, opened 2025-03-26
Add `trial_tokens` key to specification

oliverdunk:
Adds information on the `trial_tokens` key to the specification, based on the proposal [1]. The proposal is quite in depth but I have kept the specification more limited as I don't think we neccessarily need to specify each algorithm here and the motivation is not something we need in normative text.

[1] https://github.com/w3c/webextensions/pull/598

### DIFF
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -149,6 +149,10 @@ This key may be present.
 
 This key may be present.
 
+### Key `trial_tokens`
+
+The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognised, expired or malformed keys but must only show a warning and must not prevent the extension from loading.
+
 ## Reserved file names
 
 Filenames beginning with an underscore (`_`) are reserved for use by user agent.
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -151,7 +151,7 @@ This key may be present.
 
 ### Key `trial_tokens`
 
-The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognized, expired or malformed keys but must only show a warning and must not prevent the extension from loading.
+The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognized, expired or malformed keys but any such entries must not prevent the extension from loading.
 
 ## Reserved file names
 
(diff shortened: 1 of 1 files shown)


specification/index.bs
```diff
@@ -149,6 +149,10 @@ This key may be present.
 
 This key may be present.
 
+### Key `trial_tokens`
+
+The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognised, expired or malformed keys but must only show a warning and must not prevent the extension from loading.
```
rdcronin:
```suggestion
The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognized, expired or malformed keys but must only show a warning and must not prevent the extension from loading.
```

Per chat, prefer American spelling for consistency

rdcronin:
```suggestion
The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognised, expired or malformed keys but any such entries must not prevent the extension from loading.
```

Saying the user agent may show a warning but must only show a warning read a little strangely to me

rdcronin (approved):

specification/index.bs
```diff
@@ -149,6 +149,10 @@ This key may be present.
 
 This key may be present.
 
+### Key `trial_tokens`
+
+The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognized, expired or malformed keys but must only show a warning and must not prevent the extension from loading.
```
oliverdunk:
```suggestion
The <a href="#key-trial_tokens">`trial_tokens`</a> key is an optional [=list=] of strings. These may be used by the user agent to enable experimental features for the [[#extension-origin]]. The user agent may ignore or warn about any unrecognized, expired or malformed keys but any such entries must not prevent the extension from loading.
```

xeenon (approved):

mukul-p (approved):

oliverdunk:
@zombie Just adding a quick reminder here, I know you had a few conversations you wanted to have internally 😄

specification/index.bs
```diff
@@ -149,6 +149,10 @@ This key may be present.
 
 This key may be present.
 
+### Key `trial_tokens`
```
Rob--W, line 153: There is another section in the doc that enumerates all keys, under the following heading:

> The following keys must be considered valid:

Shouldn't we also add this new key there?

oliverdunk: Good spot - I've added that.

Issue #789, opened 2025-03-25
Proposal: `runtime.onInvalidated.addListener`


carlosjeurissen:
During the Berlin session, the need to figure out when a background script has been invalidated was talked about. This is especially useful for contentScripts to cleanup. There are more use-cases as well. @oleksiilevzhynskyi 

A way to address this would be a new onInvalidated event which fires when the background script gets invalidated / disconnected from the contentScript. This would not fire in case the background script gets suspended, only when it is invalidated.

```js
browser.runtime.onInvalidated.addListener(() => {
  // do cleanup in content scripts
});
```

fregante:
Related request:

- https://github.com/w3c/webextensions/issues/138

Some context and existing solution in my [`webext-events` package](https://github.com/fregante/webext-events/blob/main/source/on-context-invalidated.md)

> When an extension is disabled, restarted or updated, some contexts are left behind, specifically any content scripts in open tabs and any `*-extension://` pages open in iframes (such as the options page).
> 
> These contexts however no longer have access to the messaging and storage APIs, among others, so you get awkward "context invalidated" errors.
> 
> You can use this listener to gracefully disable the context and inform the user if necessary.

carlosjeurissen:
@fregante Thanks for bringing that up! Seems a unified event with a reasons parameter would be preferred.

xeenon:
I'm confused by this, we don't have a similar issue in Safari. The background page / worker can go away and storage and messaging are still available in extension pages and content scripts. This sounds more like a bug in the affected browsers.

fregante:
@xeenon that doesn't seem to be accurate. Safari still breaks various APIs when the extension unloads (although I haven't checked exactly if it happens on updates too)

- https://github.com/fregante/webext-events/issues/13

xeenon:
> @xeenon that doesn't seem to be accurate. Safari still breaks various APIs when the extension unloads (although I haven't checked exactly if it happens on updates too)
> 
> - https://github.com/fregante/webext-events/issues/13

I guess I don't know what "invalidated" means here. Is the extension uninstalled? Disabled?

fregante:
If I'm not mistaken, in Chrome the context is invalidated when the extension is

- uninstalled
- disabled
- updated
- reloaded

Any `chrome-extension://` tabs are closed and contexts are unloaded. Some are left behind (frames, dev tools extensions). All content scripts are left behind. All those cannot: 


- send messages to runtime
- use storage APIs

[...]

Issue #138, opened 2021-11-12
Proposal: let content scripts listen for event to detect when the extension is unloaded (for various reasons)


bershanskiy:
## Problem
Context scripts continue running even after their extension is uninstalled or disabled, and they have no convenient way to be notified of these events to start their cleanup. I contribute to multiple extensions which need this functionality to prevent old context scripts from interfering with new context scripts after extension update.

## Current workarounds

Right now, an extension content script can detect extension uninstall/shutdown by:
 - registering a persistent port and waiting for its shutdown (port can be closed by either crash, extension uninstall or periodic shutdown)
 - try to send a message from content script to background and get error that receiving end (background context) does not exist
 - listen for `chrome.runtime.onUpdateAvailable` event (extension listens for this event, asks content scripts to clean themselves up, then installs update)

Workarounds 1 and 3 can be combined into one to achieve very precise notification timings if needed.

## Proposal
Let content pages listen for a new event, `chrome.runtime.onUninstalled`. If we want to be very helpful, we can even supply a `reason` attribute, which can be `disabled` (extension was disabled), `uninstalled` (user completely uninstalled extension), or `update` (old version was uninstalled and a new version was installed in its place).

## Response to security concerns
This problem was raised multiple times in the past and received responses like "notifying content scripts about extension shutdown is dangerous because the content script can retaliate against uninstalling and do something bad". I don't understand this argument, since clearly continuing to run an orphaned content script might be even worse.

tophf:
Another workaround that doesn't depend on timing and seems reliable in my practice is to send a synchronous DOM message from the newly injected content script to the old orphaned one so it performs the cleanup, and then the new content script initializes. It's used in [Stylus extension](https://github.com/openstyles/stylus/blob/master/content/apply.js) (search for `orphan`, there's just a few lines).

bershanskiy:
> Another workaround that doesn't depend on timing and seems reliable in my practice is to send a synchronous DOM message from the newly injected content script to the old orphaned one so it performs the cleanup, and then the new content script initializes. It's used in [Stylus extension](https://github.com/openstyles/stylus/blob/master/content/apply.js) (search for orphan, there's just a few lines).

Yes, that is yet another workaround. I believe some closed-source extensions like Grammarly use it too.

bershanskiy:
Relevant Chromium bug: [414213](https://bugs.chromium.org/p/chromium/issues/detail?id=414213) created in 2014.

rdcronin:
[...]

PR #792, opened 2025-03-26
Proposal: `runtime.onInvalidated` event

carlosjeurissen:
Proposal for https://github.com/w3c/webextensions/issues/138 and https://github.com/w3c/webextensions/issues/789

It introduces a new event which fires in contentScripts and extension frames when the background scripts get invalidated. In the event is a reason property stating the reason for the invalidation.

@bershanskiy @fregante love to hear your feedback.

### DIFF
--- a/proposals/runtime_on_invalidated.md
+++ b/proposals/runtime_on_invalidated.md
@@ -11,8 +11,8 @@ out.  Remove this section once the template is filled out.
 
 `browser.runtime.onInvalidated`
 
-This event allows extension frames and content scripts to be notified when
-context get invalidated.
+This event allows extension contexts which stay alive to be notified when the
+extension context get invalidated.
 
 **Document Metadata**
 
@@ -30,15 +30,16 @@ context get invalidated.
 
 ### Objective
 
-This event allows extension frames and content scripts to be notified when
-context get invalidated.
+This event allows extension contexts which stay alive to be notified when the
+extension context get invalidated.
 
 #### Use Cases
 
-When the extension context gets invalidated, this means the content script
-or frame can no longer use the chrome APIs or communicate with the extension.
+When the extension context gets invalidated, this means extension contexts
+which stay alive can no longer use the chrome APIs or communicate with other
+extension contexts.
 
-As content script or frame, knowing when this happens is crucial for moving
+For extension contexts, knowing when this happens is crucial for moving
 forward. Alternatives can be used in this situation, like `window.open` instead
 of `browser.tabs.create`. If moving forward is no option, this event can be
 used for initiating the cleanup of any side effects caused in the frames
--- a/proposals/runtime_on_invalidated.md
+++ b/proposals/runtime_on_invalidated.md
@@ -94,7 +94,7 @@ OnInvalidatedReason:
 
 ### Behavior
 
-The event would be fired when the extension context get invalidated.
+The event will be fired when the extension context get invalidated.
 The event comes with a reason which specifies why an extension got invalidated.
 
 ### New Permissions
--- a/proposals/runtime_on_invalidated.md
+++ b/proposals/runtime_on_invalidated.md
@@ -1,0 +1,157 @@
+# Proposal: browser.runtime.onInvalidated event>
+
+** How to Use This Template **
+
+See [Proposal Process](proposal_process.md) for the detailed process on how to
+propose new APIs and use this template.  Each section includes instructions on
+what to include.  Delete the instructions for a given section once it's filled
+out.  Remove this section once the template is filled out.
+
+**Summary**
+
+`browser.runtime.onInvalidated`
+
+This event allows extension frames and content scripts to be notified when
+context get invalidated.
+
+**Document Metadata**
+
+**Author:** carlosjeurissen
+
+**Sponsoring Browser:** Chrome
+
+**Contributors:** TBD
+
+**Created:** 2025-03-26
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/138
+
+## Motivation
+
+### Objective
+
+This event allows extension frames and content scripts to be notified when
+context get invalidated.
+
+#### Use Cases
+
+When the extension context gets invalidated, this means the content script
+or frame can no longer use the chrome APIs or communicate with the extension.
+
+As content script or frame, knowing when this happens is crucial for moving
+forward. Alternatives can be used in this situation, like `window.open` instead
+of `browser.tabs.create`. If moving forward is no option, this event can be
+used for initiating the cleanup of any side effects caused in the frames
+extensions have been operating in.
+
+### Known Consumers
+
+Any extension with a content script which applies certain side effects to the
+web page.
+
+## Specification
+
+### Schema
+
+New onInvalidated event:
+```json
+{
+  "name": "onInvalidated",
+  "type": "function",
+  "description": "Fired when the extension context get invalidated.",
+  "parameters": [
+    {
+      "type": "object",
+      "name": "details",
+      "properties": {
+        "reason": {
+          "$ref": "OnInvalidatedReason",
+          "description": "The reason that this event is being dispatched."
+        }
+      }
+    }
+  ]
+}
+```
+
+Optionally gives a reason of type OnInvalidatedReason
+
+OnInvalidatedReason:
+
+```json
+{
+  "id": "OnInvalidatedReason",
+  "type": "string",
+  "enum": [
+    {"name": "uninstalled", "description": "Specifies the event reason as an uninstallation."},
+    {"name": "update", "description": "Specifies the event reason as an extension update."},
+    {"name": "reload", "description": "Specifies the event reason as an extension reloading."},
+    {"name": "disabled", "description": "Specifies the event reason as an extension disabling."}
+  ],
+  "description": "The reason that this event is being dispatched."
+}
+```
+
+### Behavior
+
+The event would be fired when the extension context get invalidated.
+The event comes with a reason which specifies why an extension got invalidated.
+
+### New Permissions
+
+No new permissions introduced.
+
+### Manifest File Changes
+
+No new manifest fields.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Document any sensitive data or personally-identifiable information the API
+exposes to the extension.
+
+### Abuse Mitigations
+
+Currently extensions could try checking if an API gets thrown with an
+invalidated error. This would allow an extension to figure out if the extension
+is invalidated. No new abuse surface is introduced.
+
+As for the returned reason. The uninstalled reason can not be abused more than
+currently `browser.runtime.setUninstalledURL`.
+
+As for the `update` reason, this can be figured out by attempting to negotiate
+between the old content script and the new content script.
+
+The `disabled` reason could provide additional information. However this seems
+to not be a big abuse surface.
+
+### Additional Security Considerations
+
+Extensions could potentially leak the onInvalidatedReason if they try hard.
+This does not seem to introduce any attack surfaces.
+
+## Alternatives
+
+### Existing Workarounds
+
+An extension could attempt to run an API which throws when an extension is
+invalidated. This is very expensive as it would require to run this API in a
+loop. In addition, the lack of an onInvalidated event leads to many
+extension developers to not deal with the situation in which the context is
+invalidated. Introducing the onInvalidated encourages developers to handle
+these situations.
+
+### Open Web API
+
+There is no viable path to turn this into an Open Web API.
+
+## Implementation Notes
+
+The event should be fired in the situation in which extension APIs would throw
+with an context invalidated error.
+
+## Future Work
+
+Not at this moment
--- a/proposals/runtime_on_invalidated.md
+++ b/proposals/runtime_on_invalidated.md
@@ -1,12 +1,5 @@
 # Proposal: browser.runtime.onInvalidated event
 
-** How to Use This Template **
-
-See [Proposal Process](proposal_process.md) for the detailed process on how to
-propose new APIs and use this template.  Each section includes instructions on
-what to include.  Delete the instructions for a given section once it's filled
-out.  Remove this section once the template is filled out.
-
 **Summary**
 
 `browser.runtime.onInvalidated`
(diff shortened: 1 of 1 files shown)


proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
...
+
+OnInvalidatedReason:
+
+```json
+{
+  "id": "OnInvalidatedReason",
+  "type": "string",
+  "enum": [
+    {"name": "uninstalled", "description": "Specifies the event reason as an uninstallation."},
+    {"name": "update", "description": "Specifies the event reason as an extension update."},
```
fregante, line 81: While it's a dev-only event, does "reload" fall into this one?

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
+# Proposal: browser.runtime.onInvalidated event>
+
+** How to Use This Template **
```
oliverdunk: Can we drop this?

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
...
+See [Proposal Process](proposal_process.md) for the detailed process on how to
+propose new APIs and use this template.  Each section includes instructions on
+what to include.  Delete the instructions for a given section once it's filled
+out.  Remove this section once the template is filled out.
+
+**Summary**
+
+`browser.runtime.onInvalidated`
+
+This event allows extension frames and contentScripts to be notified when
```
oliverdunk:
```suggestion
This event allows extension frames and content scripts to be notified when
```

...and the same throughout? I think we always use this rather than lower camelcase.

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
...
+## Motivation
+
+### Objective
+
+This event allows extension frames and contentScripts to be notified when
+background scripts get invalidated.
+
+#### Use Cases
+
+Cleanup of effects in contentScripts. After-invalidated handling of frames.
```
oliverdunk: It would be nice to elaborate on these. I know what you mean, but it would unclear to someone else why `browser.tabs` wouldn't be available after the background went away, for example :)

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
...
+  "name": "onInvalidated",
+  "type": "function",
+  "description": "Fired when the extension background scripts get invalidated.",
+  "parameters": [
+    {
+      "type": "object",
+      "name": "details",
+      "properties": {
+        "reason": {
+          "optional": true,
```
oliverdunk: Why is this optional?

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,156 @@
...
+## Specification
+
+### Schema
+
+New onInvalidated event:
+```json
+{
+  "name": "onInvalidated",
+  "type": "function",
+  "description": "Fired when the extension background scripts get invalidated.",
```
fregante: Chrome has always called it "context invalidated" so I think it makes sense to keep calling it that, rather than referring to background scripts/workers. Extensions without a background worker are also affected (e.g. content script attempts to message the sidePanel or some `chrome-extensions://**/index.html` page)

oliverdunk (changes requested):
Thanks Carlos! Left some initial comments.

proposals/runtime_on_invalidated.md
carlosjeurissen: @oliverdunk browsers may not want to expose this for either partial implementation or for security / privacy reasons. See abuse / security parts.

proposals/runtime_on_invalidated.md
carlosjeurissen: Updated. Thanks.

proposals/runtime_on_invalidated.md
carlosjeurissen: Dropped. Thanks.

proposals/runtime_on_invalidated.md
carlosjeurissen: @fregante changed to context invalidated. Thanks!

proposals/runtime_on_invalidated.md
carlosjeurissen: Added reload as reason. Love to hear from browsers on their thoughts on this.

proposals/runtime_on_invalidated.md
carlosjeurissen: Sure. Elaborated a bit more in the proposal.

proposals/runtime_on_invalidated.md
oliverdunk: Did anyone express that they would do this? I'd rather have a consistent set of properties unless there's a compelling reason otherwise.

carlosjeurissen: @oliverdunk I do not believe any browser vendor stated any kind of in favour / opposed for the reason property. So this is left purposely open. Talking of which, would Chrome be in favour / opposed for the reason property?

proposals/runtime_on_invalidated.md
oliverdunk: Hmm, it looks the same still?

carlosjeurissen: @oliverdunk try a refresh

oliverdunk: Much better! Thanks.

proposals/runtime_on_invalidated.md
oliverdunk: We're supportive. As discussed, we'd prefer to remove the `optional` modifier here.

oliverdunk (approved):
LGTM (with nit around optional field), adding Devlin to review as well.

proposals/runtime_on_invalidated.md
carlosjeurissen: @oliverdunk updated to not be optional.

proposals/runtime_on_invalidated.md
rdcronin: Doesn't look dropped yet in the latest PR

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,157 @@
...
+See [Proposal Process](proposal_process.md) for the detailed process on how to
+propose new APIs and use this template.  Each section includes instructions on
+what to include.  Delete the instructions for a given section once it's filled
+out.  Remove this section once the template is filled out.
+
+**Summary**
+
+`browser.runtime.onInvalidated`
+
+This event allows extension frames and content scripts to be notified when
```
rdcronin: extension frames should be either closed or terminated when the extension is unloaded, at least in Chrome.  Is that not the case in other browsers?  Or should we indicate this only applies for content scripts?

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,157 @@
...
+          "$ref": "OnInvalidatedReason",
+          "description": "The reason that this event is being dispatched."
+        }
+      }
+    }
+  ]
+}
+```
+
+Optionally gives a reason of type OnInvalidatedReason
```
rdcronin: as discussed yesterday, this shouldn't be optional

proposals/runtime_on_invalidated.md
rdcronin: I like it.  It's also not (necessarily) dev-only.  There are other times extensions could be reloaded, and extensions can trigger the reload themselves via runtime.reload().

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,157 @@
...
+
+OnInvalidatedReason:
+
+```json
+{
+  "id": "OnInvalidatedReason",
+  "type": "string",
+  "enum": [
+    {"name": "uninstalled", "description": "Specifies the event reason as an uninstallation."},
```
rdcronin: nit: we're mixing tenses -- "uninstalled" and "update" (I would expect "uninstall" and "update" or "uninstalled" and "updated")

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,157 @@
...
+    {"name": "reload", "description": "Specifies the event reason as an extension reloading."},
+    {"name": "disabled", "description": "Specifies the event reason as an extension disabling."}
+  ],
+  "description": "The reason that this event is being dispatched."
+}
+```
+
+### Behavior
+
+The event would be fired when the extension context get invalidated.
```
rdcronin:
```suggestion
The event will be fired when the extension context get invalidated.
```

(or "is")

rdcronin (approved):
Thank you, Carlos!  Just a few small comments on this one, otherwise, LGTM

proposals/runtime_on_invalidated.md
carlosjeurissen: Thanks!

proposals/runtime_on_invalidated.md
carlosjeurissen: Not all frames may be closed by all browsers. I can change this to extension context to be more inclusive? What do you think @rdcronin ?

proposals/runtime_on_invalidated.md
carlosjeurissen: Thanks. Resolved.

proposals/runtime_on_invalidated.md
carlosjeurissen: Fixed.

proposals/runtime_on_invalidated.md
carlosjeurissen: @rdcronin updated to use extension contexts.

proposals/runtime_on_invalidated.md
carlosjeurissen: Thanks. It is dropped now for real)

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
+# Proposal: browser.runtime.onInvalidated event
+
+**Summary**
+
+`browser.runtime.onInvalidated`
+
+This event allows extension contexts which stay alive to be notified when the
+extension context get invalidated.
```
xeenon:
```suggestion
This event notifies extension contexts that remain alive when the extension context gets invalidated.
```

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+**Created:** 2025-03-26
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/138
+
+## Motivation
+
+### Objective
+
+This event allows extension contexts which stay alive to be notified when the
+extension context get invalidated.
```
xeenon:
```suggestion
This event notifies extension contexts that remain alive when the extension context gets invalidated.
```

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Use Cases
+
+When the extension context gets invalidated, this means extension contexts
+which stay alive can no longer use the chrome APIs or communicate with other
```
xeenon:
```suggestion
which stay alive can no longer use the extension APIs or communicate with other
```

xeenon (approved):

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Manifest File Changes
+
+No new manifest fields.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Document any sensitive data or personally-identifiable information the API
```
Rob--W: This is unmodified from the template. Replace it please.

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+exposes to the extension.
+
+### Abuse Mitigations
+
+Currently extensions could try checking if an API gets thrown with an
+invalidated error. This would allow an extension to figure out if the extension
+is invalidated. No new abuse surface is introduced.
+
+As for the returned reason. The `uninstall` reason can not be abused more than
+currently `browser.runtime.setUninstalledURL`.
```
Rob--W: There is a distinction between the uninstall URL and the event: the URL is opened in a new tab, whereas the event is dispatched while the extension still has access to other pages. In theory it could modify every page to nag the user to undo the uninstallation. Is this too contrived or should we call that out?

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+## Specification
+
+### Schema
+
+New onInvalidated event:
+```json
+{
+  "name": "onInvalidated",
+  "type": "function",
+  "description": "Fired when the extension context get invalidated.",
```
Rob--W:
```suggestion
  "description": "Fired when the extension context get invalidated while the underlying document remains valid.",
```

Suggested tweak to the description. Otherwise you would also have to fire the event on page unload and browser shutdown.

This description also precludes non-content scripts. If you want to include non-content scripts I'd like to see a good justification. Regular document have the pagehide, beforeunload and unload events already. Background contexts have runtime.onSuspend.

proposals/runtime_on_invalidated.md
```diff
@@ -0,0 +1,151 @@
...
+
+**Created:** 2025-03-26
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/138
+
+## Motivation
+
+### Objective
+
+This event allows extension contexts which stay alive to be notified when the
```
Rob--W: "extension context" is very generic, while this seems to primarily be targeting content scripts. Can we restrict this to content scripts?

carlosjeurissen: @Rob--W it is generic on purpose as to cover both content scripts and iframed extension contexts on third party websites. These iframes stay alive on some browsers and require this event to handle their state. The objective states `contexts which stay alive` to exclude extension contexts which get killed on invalidation.

Updated the objective to be more explicit on this behaviour.

proposals/runtime_on_invalidated.md
carlosjeurissen: Fixed.

proposals/runtime_on_invalidated.md
carlosjeurissen: @Rob--W Good catch. Updated the PR to mention this scenario.

fregante:
What are the next steps for this PR? Seems to be approved but not merged?

carlosjeurissen:
@Rob--W can you do a final review / give approval for this proposal?

Rob--W (approved):

Issue #747, opened 2025-01-14
API to invalidate BFCache'd webpage with outdated content added by the extension


tophf:
Problem:
1. Use an extension that modifies DOM of webpages e.g. applies themes (dark mode), adds DOM stuff like toolbars, etc.
2. Navigate to another page, so the modified one goes into bfcache
3. Change an option in the extension that substantially alters the way it affects DOM
4. Go back to the bfcache'd page and observe the completely wrong content/appearance.

If we try to use a listener on `pageshow` event and retrieve the configuration, there'll be a noticeable delay before the up-to-date  configuration is applied, which is very jarring when the visual changes are substantial.

Possible solution:
```js
chrome.tabs.invalidateBFCache(tabId); // all cached documents
chrome.tabs.invalidateBFCache(tabId, {documentId: '...........'}); // just one
```

Alternatives:
```js
chrome.tabs.update(tabId, {removeBFCache: true});
chrome.tabs.update(tabId, {removeBFCache: documentId1});
chrome.tabs.update(tabId, {removeBFCache: [documentId1, document2]});
```

tophf:
A potentially [much] simpler and cleaner solution might be to allow `chrome.scripting.executeScript` on a bfcached `documentId` so it will run when the document is restored:

```js
chrome.scripting.executeScript({
  target: {tabId, documentIds: ['.........']},
  func: config => {/* apply config to the page */},
  args: [config],
});
```

Right now it ignores documentIds and just runs in the current document of the tab.

dotproto:
Note from today's meeting WECG meeting: using extension messaging to post a message to a BFCached document used to evict a document, but doesn't any more.

Rob--W:
The general consensus is to not introduce APIs to enable extensions to blanket evict bfcache'd contexts (meeting notes pending review at #750).

Instead, the idea is that the `pageshow` event is thought to be the right API, and that the problem of latency could be addressed by introducing an API to update values that can be retrieved synchronously from the content script. That topic has been discussed many times before, and #536 is one of the issues related to it.

tophf:
Indeed, config registration would help, I totally forgot about it. I vaguely remember now it was about to be implemented 5+ years ago in Chrome.

Issue #703, opened 2024-09-28
State in background scripts, synchronously available across restarts


Rob--W:
During TPAC 2024 (#659), we discussed the topic “there is not a good way to bootstrap state in the listener” in the context of background scripts. The outcome of the discussion is that we need an API that enables extension to store some state that persists across background restarts. Some limited amount of data may even be available across browser restarts. The exact details have to be figured out in a proposal.

Note: the request for state to be available at the top level has been requested before, in the context of content scripts, "globalParams" (#536). The request here is independent of that, and should not be confused with it.

tophf:
Multiple developers strongly suggested that this should be implemented in MV3 many years ago, but the counter-argument from Chromium team was that there's no synchronous API in a service worker by design.

It may be easier to implement now that we saw heisenbugs in MV3 that Chromium team couldn't fix using the asynchronous methods and they're reworking Chromium's internals to make some parts of it synchronous: https://crbug.com/334940006

Conceptually, it's fine to extend this design because such a synchronous API is a practical necessity that demonstrably improves the extension platform by removing the boilerplate code and possible fragility due to race conditions e.g. in chrome.storage when the extension writes to it from multiple content scripts and its other open pages.

Issue #501, opened 2023-12-01
Proposal: Toggleable event listeners 


newRoland:
There's no good way to make event listeners optional for the background (service worker or event page). 

An extension might have an optional feature that requires listening to chatty events. Always having these event listeners active is not ideal, especially when not all users enable that feature.  Listening to chatty events impact performance by waking the background frequently or deprive it from rest. Currently, there's no standard method to make event listeners optional. The usual approach of attaching and detaching them is impractical for the background page.

The current workaround is a hacky process that needs to be done every time the background loads. 
```js 
// on background page start, add the listener. 
browser.webNavigation.onCommitted.addListener(onListener)

// Check if the user has enabled the feature, and if not, remove the listener. 
browser.storage.local.get("someFeature").then(({someFeature}) => {
    update(someFeature) 
})

// Since the feature state can change while the background is on, need to also listen to changes
browser.storage.local.onChanged.addListener(changes => {
    if (changes.someFeature) {
         update(changes.someFeature.newValue) 
    }
})

function update(someFeature) {
     if (someFeature) {
          webNavigation.onCommitted.addListener(onListener)
     } else {
          webNavigation.onCommitted.removeListener(onListener)
     }
}
```

# Proposal 
If requiresFlag was specified,  the browser will ensure that flag is true before dispatching to the listener. 

```js
browser.webNavigation.onCommitted.addListener({
    requiresFlag: "someFlag",
    callback: onListener
})

// In options page, if user enables a feature that needs it. 
browser.runtime.setListenerFlag("someFlag", true) 

// Or to disable 
browser.runtime.setListenerFlag("someFlag", false) 
```

bershanskiy:
We had discussed the "chattiness" of some APIs a few times already, and so far did not come to a consensus. This specific use case, however, does have a decent workaround: the background context listens only for `sync` storage changes but not `local` changes, and the content script saves changes to `sync` and/or `local` storage, and explicitly sends a message to background context for all `local` changes which are relevant to the background.

Overall:
 1. Background context registers event listeners for extension update, browser startup (session start), `sync` storage update (to get notified of remote setting changes). Background does not listen to `local` changes at all.
 2. Content scripts, popups, options pages, DevTools pages read from/save to any storage, but whenever writing to `local` something relevant to background, they also send a message to background context with the change.

newRoland:
[...]

PR #793, opened 2025-03-27
Add proposal: Synchronous data at startup

Rob--W:
This proposal offers a mechanism for extensions to specify values that are synchronously available when a content script or background script executes

This covers the use cases of #536, #703, #501, #747, among others.

### DIFF
--- a/proposals/synchronous_data_at_startup.md
+++ b/proposals/synchronous_data_at_startup.md
@@ -1,0 +1,240 @@
+# Proposal: Synchronous data at startup
+
+**Summary**
+
+A mechanism for extensions to specify values that are synchronously available when a content script or background script executes.
+
+**Document Metadata**
+
+**Author:** Rob--W
+
+**Sponsoring Browser:** Mozilla
+
+**Contributors:** @oliverdunk
+
+**Created:** 2025-03-26
+
+**Related Issues:**
+
+* [Issue 536 comment](https://github.com/w3c/webextensions/issues/536#issuecomment-2200692043): `scripting.globalParams` in content scripts
+* [Issue 703](https://github.com/w3c/webextensions/issues/703): State in background scripts, synchronously available across restarts
+* Seemingly related but explicitly out of scope: [Issue 284](https://github.com/w3c/webextensions/issues/284): Main world User Script shared params
+* Resolves use case: [Issue 501](https://github.com/w3c/webextensions/issues/501): Proposal: Toggleable event listeners
+* Resolves use case: [issue 747](https://github.com/w3c/webextensions/issues/747): API to invalidate BFCache'd webpage with outdated content added by the extension
+
+## Motivation
+
+### Objective
+
+Provide a way for extensions to set and update state that is synchronously
+available at the start of script execution. The state can be set from
+privileged extension contexts, such as background scripts and options pages.
+Once the state has propagated (asynchronously), the state can *synchronously*
+be read from content scripts or other privileged contexts.
+
+This state may be session-scoped or persist across browser restarts and
+extension updates.
+
+Content scripts have the following additional limitations:
+- They can only read this shared state, and not propagate modifications.
+- They may not access state specific to privileged contexts.
+
+#### Use Cases
+
+Some use cases depend on state to be synchronously available at the start of
+script execution. This applies both to content scripts and background scripts.
+
+In the context of content scripts, initial state may be needed to run logic
+at `document_start`, before the web page has had a chance to run. Content
+scripts may also want to revalidate their state without latency after returning
+from the bfcache ([issue 747](https://github.com/w3c/webextensions/issues/747)).
+
+This concept has also been discussed before as `globalParams` in the context of
+dynamic content scripts (see https://crbug.com/1054624).
+
+In the context of background scripts, initial state may be needed for
+initialization, for example registering listeners based on user configuration
+([issue 501](https://github.com/w3c/webextensions/issues/501)).
+
+
+### Known Consumers
+
+Classes of privacy extentensions, such as NoScript
+([issue 536](https://github.com/w3c/webextensions/issues/536)).
+
+Classes of extensions that modify web pages based on user input,
+where latency in state retrieval degrades the user experience.
+The Violentmonkey author expressed a desire to use such an API in
+[issue 747](https://github.com/w3c/webextensions/issues/747).
+
+
+## Specification
+
+### Schema
+
+This proposal builds upon the existing `browser.storage` API. The `StorageArea`
+type that commonly describes the various areas within `browser.storage` is
+extended with new methods to allow synchronous access to the data.
+
+The `set` method includes a new optional `options` parameter to configure the
+persistency of the specified parameters.
+
+```
+// When execution context is allowed to read synchronously.
+               any ConfigStorageArea.getSync(keyOrKeysOrObjectWithDefaults: string | string[] | null | object)
+          string[] ConfigStorageArea.getKeysSync();
+
+// These are the same interfaces as storage.StorageArea
+   Promise<object> ConfigStorageArea.get(keyOrKeysOrObjectWithDefaults: string | string[] | null | object)
+   Promise<number> ConfigStorageArea.getBytesInUse()
+ Promise<string[]> ConfigStorageArea.getKeys()
+Promise<undefined> ConfigStorageArea.set(items: object, options?: ConfigStorageAreaSetOptions)
+Promise<undefined> ConfigStorageArea.remove(keyOrKeys: string | string[])
+Promise<undefined> ConfigStorageArea.clear()
+
+    ExtensionEvent ConfigStorageArea.onChanged;
+            number ConfigStorageArea.QUOTA_BYTES;
+
+// Read-only to content scripts; no synchronous reading from extension contexts.
+ConfigStorageArea browser.storage.contentScriptConfig;
+// Extension contexts only; can read synchronously and write.
+ConfigStorageArea browser.storage.extensionConfig;
+
+dictionary ConfigStorageAreaSetOptions {
+  persistent: boolean  // default false
+}
+```
+
+### Behavior
+
+#### ConfigStorageArea
+
+The `ConfigStorageArea` type is a composition of the existing `StorageArea`
+interface and two new methods, `getSync` and `getKeysSync`. Upon invoking the
+methods that update the storage area, the data is propagated to all processes
+where a context may exists with a need for synchronous data access.
+
+Content scripts can *synchronously* access a copy of the available data with:
+
+- `browser.storage.contentScriptConfig.getSync(...)`
+- `browser.storage.contentScriptConfig.getKeysSync(...)`
+
+These methods are unavailable to privileged extension contexts, to encourage
+extensions to use the `extensionConfig` storage area that is hidden from
+content scripts and only available to privileged extension contexts:
+
+- `browser.storage.extensionConfig.getSync(...)`
+- `browser.storage.extensionConfig.getKeysSync(...)`
+
+The asynchronous methods (inherit from the `StorageArea` interface) are only
+available to privileged extension contexts, and allow them to update the data.
+Notably, the asynchronous `get`, `getKeys` and `getBytesInUse` methods are not
+available to content scripts, despite their read-only capabilities.
+
+#### getSync
+
+The `getSync()` method has similar semantics as the `get()` method, except it
+*synchronously* returns keys and values that are known locally in the process.
+
+#### getKeysSync
+
+The `getKeysSync()` method has similar semantics as the `getKeys()` method,
+except it *synchronously* returns keys that are known locally in the process.
+
+#### set
+
+The `set()` method takes a (new) optional options object as a second parameter.
+
+The `persistent` option defaults to false, which means that the data is not
+persisted across browser restarts. When `persistent` is set to `true`, the
+specified key-value pairs are persisted across browser and extension restarts.
+
+Each key has its own associated `persistent` flag, which affect the behavior of
+the latest `set()` call, and any following `remove()` or `clear()` calls.
+
+Persisted data should be stored locally, comparable to `storage.local`.
+
+#### onChanged
+
+When a storage change is observed, the `browser.storage.onChanged`,
+`browser.storage.contentScriptConfig.onChanged` or
+`browser.storage.extensionConfig.onChanged` events are dispatched as needed.
+
+The `onChanged` event receives an object with all modified keys.
+
+TODO: Should the `oldValue` and `newValue` fields be omitted? This would enable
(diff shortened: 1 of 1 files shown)


hackademix:
Nice to see some progress to fix this very ancient problem, thank you!

I'm quite happy with this proposal, but I must signal two major problems making it unsuitable as a replacement for the current XHR work-around (one is a blocker, the second is slightly less critical):

1. __Blocker__: without fixing https://github.com/w3c/webextensions/issues/469 there would be no way to tailor this configuration for certain tabs. Possible solution: adding filters such as `includeTabIds[]` / `excludeTabIds[]` to the `ConfigStorageAreaSetOptions` type.
2. Site-specific configuration would require a very large amount of useless data to be marshaled within this storage blob, making the understandable size limit a potentially serious constrain. Similarly to n.1, a possible solution would be adding filters such as `matches[]` / `excludeMatches[]` to the `ConfigStorageAreaSetOptions` type.

carlosjeurissen:
@Rob--W Considering there seems to not be agreement on this PR yet, would it make sense to keep the issues open like https://github.com/w3c/webextensions/issues/501? And potentially open a new issue to express support labels?

xeenon:
@Rob--W Thanks for the proposal — this solves a real problem, but I strongly prefer a _single_ `storage.config` area over the proposed two-area model.

Splitting into `extensionConfig` and `contentScriptConfig` introduces unnecessary complexity and developer confusion, with little benefit. The use cases are fully covered by a single area where:
 * Extension contexts have read/write access.
 * Content scripts have read-only access.

We already enforce content script access control for `storage.session` without needing separate namespaces, so duplicating the storage area purely for access restrictions feels unnecessary.

There’s also no precedent for changes in one namespace triggering events in another, and introducing that pattern (e.g., `extensionConfig.set()` firing `contentScriptConfig.onChanged`) adds avoidable complexity.

Also, I support keeping `oldValue` and `newValue` in the `onChanged` event — browsers should be able to detect changes without needing to deserialize values first (e.g., by comparing the binary representation of old and new values).

xeenon (changes requested):
Requesting changes based on my above comment.

Rob--W:
> I strongly prefer a _single_ `storage.config` area over the proposed two-area model.
>
> (...)
> There’s also no precedent for changes in one namespace triggering events in another, and introducing that pattern (e.g., `extensionConfig.set()` firing `contentScriptConfig.onChanged`) adds avoidable complexity.

These two areas are not linked to each other; they are independent. One is meant to be used by content scripts, the other by the background script. I'll clarify.

> Splitting into `extensionConfig` and `contentScriptConfig` introduces unnecessary complexity and developer confusion, with little benefit. The use cases are fully covered by a single area where:
> 
>    * Extension contexts have read/write access.
>    * Content scripts have read-only access.
> 
> We already enforce content script access control for `storage.session` without needing separate namespaces, so duplicating the storage area purely for access restrictions feels unnecessary.

The availability to content scripts is controlled through the `storage.session.setAccessLevel` method, which is currently an all-or-nothing. To serve the use cases, we would need the ability to configure the availability per key (which is a future work already identified when `setAccessLevel` was offered in Chrome - https://crbug.com/40189208).

I see the following options:

- Current version: two separate storage areas (and quotas)
- alternative B: single area, with `setAccessLevel` that makes all or nothing available to content scripts.
- alternative C: single area, with `setAccessLevel` updated to support configuration of keys (instead of the whole area).
- alternative D: single area, with an option, e.g. `ConfigStorageAreaSetOptions.accessLevel` to specify what the access level should be.

Alternative B is better than nothing, but forces extension devs to choose sync data in content scripts XOR extension-private initialization data in background scripts.

Alternative C requires a bit more thought on the semantics of `setAccessLevel`:
- What is the scope/lifetime of `setAccessLevel`? For the global storage the level is persistent. If we introduce key-specific access levels, we need to tie the access level to the key, or otherwise deal with access levels in limbo (e.g. if `setAccessLevel` is called before `set()`, or if `remove()` was called).
- When should `setAccessLevel` be called, before or after `set()`? If called before (and there being some limbo state with access level defined for a non-existing key), `storage.onChanged` can fire, if called later, `storage.onChanged` is not guaranteed to fire.
- When a key is removed, should its access level be cleared as well?

Alternative D is a version of C without the need to worry about the following getting out of sync: access level, persistent flag, key-value pair.

ylemkimon:
I think splitting into two storage areas would both benefit security and developer's experience. Many developers don't realize  and consider that a compromised renderer may access the extension storage, and put sensitive data on it. Naming the storage "contentScriptConfig" would definitely signal developers not to.

Rob--W:
I have updated the proposal based on previous feedback and the feedback from the WECG F2F meeting in London.

There was a request from @xeenon to use one namespace instead of the proposed two. @rdcronin expressed a strong preference for keeping them separate as initially proposed. I kept them separate, with extra clarifications in the proposal.

I dropped `getSync` and `getKeysSync`, and made the `get` and `getKeys` methods synchronous instead, following the request from @rdcronin .

One consequence of this change is that `contentScriptConfig` is now also synchronously readable by privileged extension contexts (background scripts). In my initial proposal I declared the async `get` function while withholding `getSync` to discourage the use of `contentScriptConfig` by background scripts, but with only one method that is not possible. I considered alternatives, but rejected them because they felt worse:
- Omit `contentScriptConfig.get` (and `getKeys`) from privileged extension contexts. I rejected this option because it would rule out the ability for extensions to incrementally update a value (read + write), unless they use a content script to consult the current value.
- Make `contentScriptConfig.get` (and `getKeys`) return a promise (instead of the value such as seen in content scripts). I rejected this option because different return types for the same API method would be confusing.

If anyone prefers the alternatives over the current approach, I'm willing to reconsider these.

Other than the above API change, the revised proposal adds a bit more detail to leave less to the imagination:

- Explicitly list all exposed APIs for clarity.

- Emphasize independence of storage areas for clarity.

- Add expectations for flushing and method resolution behavior.

- Drop "TODO" remark about potentially excluding oldValue/newValue from onChanged event as an optimization, because the consensus is to maintain the same semantics as other storage `onChanged` events.

- Expand "Future work"

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+This proposal extends the `browser.storage` namespace with two new storage
+areas. Their interface is inspired by the `StorageArea` type, except the
+methods to look up stored keys/values are synchronous (`get` and `getKeys`).
+
+The `set` method includes a new optional `options` parameter to configure the
+persistence of the specified parameters.
+
+```
+// Methods available to extension contexts and content scripts:
+               any ConfigStorageArea.get(keyOrKeysOrObjectWithDefaults: string | string[] | null | object)
```
rdcronin, line 84: get() is pretty confusing on the existing StorageArea interface as a result of the various different kinds of inputs it can take.  Do you think it would make sense to use this opportunity to straight this out?  Especially since this is designed to be synchronous (and reads necessarily must be cheaper), it seems like we could just have a single get() call that takes a string and returns a value, and the getting-multiple-strings or getting-with-defaults cases can be handled extension side without the magic transformations we currently apply for storage.

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+
+// Methods available to extension contexts only, not content scripts:
+   Promise<number> ConfigStorageArea.getBytesInUse()
+Promise<undefined> ConfigStorageArea.set(items: object, options?: ConfigStorageAreaSetOptions)
+Promise<undefined> ConfigStorageArea.remove(keyOrKeys: string | string[])
+Promise<undefined> ConfigStorageArea.clear()
+
+            number ConfigStorageArea.QUOTA_BYTES;
+
+// In content scripts, the set/remove/clear methods are unavailable:
```
rdcronin, line 96: Is this available in user scripts?  (Conditionally, configurably, as with messaging?)

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+ConfigStorageArea browser.storage.contentScriptConfig;
+
+// Extension contexts only:
+ConfigStorageArea browser.storage.extensionConfig;
+
+dictionary ConfigStorageAreaSetOptions {
+  persistent: boolean  // default false
```
rdcronin, line 103: Having this be optionally-persistent means that we need to have a pass-through storage mechanism where we write values to disk (in yet-another new storage area) and also update shared memory, which is passed through to all processes.  That's doable, of course, but I wonder if we would get 95% of the benefit of this API with having this be in-memory / session only.

This:
* Is simpler architecturally, and thus also provides more guarantees around consistency (less chance that the in-memory store and on-disk store get out of sync -- though, of course, they may still be out of sync between renderers during the course of a storage update)
* Incentivizes using this for lightweight data and reduces the likelihood that it would grow with stale data that wasn't collected (though this is also mitigated by an aggressive storage limit)
* Anecdotally, would be used by many of these use cases, which use it for "secrets" that are better generated at runtime rather than persisted.

Downside: Extensions that _don't_ need dynamic / changing values would need to set this once per chrome session, possibly waking up unnecessarily at startup.  It also means this probably wouldn't be available synchronous for the first content scripts that run at browser run -- but that's already not guaranteed and would also not be guaranteed with a disk-backed storage (since we wouldn't block renderer load on loading that storage, at least in Chrome).

And, of course, if we decide later we definitely *do* need persistence, we could always add it in later.  WDYT?

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+// Methods available to extension contexts only, not content scripts:
+   Promise<number> ConfigStorageArea.getBytesInUse()
+Promise<undefined> ConfigStorageArea.set(items: object, options?: ConfigStorageAreaSetOptions)
+Promise<undefined> ConfigStorageArea.remove(keyOrKeys: string | string[])
+Promise<undefined> ConfigStorageArea.clear()
+
+            number ConfigStorageArea.QUOTA_BYTES;
+
+// In content scripts, the set/remove/clear methods are unavailable:
+ConfigStorageArea browser.storage.contentScriptConfig;
```
rdcronin, line 97: I find the "contentScriptConfig" naming to be both clunky and confusing (I'd expect this to be a configuration around a content script, which is much more what scripting.RegisteredContentScript is).

Preferred alternative: cache?  Maybe storage.cache for trusted contexts; storage.untrustedCache for untrusted contexts?  IMO, that leans more into the "rapid, easy-access storage that probably shouldn't be super large."  (And, see also below.)

Alternative: "lite"?  Don't love that one, though.

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+`contentScriptConfig` and `extensionConfig`. Privileged extension contexts have
+full read and write access to both areas, whereas content scripts can only
+read from `contentScriptConfig`.
+
+The `ConfigStorageArea` type is based on the `StorageArea` interface, except
+with the `get` and `getKeys` methods returning data synchronously instead of
+asynchronously.
+
+Updates to the storage area are propagated to all processes where a context may
+exists with a need for synchronous data access. This proposal specifies the
```
rdcronin, line 151:
```suggestion
exist with a need for synchronous data access. This proposal specifies the
```

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+Once flushed, saved data should immediately be readable. In particular:
+
+- Background scripts / extension service workers should be able to read from
+  `extensionConfig` at any stage of their life cycle, including startup.
+- Content scripts should be able to read from `contentScriptConfig`, even at
+  the earliest execution (`document_start`).
+
+By default, data only lasts for the duration of the browser session. The `set`
+method has a `persistent` option that extends the lifetime past browser and
+extension restarts. The data may be cleared when an extension is uninstalled.
```
rdcronin, line 163:
```suggestion
extension restarts. The data will be cleared when an extension is uninstalled.
```

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+applies to all specified keys. To have different `persistent` flags for keys,
+call the `set()` method again, with a different `persistent` option value.
+
+Persisted data should be stored locally, comparable to `storage.local`.
+
+Updates are asynchronous. The returned `Promise` may await writes to disk when
+the persistent flag is set, but does not guarantee the delivery of data to
+other processes. This ensures that non-responsive processes cannot delay the
+resolution of the `Promise`. The caller resides in the extension process and
+is guaranteed to receive the updated value when `get` or `getSync` are called.
```
rdcronin, line 202: > This ensures that non-responsive processes cannot delay the resolution of the `Promise`

I understand the motivation, but that also means that the extension won't know if it can expect that a message it sends to a renderer would be able to access the data it set.

Brainstorming: Do you think it would be reasonable to say that it's guaranteed that any message sent after the promise resolves would arrive after the storage is updated in the renderer?  I think we can make that guarantee in Chrome -- we'd update the shared memory and any messages sent would go across the same IPC channel that the shared memory update would have already gone over.

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+
+The `persistent` option defaults to false, which means that the data is not
+persisted across browser restarts. When `persistent` is set to `true`, the
+specified key-value pairs are persisted across browser and extension restarts.
+
+Each key has its own associated `persistent` flag, which affect the behavior of
+the latest `set()` call, and any following `remove()` or `clear()` calls.
+The `set` method can update multiple keys at once, and the `persistent` option
+applies to all specified keys. To have different `persistent` flags for keys,
+call the `set()` method again, with a different `persistent` option value.
```
rdcronin, line 194: see above about just not having "persistent", but if we *do* have it, we should call out the behavior of conflicting persistent values.  I assume that if a mark a key as persistent, then later set it to a non-persistent value, the persistent value is wiped.  That is, there can only be a single key, and the latest setting for persistent / non-persistent is used.

proposals/synchronous_data_at_startup.md
```diff
@@ -0,0 +1,313 @@
...
+- `browser.storage.extensionConfig.get`
+- `browser.storage.extensionConfig.getKeys`
+- `browser.storage.extensionConfig.getBytesInUse`
+- `browser.storage.extensionConfig.set`
+- `browser.storage.extensionConfig.remove`
+- `browser.storage.extensionConfig.clear`
+- `browser.storage.extensionConfig.onChanged`
+- `browser.storage.extensionConfig.QUOTA_BYTES`
+
+### Behavior
```
rdcronin, line 137: We should include a section on incognito.

I assume that, similar to the storage API, this does *not* respect incognito boundaries.

rdcronin (changes requested):
Thanks, Rob!  Excited to see this!

PR #794, opened 2025-03-27
Fixup spec deprecated list type in webdriver-classic.bs

Rob--W:
All PRs are currently failing CI because webdriver-classic fails markup validation.


oliverdunk:
Thanks for this! I just went to open one and then when resolving the merge conflicts found this PR :)

PR #788, opened 2025-03-24
Proposal for "Hybrid Permissions"

oleksiilevzhynskyi:
Hey. This proposal introduces a new concept of “hybrid_permissions” which will offer existing extensions a simplified new-install experience by prompting users with both (1) install-time (required) and (2) optional permissions and maintain current behavior for “optional_permissions” for users who already have the extension installed.

Extension manifest file example:

```json
{
  "name": "Permissions Extension",
  "hybrid_permissions": [
    "notifications"
  ]
}
```

Issue #711, opened 2024-10-17
Proposal: Enable changing the permissions for only new user


wydgetlabsadmin:
Wanted to flag a feature request where you can change the permissions for only new users so that new users don't need to opt in to a permission again when there is a permission update in the extension.

Right now when you update the permissions, it updates it for all users - disabling the extension and making it impossible to expand the feature set without using optional permissions.

fregante:
I don’t think this strictly needs new config/flags, but browsers could handle new permissions differently. 

Browsers are already able to deny access to certain hosts even if specified in the manifest. For example, the user can go visit `chrome://extensions`, open the details of an extension, and disable certain hosts. 

#### Proposal

Likewise, the browser could install an updated manifest and keep the new host disabled, using this capability. It is up to the extension to use `browser.permissions.request()` to obtain access to this new host


#### Notes

- Safari is not affected by this issue, at least for hosts
- Chrome does not have the ability to block non-host permissions this way
- Firefox does not have this capability at all AFAIK

xeenon:
Safari is fine for host permissions, but _API permissions_, like Chrome and Firefox, Safari will also disable the extension if a new non-optional one is added and it expands the capabilities of the extension into a new "bucket" — like adding injected content when none were present before, or adding `declarativeNetRequestWithHostAccess`.

brianshultz:
> _making it impossible to expand the feature set without using optional permissions._

This one resonates deeply for Tango. We’ve also been extremely careful to expand our feature set because of the current limitations. Today the choice has been:

- Introduce a new required permission and disable Tango for all users (which is too risky and disruptive for users)
- Introduce an optional permission and endure having to manage different states in perpetuity for different users who accept/reject permissions

The proposal outlined is a great balance because:
- it continues to protect both new and existing users (I think we all agree that this is the most important consideration)
- the payoff to introduce new permissions substantially increases over time (as time goes on the % of the user base that will have already accepted the permission increases)

Rob--W:
Added https://github.com/w3c/webextensions/labels/follow-up%3A%20chrome because @oliverdunk mentioned in the meeting that he'd like to follow up with Devlin to poll for interest in a solution to this.

For those interested, the meeting notes are pending review at #720 and once merged will be available at https://github.com/w3c/webextensions/blob/main/_minutes/2024-11-07-wecg.md

dotproto:
[...]

Issue #116, opened 2021-10-21
Proposal: stop adding `host_permissions` from `content_scripts`


carlosjeurissen:
Currently when using specific matches in `content_scripts`, they will be added to the list of `host_permissions` the extension requests / has access to (in some cases, only when granted).

My proposal is to stop this behavior and require devs to define this in either `host_permissions`, `optional_permissions` or `permissions`. This would align it with other APIs like the webRequest API. In addition to aligning it with other APIs, there are two more reasons for this change:

1) Some `host_permissions` you only want to apply in specific cases. So the content_scripts will only be applied when the `host_permissions` are granted. This makes sure not all `host_permissions` are required to be granted right from the install of the extension.

2) When having multiple `content_scripts` and `matches`, you will end up polluting the `host_permissions`. Say you have matches like these: "https://example.com/*something=true*", "https://example.com/*something=false*", "https://example.com/*test=true*". All of them will be added to the `host_permissions` causing conflicts with granting / removing permissions for specific domains. Easy solution here would just be to add `https://example.com/*` to the permissions.

Due to the backward compatibility issues, this may only be doable for manifest v4.

xPaw:
I just ran into this issue where I wanted to support a new domain my extension, I added it to `optional_host_permissions`, and then listed new urls in the content scripts. Upon extension update it disabled the addon in Chrome because the matches of content scripts are in fact not optional.

So either the fix here is this issue, or adding something like `optional_matches` to `content_scripts`.


https://github.com/SteamDatabase/BrowserExtension/commit/eba3f65a6981a3b96b74072c7ec7435d361828c5#diff-ffa5b716b5a57837f7929dfcca4b4dfdeb97210a7fd5a12d2f1978846d6f1743

fregante:
It helps to clarify what the current situation is exactly.

- injecting content scripts effectively gives access to the specific host and all of its contents
- in some browsers, having a host in `matches` but not in `host_permission` puts the extension in a limbo where it has access to everything, but it's not reported in `permissions.getAll()` (if I remember correctly)

What is it that an extension with `"matches": ["https://example.com/*"]` should or shouldn't be able to do?

xPaw:
I believed that hosts had to be listed in `host_permissions` for them to work in content scripts, but that clearly is not the case.

This also explains why I have a *.example.com host permission, but content scripts list specific subdomains which makes Firefox list these subdomains in permissions specifically.

fregante:
[...]

PR #798, opened 2025-03-28
Proposal "Allow to change permissions behavior on extension updates"

oleksiilevzhynskyi:
**Context**

During a recent WECG meeting (["Upgrade Permissions" section](https://docs.google.com/document/d/1vNpYgFydLZmboBMX2G8YV0T-m7YwI17evtytT48TQ6E/edit?tab=t.0#heading=h.h2tud2flqhs)), we discussed a #711 and [Proposal for "Hybrid Permissions"](https://github.com/w3c/webextensions/pull/788). The group agreed that the current behavior with "warning" is suboptimal and needs to be updated. Also, the group decided that adding a new manifest key for permissions doesn't make sense. Instead, we may add a property that will suppress permissions warning behavior and fall to "optional" behavior for new permissions during the update.

**Proposal Summary**

Add a new property, `permissions_behavior.updates`, to the manifest to specify how browsers should act when extension updates introduce new `permissions`. When set to `defer`, new permissions will behave as if specified in `optional_permissions` for existing. They are not requested or granted on update thus must be explicitly requested by the extension. The default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).

The new behavior is not applicable during install time or for permissions that do not trigger a warning.

Extension manifest file example:

```json
{
  "permissions_behavior": {
    "updates": "defer"
  }
}
```

To use the original browser's behavior, use "auto" or omit `"updates"` or `"permissions_behavior"` properties:

```json
{
  "permissions_behavior": {
    "updates": "auto"
  }
}
```

or

```json
{
  "permissions_behavior": {}
}
```

**Open questions**

- [x] Double-check the name. Is it clear?
- What should happen when the extension is already turned off or further updates are suppressed due to a permission warning? Should new property work retroactively and "restore" extension?

### DIFF
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -202,7 +202,7 @@ The proposal doesn't introduce any additional possibilities for abuse. For alrea
 
 ### Additional Security Considerations
 
-N/A
+Without `permissions_behavior_on_update="defer"`, security can be at risk by either keeping extensions on older extension versions or by accidentality removing important security/privacy extensions.
 
 ## Alternatives
 
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -2,7 +2,7 @@
 
 ** Proposal Summary**
 
-Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
+Add a new property, `permissions_behavior_on_update`, to the manifest to specify how browsers should act when extension updates introduce new `permissions`. When set to `defer`, new permissions will behave as if specified in `optional_permissions` for existing. They are not requested or granted on update thus must be explicitly requested by the extension. The default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
 
 The new behavior is not applicable during install time or for permissions that do not trigger a warning.
 
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -117,7 +117,7 @@ document.querySelector('#my-button').addEventListener('click', (event) => {
 });
 ```
 
-Once permission is granted, it behaves like a regular non-optional "permission" and cannot be removed with "permissions.remove". Once "permissions.remove" is called, then “false” will be returned, and an exception will be printed.
+Once permission is granted, it behaves like a regular non-optional "permission" and cannot be removed with `permissions.remove()`. Once `permissions.remove()` is called, then `false` will be returned, and an exception will be printed.
 
 ```js
 permissions.remove({
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -91,7 +91,7 @@ Changes listed in this proposal will enable well-established extensions to intro
 
 ### Schema
 
-The permissions' granting behavior only changes when "permissions_behavior_on_update" is set to "defer" and only for new permissions that may trigger a warning during the update. It doesn't change the behavior of permissions that don't trigger a warning or all permissions during install time.
+The permissions' granting behavior only changes when `permissions_behavior_on_update` is set to `defer` and only for new permissions that may trigger a warning during the update. It doesn't change the behavior of permissions that don't trigger a warning or all permissions during install time.
 
 When the "permissions_behavior_on_update" is absent from the manifest, the behavior remains the same as it is right now, and new permissions with a warning will turn off the extension during the update or stop updates until the user explicitly approves it (behavior varies across browsers).
 
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -170,7 +170,7 @@ Example:
 
 ### Behavior
 
-When "permissions_behavior_on_update" is set to "defer", then for existing installs (extension update flow), nothing should be requested. However, the extension should be able check and request new permission on runtime. Otherwise, permissions granting behavior remains the same.
+When `permissions_behavior_on_update` is set to `defer`, then for existing installs (extension update flow), nothing should be requested. However, the extension should be able check and request new permission on runtime. Otherwise, permissions granting behavior remains the same.
 
 ### New Permissions
 
--- a/proposals/permissions_behavior_on_update.md
+++ b/proposals/permissions_behavior_on_update.md
@@ -1,14 +1,16 @@
-# Proposal: Suppress permissions warning on update
+# Proposal: Allow to change permissions behavior on extension's update
 
 ** Proposal Summary**
 
-Add a new boolean property to the manifest that changes the default behavior of added "permissions" during the extension's update. Instead of showing a "warning," permissions shouldn't be granted; they should behave like they are optional until an extension is requested. It doesn't change behavior during install time or for permission that doesn't trigger a warning.
+Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
+
+The new behavior isn't applicable during install time or for permission that doesn't trigger a warning.
 
 Extension manifest file example:
 
 ```json
 {
-  "suppress_permissions_warning_on_update": true
+  "permissions_behavior_on_update": "defer"
 }
 ```
 
@@ -29,6 +31,10 @@ Extension manifest file example:
 
 ## Motivation
 
+### Glossary
+
+The phrase "permissions that trigger warnings" refers to permissions that require additional action from the user after the extension's update. The actual UX varies across browsers; some may silently turn off the extension, while others display an extra warning or pause further extension updates until the user explicitly permits the added permission.
+
 ### Objective
 
 Today, developers have two options for adding [new permissions that trigger warnings](https://developer.chrome.com/docs/extensions/develop/concepts/permission-warnings):
@@ -41,13 +47,13 @@ On the manifest-files level, it's a "permissions" property that represents a lis
 
 On the manifest-files level, it's an "optional_permissions" property that represents a list of permissions that the extension may request at any point, and the user agrees or declines them.
 
-**The proposal is to introduce a way to suppress permission's warning during the extension's update and, instead of turning off the extension or stopping further updates for existing users, make it behave optional where the extension has to request permission explicitly.**
+**The proposal is to introduce a way to avoid permission's warning during the extension's update and, instead of turning off the extension or stopping further updates for existing users, make it behave optional where the extension has to request permission explicitly.**
 
 **For New Installations**: The existing behavior remains, and all permissions listed in the "permissions" key will be granted. 
 
 **For Current users**: The new permissions that trigger the warning won't be granted, and the permissions warning won't be shown. Instead, the extension has to explicitly request permission within the user gesture.
 
-Important note: such behavior only applies to permissions that trigger warnings during extension updates. In the case of a new installation, all "permissions" are granted as part of existing installation flows, and no additional requests should be made.
+Important note: such behavior only applies to permissions that trigger "warnings" during extension updates. In the case of a new installation, all "permissions" are granted as part of existing installation flows, and no additional requests should be made.
 
 a) Why fallback to "optional" behavior during the update?
 
@@ -55,15 +61,15 @@ It's hard to imagine a case when the developer wants their extension to be silen
 
 b) Why not a new field proposed in ["Hybrid permissions" proposal](https://github.com/w3c/webextensions/pull/788)
 
-The intent is to avoid easting permissions keys as this may lead to ambiguity and make everyone simply avoid "permissions."
+The intent is to avoid adding new kinds of permissions as this may lead to ambiguity and make everyone simply avoid existing "permissions" key.
 
 #### User Benefits
 
-By adding "suppress_permissions_warning_on_update", we can offer the best user experience by being open and upfront about the permissions our application will require to function optimally as well as reducing user confusion with a single prompt as opposed to multiple.
+By adding "permissions_behavior_on_update", extension developers can offer the best user experience by being open and upfront about the permissions extension will require to function optimally as well as reducing user confusion with a single prompt as opposed to multiple.
 
 Also, by suppressing warnings we will reduce the fear of accidentally "turn off" extensions for the existing user and enable extension developers to actively evolve extensions by adopting new permissions as soon as they make sense.
 
-Here is an official showcase from the "Chrome for Developers” documentation that illustrates the user experience when permissions trigger a warning. Such behavior that we would like to avoid with the support for hybrid permissions
+Here is an official showcase from the "Chrome for Developers” documentation that illustrates the user experience when permissions trigger a warning. Such behavior that we would like to avoid with the proposed changes in this proposal.
 
 ![image](https://developer.chrome.com/static/docs/extensions/develop/concepts/permission-warnings/image/example-an-extension-is-6772580db51de.gif)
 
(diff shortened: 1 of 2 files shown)


xeenon (changes requested):
Thanks for the proposal — the behavior makes sense and generally aligns with what Safari does (we disable extensions on update if they cross a permission threshold, but without showing warnings).

That said, I think the property name could be more flexible. `suppress_permissions_warning_on_update` is a bit too UI/UX specific and doesn’t map well in all browsers.

At the Berlin meeting, an enum was suggested — I think something like this would scale better:

```json
"permissions_behavior_on_update": "defer"
```

This is more future-proof and avoids locking us into a UI-specific interpretation.

I plan to have a more detailed review of the proposal soon.

oleksiilevzhynskyi:
> ```json
> "permissions_behavior_on_update": "defer"
> ```

@xeenon, thank you; that's a great point! Enum `permissions_behavior_on_update` makes much more sense! 

I've updated the proposal's name and wording to account for the new name and value.

proposals/permissions_behavior_on_update.md
```diff
@@ -1,14 +1,16 @@
-# Proposal: Suppress permissions warning on update
+# Proposal: Allow to change permissions behavior on extension's update
```
carlosjeurissen:
```suggestion
# Proposal: Allow to change permissions behavior on extension updates
```

proposals/permissions_behavior_on_update.md
```diff
@@ -1,14 +1,16 @@
...
+Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
+
+The new behavior isn't applicable during install time or for permission that doesn't trigger a warning.
 
 Extension manifest file example:
 
 ```json
 {
-  "suppress_permissions_warning_on_update": true
+  "permissions_behavior_on_update": "defer"
```
carlosjeurissen: Probably disregard; however want to mention it, we could be further future-proofed by making it an object.
```json
{
  "permissions_behavior": {
    "updates": "defer",
  }
}
```

allowing other types of permission behaviours we might want to introduce in the future.

proposals/permissions_behavior_on_update.md
```diff
@@ -1,14 +1,16 @@
-# Proposal: Suppress permissions warning on update
+# Proposal: Allow to change permissions behavior on extension's update
 
 ** Proposal Summary**
 
-Add a new boolean property to the manifest that changes the default behavior of added "permissions" during the extension's update. Instead of showing a "warning," permissions shouldn't be granted; they should behave like they are optional until an extension is requested. It doesn't change behavior during install time or for permission that doesn't trigger a warning.
+Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
+
+The new behavior isn't applicable during install time or for permission that doesn't trigger a warning.
```
carlosjeurissen:
```suggestion
The new behavior is not applicable during install time or for permissions that do not trigger a warning.
```

xeenon:
> > ```json
> > "permissions_behavior_on_update": "defer"
> > ```
> 
> @xeenon, thank you; that's a great point! Enum `permissions_behavior_on_update` makes much more sense!
> 
> I've updated the proposal's name and wording to account for the new name and value.

@oleksiilevzhynskyi You will want to define the opposite value too. `"disable"`, `"prompt"`, `"warn"`, …?

proposals/permissions_behavior_on_update.md
```diff
@@ -1,14 +1,16 @@
-# Proposal: Suppress permissions warning on update
+# Proposal: Allow to change permissions behavior on extension's update
 
 ** Proposal Summary**
 
-Add a new boolean property to the manifest that changes the default behavior of added "permissions" during the extension's update. Instead of showing a "warning," permissions shouldn't be granted; they should behave like they are optional until an extension is requested. It doesn't change behavior during install time or for permission that doesn't trigger a warning.
+Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```
carlosjeurissen:
```suggestion
Add a new property, `permissions_behavior_on_update`, to the manifest to specify how browsers should act when extension updates introduce new `permissions`. When set to `defer`, new permissions will behave as if specified in `optional_permissions` for existing. They are not requested or granted on update thus must be explicitly requested by the extension. The default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Known Consumers
+
+Changes listed in this proposal will enable well-established extensions to introduce new features to users that require additional permissions without the risk of accidentally disabling the extension, but at the same time, it will help avoid confusing new users with a suboptimal permission request flow immediately after installation.
+
+## Specification
+
+### Schema
+
+The permissions' granting behavior only changes when "permissions_behavior_on_update" is set to "defer" and only for new permissions that may trigger a warning during the update. It doesn't change the behavior of permissions that don't trigger a warning or all permissions during install time.
```
carlosjeurissen:
```suggestion
The permissions' granting behavior only changes when `permissions_behavior_on_update` is set to `defer` and only for new permissions that may trigger a warning during the update. It doesn't change the behavior of permissions that don't trigger a warning or all permissions during install time.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+
```
carlosjeurissen, line 115: We should add the keywords `auto` and consider adding the keyword `prompt` (chrome behaviour)

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+When the "permissions_behavior_on_update" is absent from the manifest, the behavior remains the same as it is right now, and new permissions with a warning will turn off the extension during the update or stop updates until the user explicitly approves it (behavior varies across browsers).
+
+### Suppress permissions warning on update
+
+In case "permissions_behavior_on_update" is set to "defer", then any new permission that triggers a "warning" isn't granted but instead should be explicitly requested within a user gesture using "permissions.request"(https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request). Expect that user may decline permission.
```
carlosjeurissen:
```suggestion
In case `permissions_behavior_on_update` is set to `defer`, then any new permission that triggers a "warning" isn't granted but instead should be explicitly requested within a user gesture using "permissions.request"(https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request). Expect that user may decline permission.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+    if (granted) {
+      doSomething();
+    } else {
+      doSomethingElse();
+    }
+  });
+});
+```
+
+Once permission is granted, it behaves like a regular non-optional "permission" and cannot be removed with "permissions.remove". Once "permissions.remove" is called, then “false” will be returned, and an exception will be printed.
```
carlosjeurissen:
```suggestion
Once permission is granted, it behaves like a regular non-optional "permission" and cannot be removed with `permissions.remove()`. Once `permissions.remove()` is called, then `false` will be returned, and an exception will be printed.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+}
+```
+
+### What if extension was already turned off due to the permission "warning", but updates to a new version with "permissions_behavior_on_update" set to "defer"
+
+<TODO>
+
+### Behavior
+
+When "permissions_behavior_on_update" is set to "defer", then for existing installs (extension update flow), nothing should be requested. However, the extension should be able check and request new permission on runtime. Otherwise, permissions granting behavior remains the same.
```
carlosjeurissen:
```suggestion
When `permissions_behavior_on_update` is set to `defer`, then for existing installs (extension update flow), nothing should be requested. However, the extension should be able check and request new permission on runtime. Otherwise, permissions granting behavior remains the same.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+
+This proposal doesn't change the existing browser API or expose PII data.
+
+### Abuse Mitigations
+
+The proposal doesn't introduce any additional possibilities for abuse. For already installed extensions, newly added permissions behave like optional ones that should be explicitly requested.
+
+### Additional Security Considerations
+
+N/A
```
carlosjeurissen:
```suggestion
Without `permissions_behavior_on_update="defer"`, security can be at risk by either keeping extensions on older extension versions or by accidentality removing important security/privacy extensions.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,227 @@
...
+
+When "permissions_behavior_on_update" is set to "defer", then for existing installs (extension update flow), nothing should be requested. However, the extension should be able check and request new permission on runtime. Otherwise, permissions granting behavior remains the same.
+
+### New Permissions
+
+The proposal does not introduce any new permissions. However, it does introduce a new behavior for requesting existing permissions. 
+
+### Manifest File Changes
+
+Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to "defer," newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```
carlosjeurissen:
```suggestion
Add a new property, "permissions_behavior_on_update," to the manifest to alter the newly added "permissions" behavior during the extension's update. When set to `defer`, newly added permissions on update won't be granted by default; they will behave like optional permissions and must be explicitly requested by the extension. Also, the default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```

carlosjeurissen (commented):
Thanks for the initial proposal doc!

@xeenon would advocate for `auto` as well to have a neutral explicit default value.

proposals/permissions_behavior_on_update.md
oleksiilevzhynskyi: Good catch! Thank you!

oleksiilevzhynskyi:
Hey @xeenon @carlosjeurissen. Thank you for the comments. I updated the proposal with a new "permissions_behavior" namespace and added an "auto" value that represents the current (default) permissions behaviour during update. Please take another look.

The only part of the proposal I'm unsure of is what should happen when the extension is turned off or further updates are suppressed due to a permission warning. Would the browser be able to "restore" such an extension if the next version has "defer"? This will provide a way to rescue unintended permissions updates, but this may not be feasible from the implementation side.

proposals/permissions_behavior_on_update.md
oleksiilevzhynskyi: Make sense! I added an "auto" value to the proposal.

xeenon (approved):

oleksiilevzhynskyi:
@xeenon, thank you!

As promised, I filled in the section about rescue for extension "blocked" due to a warning. Could you check if it makes sense? https://github.com/w3c/webextensions/pull/798/commits/995df62b5daad482ef10aa330a75a87d56ab34e0

> ### [Optional feature] What if extension was already turned off due to the permission "warning", but updates to a new version with "permissions_behavior.updates" set to "defer"
> 
> We would like to provide a way to rescue an extension blocked by unintended permission updates or the accidental removal of `permissions_behavior.updates`. It should be possible to restore the extension by publishing an updated version of the extension with `permissions_behavior.updates` set to `defer`.
> 
> However, we cannot confirm the timeline of such a feature in this proposal. It remains for the browser vendor to decide if it's feasible to implement and when to implement it.

carlosjeurissen:
@oleksiilevzhynskyi alternative which would be backwards compatible is another key as discussed here:
https://github.com/w3c/webextensions/issues/711#issuecomment-2551414029

However this can be confusing during the transition period.

oleksiilevzhynskyi:
> @oleksiilevzhynskyi alternative which would be backwards compatible is another key as discussed here: [#711 (comment)](https://github.com/w3c/webextensions/issues/711#issuecomment-2551414029)
> 
> However this can be confusing during the transition period.

Agreed. The proposed change will result in a suboptimal experience for users of older browsers that do not support it. 

Your comment reminded me about the "[Hybrid permissions"](https://afhdqf.hyj/l3a/dqrstevwxyzab/defg/568) proposal. The idea is similar to `permissions_on_install` from your comment and is backward compatible. The concern about it was that the new permissions key adds **an ambiguity** and may **cause everyone to abandon the `permissions` key** at all.

Rob--W:
Do you anticipate more keys in `permissions_behavior`?

I'm wondering whether it makes more sense to specify the `update_behavior` instead. For this, I can see more general utility, e.g. specifying a desired minimum supported extension version. For example, a dev could use this if they know older versions to be harmfully broken (e.g. breaking websites or security issues), or simply if they want to discontinue support for ancient versions.

Example of revised object:

```json
  "update_behavior": {
    "new_permissions_are_optional": true
  }
```

The following is NOT a must-have, but just for illustration why it would be useful for `update_behavior` to be the top-level key:

```json
  "update_behavior": {
    "minimum_extension_version": "1.2.3",
    "new_permissions_are_optional": false
  }
```

The `minimum_extension_version` in the above example would default to "0" in Firefox (to accept any previous extension),
whereas in Chrome it could default to the new extension's version (disable older versions).

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,275 @@
+# Proposal: Allow to change permissions behavior on extension updates
+
+**Proposal Summary**
+
+Add a new property, `permissions_behavior.updates`, to the manifest to specify how browsers should act when extension updates introduce new `permissions`. When set to `defer`, new permissions will behave as if specified in `optional_permissions` for existing. They are not requested or granted on update thus must be explicitly requested by the extension. The default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```
carlosjeurissen:
```suggestion
Add a new property, `permissions_behavior.updates`, to the manifest to specify how browsers should act when extension updates introduce new `permissions` and `host_permissions`. When set to `defer`, new permissions will behave as if specified in `optional_permissions` for existing. They are not requested or granted on update thus must be explicitly requested by the extension. The default browser behavior will NOT be applied (warning, turn off, stop further updates, etc.).
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,275 @@
...
+
+The phrase "permissions that trigger warnings" refers to permissions that require additional action from the user after the extension's update. The actual UX varies across browsers; some may silently turn off the extension, while others display an extra warning or pause further extension updates until the user explicitly permits the added permission.
+
+### Objective
+
+Today, developers have two options for adding [new permissions that trigger warnings](https://developer.chrome.com/docs/extensions/develop/concepts/permission-warnings):
+
+1. Add a new permission as install-time permission which will disable the extension for 100% of current users or stop further updates until they take manual steps to grant the new permission. This is not a viable option for extension with large user install bases, as [The UI for re-approval of such permission is very unclear](https://developer.chrome.com/docs/extensions/develop/concepts/permission-warnings), and in most cases, it means a loss of most active users.
+
+On the manifest-files level, it's a "permissions" property that represents a list of permissions the user agreed upon installation.
```
carlosjeurissen:
```suggestion
On the manifest-files level, the "permissions" and "host_permissions" properties represent a list of permissions the user agreed upon installation.
```

proposals/permissions_behavior_on_update.md
```diff
@@ -0,0 +1,275 @@
...
+
+2. Add new permissions as 'optional,' which will limit access to new features until the user grants access to the new permission.  <—  This is not optimal as it creates a confusing experience for new users who reviewed and accepted permissions a few seconds earlier while installing the extension.
+
+On the manifest-files level, it's an "optional_permissions" property that represents a list of permissions that the extension may request at any point, and the user agrees or declines them.
```
carlosjeurissen:
```suggestion
On the manifest-files level, the "optional_permissions" and "optional_host_permissions" properties represent a list of permissions that the extension may request at any point, and the user agrees or declines them.
```

carlosjeurissen (commented):
Added notes on `host_permissions`

xeenon (approved):

carlosjeurissen:
@Rob--W one argument in favor of `permissions_behavior` instead of `update_behavior` is mentioned in:
https://github.com/w3c/webextensions/issues/116#issuecomment-3032370321

Having the ability to change how host permissions should be prompted / if content script matches should be added to `host_permissions` or not.

carlosjeurissen:
@oleksiilevzhynskyi could you follow up on some of the suggested changes?

@oliverdunk do you have some thoughts on this PR from the Chrome side?

oleksiilevzhynskyi (commented):
(deleted)

oleksiilevzhynskyi:
Hey. Here is a summary of the discussion and the current status

1. `permissions_behavior` object provides a room for extensibility without the need to bump the manifest. The current proposal adds `updates`. Additionally, [#116 (comment)](https://github.com/w3c/webextensions/issues/116#issuecomment-3032370321) proposes the possibility to add `"content_scripts": "suppress"`.

```json
"permissions_behavior": {
   "updates": "defer",
   "content_scripts": "suppress" // Not a part of this proposal
}
```

2. **[Important]** `permissions_behavior` isn't backward compatible, as it will only work starting from the browser version that supports it; older versions will simply ignore it. 

Next steps:
- on upcoming "WebExtensions CG" agreed on"backward compatibility" concern (2). 
- browser vendors to review and agree on the proposal

carlosjeurissen:
@oleksiilevzhynskyi to give a clear alternative for the discussion, a backwards compatible way of dealing with the permissions could be a new `permissions_on_install`.

`permissions_on_install` would be an array of permissions just like other permission containers like `permissions` and `optional_permissions`.

Permissions in `permissions_on_install` will be prompted to the user on installation similar to the current browser behaviour for `permissions`. With the difference being during an update. If an extension updates with additional permissions in `permissions_on_install`, new permissions will not be prompted to existing users. Only to new users. This covers the initial use case of this proposal.

In addition, if `permissions_on_install` is present and understood by the browser. The `prompting` behaviour of the `permissions` key and match patterns in `content_scripts` will be suppressed. This would allow for maximum backwards compatibility while opening up for the new behaviour for new browser versions.

Since the presence of `permissions_on_install` would downgrade `permissions` to being optional. Eventually the `optional_permissions` key could be dropped and all permissions not in `permissions_on_install` be treated as optional.

carlosjeurissen:
Created a [proposal doc 883](https://github.com/w3c/webextensions/pull/883) for `initial_permissions`, which is similar to the `permissions_on_install` proposed above. This proposal has the benefit of allowing full backwards compatibility right from the start.

Issue #790, opened 2025-03-26
Proposal: `browser.permissions.canAccess()`


carlosjeurissen:
### Background
Currently `browser.permissions.contains()` is not sufficient for determining if an extension can act on a certain origin. An extension could have access to `<all_urls>` while it can not act on `https://chromewebstore.google.com/`. `browser.permissions.contains()` with origin `https://chromewebstore.google.com/*` would simply return true.

Other examples include specific new tab pages and `https://edgeservices.bing.com/chat`, see https://github.com/microsoft/MicrosoftEdge-Extensions/issues/156

Given other proposals like `permissions.blocked` https://github.com/w3c/webextensions/issues/701 and `disallow_host_permissions` https://github.com/w3c/webextensions/issues/123, a way to know as extension if it can act on a specific origin would be very helpful.

### Proposal
Add a new method under browser.permissions which returns a boolean whether or not an extension can act upon it. If possible, it can be helpful for this to be synchronous.

The parameter passed is an exact origin. No match patterns allowed as this would run in similar issues as permission.contains().

Example:
```js
const allowedOnChromeWebStore = await browser.permissions.canAccess('https://chromewebstore.google.com');
```

fregante:
Chrome calls this "scriptable":

- https://chromium.googlesource.com/chromium/src/+/d774155ed89969a8181932f9f2db8ebc179f5b5e/chrome/common/extensions/chrome_extensions_client.cc#249

Firefox "restricted"

- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts#restricted_domains

I have my own helper that uses a static list of URLs extracted from the source code of browsers. It can definitely be a sync check

- https://github.com/fregante/webext-content-scripts#isscriptableurlurl

xeenon:
I’m not too keen on having two methods with contains. I’d lean toward something like `accessible()` or `isAccessible()` — clearer intent and avoids overloading the meaning of “contains.”

carlosjeurissen:
@xeenon Good point! Sounds good. Updated to reflect this.  Are you supportive of an API like this? If any browser is willing to sponsor I can work on a proposal PR.

xeenon:
@carlosjeurissen Yes, we can sponsor the proposal.

carlosjeurissen:
@xeenon Wonderful! Thank you.

[...]

PR #799, opened 2025-03-28
Add `browser.permissions.canAccess()` proposal

carlosjeurissen:
Related issue: https://github.com/w3c/webextensions/issues/790

### DIFF
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -39,7 +39,7 @@ as without host permissions extension APIs may not expose URLs.
 Extensions which currently maintain guess work on what hosts are restricted.
 For example: https://github.com/fregante/webext-content-scripts#isscriptableurlurl
 And extensions which expect `permissions.contains()` to be reliable while it
-may not be reliable in all situations.
+may not be reliable in all situations. See workarounds.
 
 ## Specification
 
@@ -56,9 +56,9 @@ dictionary AccessQuery {
 }
 
 dictionary AccessResult {
-  // extension context can executeScript on host
-  scriptingAccess: boolean,
-  // extension context has temporary host access allowing fetch requests
+  // extension context can executeScript on host. Not returned when only a url is passed as AccessQuery
+  scriptingAccess?: boolean,
+  // extension context has host access allowing fetch requests
   hostAccess: boolean,
 }
 
@@ -109,10 +109,14 @@ No new attack vectors are created.
 
 ### Existing Workarounds
 
-Currently extensions can call the `browser.permissions.contains()` API. This
+Currently extensions can call the `permissions.contains()` API. This
 would be insufficient to determine access to potential restricted hosts
 as a result of enterprise policies or for security reasons. For example,
-some browsers block access to extension store hosts.
+some browsers block access to extension store hosts. `permissions.contains()`
+in this scenario would give false positives.
+
+In addition, with `activeTab`, `permissions.contains()` might return false while
+the extension context has `hostAccess`.
 
 Another workaround is actually calling APIs which could throw if the extension
 has no access. This is slower and could lead to side-effects. In general it is
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -77,7 +77,7 @@ Promise<AccessResult[]> browser.permissions.canAccess(
 
 ### Behavior
 
-The accessOnce can be set by browsers which in case of the `activeTab`
+The `accessOnce` can be set by browsers which in case of the `activeTab`
 permission grant host access once.
 
 ### New Permissions
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -125,7 +125,7 @@ web thus does not make sense.
 
 ## Implementation Notes
 
-Most browsers currently already have the concept of canAccess internally.
+Most browsers currently already have the concept of `canAccess()` internally.
 This can be leveraged during implementation.
 
 ## Future Work
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -38,7 +38,7 @@ as without host permissions extension APIs may not expose URLs.
 
 Extensions which currently maintain guess work on what hosts are restricted.
 For example: https://github.com/fregante/webext-content-scripts#isscriptableurlurl
-And extensions which expect permissions.contains to be reliable while it
+And extensions which expect `permissions.contains()` to be reliable while it
 may not be reliable in all situations.
 
 ## Specification
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -56,10 +56,10 @@ dictionary AccessQuery {
 }
 
 dictionary AccessResult {
-  // The current extension access state of the given host
-  canAccess: boolean,
-  // Can inject scripts only
-  injectOnly: boolean,
+  // extension context can executeScript on host
+  scriptingAccess: boolean,
+  // extension context has temporary host access allowing fetch requests
+  hostAccess: boolean,
 }
 
 ```
@@ -77,8 +77,8 @@ Promise<AccessResult[]> browser.permissions.canAccess(
 
 ### Behavior
 
-The `accessOnce` can be set by browsers which in case of the `activeTab`
-permission grant host access once.
+In the case of `activeTab`, `scriptingAccess` is set true. If the browser
+grants temporary host access with `activeTab`, `hostAccess` is set true as well.
 
 ### New Permissions
 
@@ -92,8 +92,8 @@ No new manifest fields.
 
 ### Exposed Sensitive Data
 
-Document any sensitive data or personally-identifiable information the API
-exposes to the extension.
+No sensitive data is exposed in this API. The information given to extensions
+in this API is only what access the extension has, not granting actual access.
 
 ### Abuse Mitigations
 
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -120,7 +120,7 @@ better to prevent throwing errors when they can be circumvented.
 
 ### Open Web API
 
-Host access generally is unique to webExtensions. Having such API on the open
+Host access generally is unique to extensions. Having such API on the open
 web thus does not make sense.
 
 ## Implementation Notes
--- a/proposals/permissions-canAccess.md
+++ b/proposals/permissions-canAccess.md
@@ -58,8 +58,8 @@ dictionary AccessQuery {
 dictionary AccessResult {
   // The current extension access state of the given host
   canAccess: boolean,
-  // The host permission can only be used once
-  accessOnce: boolean,
+  // Can inject scripts only
+  injectOnly: boolean,
 }
 
 ```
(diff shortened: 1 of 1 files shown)


proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+before calling a specific API.
+
+Figure out access without having to know the URL of a specific tab. Useful
+as without host permissions extension APIs may not expose URLs.
+
+### Known Consumers
+
+Extensions which currently maintain guess work on what hosts are restricted.
+For example: https://github.com/fregante/webext-content-scripts#isscriptableurlurl
+And extensions which expect permissions.contains to be reliable while it
```
xeenon:
```suggestion
And extensions which expect `permissions.contains()` to be reliable while it
```

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+```ts
+Promise<AccessResult[]> browser.permissions.canAccess(
+  query: AccessQuery
+)
+```
+
+### Behavior
+
+The accessOnce can be set by browsers which in case of the `activeTab`
+permission grant host access once.
```
xeenon: This is the only mention of `accessOnce`. Should it be in `AccessResult`?
```suggestion
The `accessOnce` can be set by browsers which in case of the `activeTab`
permission grant host access once.
```

Tying this closer to `activeTab` might be better, since it is not only once, you could inject, etc multiple times and it also carries over if you navigate to the same host in the same tab. Maybe:
```suggestion
The `activeTabOnly` property can be set if host access was granted by the `activeTab`
permission and an active user gesture in the specified tab. It is only `true` if the host permission
 hasn't been broadly granted to the whole extension.
```
Getting a result for this would require `tabId` or `documentId` in the `AccessQuery` to be supplied.

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+dictionary AccessQuery {
+  url: string,
+  documentId: string,
+  tabId: number,
+  frameId: number,
+}
+
+dictionary AccessResult {
+  // The current extension access state of the given host
+  canAccess: boolean,
```
xeenon: Repeating `canAccess` in the result from the method name is confusing to me, maybe just `allowed`?
```suggestion
  allowed: boolean,
```

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+  // Can inject scripts only
+  injectOnly: boolean,
```
xeenon: I’m not sure what `injectOnly` is intended to represent. Host access is typically binary — either the extension has access to the host or it doesn’t. When would this field be `true`? What APIs would be allowed or disallowed in that case?

Unless there’s a concrete and well-defined use case for this, I suggest removing `injectOnly` to keep the API simple and predictable.

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+### Manifest File Changes
+
+No new manifest fields.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Document any sensitive data or personally-identifiable information the API
+exposes to the extension.
```
xeenon: Should spell out that this does not release any new info that the extension can't infer otherwise. (Similar to what you have below in _Abuse Mitigations_.)

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+as a result of enterprise policies or for security reasons. For example,
+some browsers block access to extension store hosts.
+
+Another workaround is actually calling APIs which could throw if the extension
+has no access. This is slower and could lead to side-effects. In general it is
+better to prevent throwing errors when they can be circumvented.
+
+### Open Web API
+
+Host access generally is unique to webExtensions. Having such API on the open
```
xeenon:
```suggestion
Host access generally is unique to extensions. Having such API on the open
```

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+web thus does not make sense.
+
+## Implementation Notes
+
+Most browsers currently already have the concept of canAccess internally.
```
xeenon:
```suggestion
Most browsers currently already have the concept of `canAccess()` internally.
```

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Specification
+
+### Schema
+
+### types
+
+```ts
+dictionary AccessQuery {
+  url: string,
```
xeenon, line 52: We could also support `pattern` for a match pattern.

xeenon (changes requested):

hanguokai:
It would be best to add a section explaining the differences between `permissions.canAccess()` and `permissions.contains()`, as well as guidance on which API to use in different scenarios.

proposals/permissions-canAccess.md
carlosjeurissen: `injectOnly` indicates extensions are allowed to executeScripts but not fetch resources the origin resources.

Currently, in Chrome, `activeTab` grants both injection access and host access for fetch requests.

While Firefox and Safari do currently not grant host access for fetch requests.

See the following demo:
https://github.com/carlosjeurissen/webext-tech-demo-extensions/tree/main/demos/permissions-active-tab-host-access

If this is not supposed to happen I want my reward money from Google)

Weirdly enough in all browsers running `browser.permissions.contains()` still gives false.

Nonetheless, `injectOnly` is not very descriptive and implies the Chrome behaviour is default.

To resolve this, we can replace `canAccess` and `injectOnly` with `scriptingAccess` and `hostAccess`. This also resolves the repeated `canAccess` naming in the result.

proposals/permissions-canAccess.md
carlosjeurissen: Addressed by renaming it to `hostAccess` and `scriptingAccess`. See https://github.com/w3c/webextensions/pull/799#discussion_r2021220071

proposals/permissions-canAccess.md
xeenon: I see. That is currently an oversight in Safari, and maybe Firefox as well. I think we should match Chrome here and not bifurcate host permissions like this. I filed [bug 290764](https://webkit.org/b/290764) to track fixing this in Safari / WebKit.

The fact that `permissions.contains()` still gives `false` is expected, since the `activeTab` host permissions are scoped to a single tab and not global. If `permissions.contains()` took a `tabId` we could fix that too.

carlosjeurissen: @xeenon currently in Chrome the host permissions are global for the extension and not scoped to a single tab. Just confirmed this in Chrome 134.

@Rob--W What do you think about matching Chrome's behaviour here for Firefox like Safari is considering?

carlosjeurissen:
@hanguokai the difference between `permissions.contains()` vs `permissions.canAccess()` would be great to document on mdn once `permissions.canAccess()` is available. As for this PR, updated it to be more clear as to why `permissions.contains` is not sufficient.

proposals/permissions-canAccess.md
carlosjeurissen: Seems with the current behaviour in Firefox and Safari with `activeTab`, the API can only determine `scriptingAccess` if the API is supplied a `tabId`, `frameId` or `documentId`. Updated the PR to reflect this.

xeenon: @carlosjeurissen `activeTab` grants host permissions globally in Chrome? That is surprising and not what I understood the feature to do. If only we had a spec…

xeenon: I'm still not sold on bifurcating host permissions into scripting and host access. That "bug" should not be entombed into this as API.

hanguokai:
If I understand correctly, developers will use code like this:
```
async function canDoSomething(url) {
  // check host permission
  const hasPermission = await permissions.contains({ origins: [url] });
  if (!hasPermission) {
    const granted = await permissions.request({ origins: [url] });
    if (!granted) {
      return 'no-permission';
    }
  }

  // check can access
  const canAccess = await permissions.canAccess({ url: url });
  if (!canAccess) {
    return 'no-access';
  }

  return 'yes';
}

const result = await canDoSomething(url);
if (result == 'no-permission') {
  // tell the user I don't have permission to do it
} else if (result == 'no-access') {
  // tell the user that there is nothing I can do
} else {
  doSomething();
}
```

carlosjeurissen:
@hanguokai There is an edge case to consider in which extensions are granted only temporary hostAccess in the case of activeTab. In this situation, currently `permissions.contains()` would return false while `permissions.canAccess()` would return true. However not sure how much of a conscious decision this is based on. It might be a bug in the permissions.contains API.

The way you coded this example does bring up the question how extensions should address the lack of access. As with the current response no indication for this is given. The user might not have granted permanent host permissions. The host permissions might be present but the host is excluded with some enterprise policy. Introducing some reasons property could make sense. Something like disallowedReason: "browserExcluded".

hanguokai:
`activeTab` is tightly bound to a few APIs, so I never check it when I use it.

xeenon:
> As with the current response no indication for this is given. The user might not have granted permanent host permissions. The host permissions might be present but the host is excluded with some enterprise policy. Introducing some reasons property could make sense. Something like disallowedReason: "browserExcluded".

Having a reason would be helpful. Some other cases:
 * Users in Safari can deny any host in website settings (either requested or a currently open tab)
 * Users can grant them temporarily for 24 hours — we could include an expiration date property.

proposals/permissions-canAccess.md
Rob--W: > @xeenon currently in Chrome the host permissions are global for the extension and not scoped to a single tab. Just confirmed this in Chrome 134.
> 
> @Rob--W What do you think about matching Chrome's behaviour here for Firefox like Safari is considering?

Discussed today at https://github.com/w3c/webextensions/pull/823
In short, willing to consider aligning. Still wondering whether it would make sense to have a distinction between fetch vs content script access.

carlosjeurissen: @Rob--W In general extension can already do fetch requests by proxying the requests via the contentScripts. Making a distinction between fetch vs content script access would only make sense if Firefox and Safari do not plan to align with the Chrome behaviour before adding support for `permissions.canAccess()`.

As @zombie mentioned aligning is not straightforward for Firefox at this moment, inclined to design the API returning `scriptingAccess` and `hostAccess` or similar naming. Which at some point will both either be true / false once Firefox and Safari align with Chrome.

proposals/permissions-canAccess.md
```diff
@@ -0,0 +1,138 @@
...
+
+```ts
+Promise<AccessResult[]> browser.permissions.canAccess(
+  query: AccessQuery
+)
+```
+
+### Behavior
+
+In the case of `activeTab`, `scriptingAccess` is set true. If the browser
```
rdcronin, line 80: In the case of activeTab, what would we return if the developer specifies the URL, but not the tab / document?

I'd be inclined to say we should return false and that if tab / document is omitted, we should treat it as "can access on all tabs / documents for the given URL".

carlosjeurissen:
This was discussed during the 2026 London F2F meeting. A few points we agreed on:
- we can't merge the types of host access
- we are leaning towards querying the specific access type one wants information on

Will update the PR to reflect this.

Issue #117, opened 2021-10-21
Proposal: restrict `content_scripts.all_frames` to specific `main_frame` origins.


carlosjeurissen:
To reduce the attack surface of `content_scripts`, I'm proposing to allow authors to restrict `content_scripts.all_frames` to specific `main_frame` origins or matches. Imagine it like this:

```
"content_scripts": [{
  "css": [ "/content-script.css" ],
  "js": [ "/content-script.js" ],
  "matches": [ "https://www.example.com/*" ],
  "all_frames": true,
  "main_frame_origins": [ "<extension_origin>", "www.example.com" ],
}]
```

This makes sure content-scripts are not changing anything in situations they are not meant to make changes.

Basically the goal here is to only inject content_scripts if and only if the parent frame matches what the author specified.

xeenon:
I'm confused what that is hardening about content scripts. Is this about requiring the frame to be embedded in only one of the matching `main_frame_origins`?

carlosjeurissen:
It's not about making frames embedded-able. Sometimes extensions only want to inject a `content_script` in a specific `sub_frame` and not in the `main_frame`. There are specific websites (any website by default) which can be embedded on another website in an iframe. Thus if an extension wants to inject a content_script in these `sub_frames` on specific websites, it might open up holes in other websites where this specific `sub_frame` can also be embedded.

carlosjeurissen:
Closing this issue in favour of https://github.com/w3c/webextensions/issues/763

PR #802, opened 2025-03-30
Add proposal for `topFrameMatches` and `excludeTopFrameMatches`

polywock:
Related issue: #763

### DIFF
--- a/proposals/content-script-top-frame-matching.md
+++ b/proposals/content-script-top-frame-matching.md
@@ -61,7 +61,7 @@ The object definition within the `content_scripts` array in `manifest.json` is e
 }
 ```
 
-*Where `MatchPattern` is a string conforming to the standard [Extension Match Pattern syntax](https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns). 
+*Where `MatchPattern` is a string containing a single match pattern as specified in [mdn [Match patterns](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns). 
 
 
 #### `scripting.RegisteredContentScript` Type
@@ -94,7 +94,9 @@ dictionary RegisteredContentScript {
     *   And if `topFrameMatches` was specified, the **top-level document's origin** matches at least one pattern in `topFrameMatches`.
     *   And if `excludeTopFrameMatches` was specified, the **top-level document's origin** does *not* match any pattern in `excludeTopFrameMatches`.
 
+If any of the match patterns is not supported or understood by the browser, a soft error should be thrown. In `topFrameMatches`, the pattern will be skipped. If `topFrameMatches` is, or thus will end up being empty, no content scripts will be injected.
 
+If `excludeTopFrameMatches` is empty, the property will be ignored and not considered when injecting content scripts.
 **Determining the Origin for Top-Level Document Matching**
 
 The **top-level document's origin** is determined as follows:
--- a/proposals/content-script-top-frame-matching.md
+++ b/proposals/content-script-top-frame-matching.md
@@ -1,0 +1,139 @@
+
+# Proposal: Content Script Top Frame Origin Matching
+
+**Summary**
+
+API proposal to allow content script registration (both static and dynamic) to be restricted based on the origin of the top-level frame using standard [match patterns](https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns), enabling more intuitive and secure site blocking/allowing functionality for extensions.
+
+**Document Metadata**
+
+*   **Author:** [Polywock](https://github.com/polywock)
+*   **Sponsoring Browser:** *(Seeking browser sponsorship)*
+*   **Status:** Draft *(Seeking feedback and browser interest)*
+*   **Proposal Champions:** [Kzar](https://github.com/kzar), [Carlosjeurissen](https://github.com/carlosjeurissen), [Polywock](https://github.com/polywock)
+*   **Created:** 2025-03-30 
+*   **Related Issues:**
+    *   [w3c/webextensions#117](https://github.com/w3c/webextensions/issues/117)
+    *   [w3c/webextensions#668](https://github.com/w3c/webextensions/issues/668)
+    *   [Chromium Issue 40202338](https://issues.chromium.org/issues/40202338)
+
+## Motivation
+
+### Objective
+
+This proposal introduces a mechanism to further restrict content script injection by adding a filter based on the **origin** of the **top-level document**. This allows for more precise control over where scripts execute, while also enabling developers to create site blocklists or allowlists that better align with user expectations, improve performance, and enhance security.
+
+
+### Use Cases
+
+1.  **Intuitive Site Blocking/Allowing:** Many extensions offer users the ability to disable functionality on specific websites. Currently, using frame-level exclusion rules (like `excludeMatches`) on a dynamic content script is a common approach. However, this leads to counter-intuitive behavior:
+    *   If a user blocks `https://example.com/*` using `excludeMatches`, the extension's content script *still runs* on `https://example.com/page` if an embedded iframe loads content from a *different*, non-blocked domain (assuming `all_frames: true`).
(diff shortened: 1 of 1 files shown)


polywock:
If linking my GitHub to the W3 account is enough, the ipr issue should be resolved.

varish2001:
Many extensions offer, or would like to offer, users the ability to disable the extension on certain websites. But currently, the easiest and safest method of implementing that is by using a dynamic content script and adding the blocked sites to excludeMatches, but this results in behavior that doesn't align with user expectations.

When a user blocks a site like Reddit, they expect the extension to be completely disabled when they visit Reddit. However, with the usual excludeMatches approach: the content script still runs on Reddit if an embedded frame loads content from a different domain. Conversely, if a user visits another website that embeds Reddit content (e.g., a news site displaying a Reddit post), the extension is blocked from running within that embedded Reddit frame, even if the user did not intend to block it there.

Proposal
Content script registration should accept topFrameMatches / excludeTopFrameMatches. If provided, it will further restrict a content script based on the top frame's URL.

This allows extensions to, more easily and safely, implement a blocklist/allowlist system that's user intuitive.

Usage Example: Dark theme on all websites/frames unless it's on Reddit.

{
   "matches": ["https://*/*"],
   "exclude_top_frame_matches": ["https://www.reddit.com/*"],
   "all_frames": true,
   "js": ["force_dark_theme.js"]
}
Permission Warnings
Since this feature only restricts where a content script runs and does not expand its scope, the existing host permission warnings remain appropriate. Unlike matches, using top_frame_matches does not request any additional host access.

Security use case
As highlighted in @carlosjeurissen's proposal (https://github.com/w3c/webextensions/issues/117), allowing developers to restrict content scripts based on the main frame’s origin reduces the overall attack surface. This ensures scripts injected with all_frames: true only execute if the main frame matches specific origins, preventing unintended interactions or vulnerabilities from scripts running in unexpected contexts.

Current workaround limitations
You can achieve similar functionality by having the content script exit early after loading. In this approach, the script initially loads on every page, retrieves the user's blocked website list from browser.storage, and then checks the main frame's URL (e.g., via Location.ancestorOrigins) to determine whether it should continue running. However, this method comes with several drawbacks:

Inefficiency: The content script must still load, potentially across dozens or hundreds of open tabs. Even though it exits immediately without further logic, the effect of having these scripts loaded may have significant performance implications.
Asynchronous: Accessing the user's blocklist from browser.storage is asynchronous, making it unsuitable for extensions that rely on synchronous initialization.
Attack surface: If a user adds a sensitive website (e.g., a bank) to the blocklist, the content script should ideally never load on that site. Minimizing the extension’s footprint is always preferable. Even if the extension isn’t designed to be harmful, unintended vulnerabilities or bad code could still pose a risk.
Conflicts: Even if a script exits early, merely loading it can still lead to conflicts or unintended interactions with certain websites. Such conflicts might arise from polyfills, bundlers, or other underlying factors that developers may not necessarily be aware of. Such conflicts can be particularly challenging to diagnose and resolve.

malloy20001 (dismissed):

proposals/content-script-top-frame-matching.md
```diff
@@ -0,0 +1,139 @@
...
+        *   For dynamic API calls (`registerContentScripts`, `updateContentScripts`), the promise **must** be rejected with an appropriate error (e.g., `Match patterns must not specify a path other than '/*'.`).
+
+2.  **Injection Logic:** Assuming validation passes, a content script will be injected into a frame if and only if *all* the following conditions are met:
+    *   All existing checks based on the frame's own URL and context are satisfied.
+    *   And if `topFrameMatches` was specified, the origin of the top-level frame's URL matches at least one pattern in `topFrameMatches`.
+    *   And if `excludeTopFrameMatches` was specified, the origin of the top-level frame's URL does *not* match any validated pattern in `excludeTopFrameMatches`.
+
+**Origin Definition:** The origin consists of the scheme, host, and port (e.g., `https://www.example.com:443`). Match patterns are compared against this origin. The path component of the top-level URL is ignored during matching, as enforced by the validation rule disallowing specific paths in the patterns.
```
Rob--W: Although this seems precise, it is leaving room for interpretation. I think that it would make most sense to use the origin following the interpretation of `match_origin_as_fallback: true` as specified at https://w3c.github.io/webextensions/specification/index.html#determine-the-url-for-matching-a-document . That part of the spec draft was added in https://github.com/w3c/webextensions/pull/542
My motivation for this interpretation is to make sure that `about:blank`, `blob:` etc can be matched (which would naturally happen if you look at the document's origin) AND also that http(s) documents would be matched even if `Content-Security-Policy: sandbox` were to be used, which would give it an opaque origin.

And also to specify what should happen if an origin cannot be determined, so that we don't have issues like https://bugzilla.mozilla.org/show_bug.cgi?id=1798225 (`[DNR] Matching of missing initiators and/or with an opaque origin`)

Rob--W (commented):
Looks good in general, I have a request to specify the meaning of "origin" more precisely.

proposals/content-script-top-frame-matching.md
carlosjeurissen: Considering potential support for `<extension_urls>`, we should specify how browsers should handle match patterns they do not understand. Whether or not `excludeTopFrameMatches` should become a no-op or simply prevent the content script from running.

polywock: @Rob--W Updated the proposal to use the W3C spec for origin matching and to clarify behavior for unknown origin. Let me know if it didn't hit the mark. 

@carlosjeurissen  It's a cop out, but I've added -> "The browser must validate all patterns provided in `topFrameMatches` and `excludeTopFrameMatches` as it would validate patterns provided by `matches` and `excludeMatches`" as a base validation step for now.

proposals/content-script-top-frame-matching.md
```diff
@@ -0,0 +1,161 @@
...
+
+  // If specified: Only inject if the top-level frame's origin matches at least one of these patterns.
+  "top_frame_matches": ["MatchPattern"],
+
+  // If specified: Only inject if the top-level frame's origin isn't a match for any pattern.
+  "exclude_top_frame_matches": ["MatchPattern"]
+}
+```
+
+*Where `MatchPattern` is a string conforming to the standard [Extension Match Pattern syntax](https://developer.chrome.com/docs/extensions/develop/concepts/match-patterns).
```
carlosjeurissen: Currently there is no webExtensions `standard`. Mdn seems like a more neutral reference.

```suggestion
*Where `MatchPattern` is a string containing a single match pattern as specified in [mdn [Match patterns](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Match_patterns). 
```

proposals/content-script-top-frame-matching.md
```diff
@@ -0,0 +1,161 @@
...
+        *   For static declarations in `manifest.json`, this should result in a manifest parsing error, preventing the extension from loading.
+        *   For dynamic API calls (`registerContentScripts`, `updateContentScripts`), the promise must be rejected with an appropriate error (e.g., `Match patterns for top_frame_matches/exclude_top_frame_matches must not specify a path.`).
+
+2.  **Injection Logic:** Assuming validation passes, a content script will be injected into a frame if and only if *all* the following conditions are met:
+    *   All existing checks based on the frame's own URL and context are satisfied (e.g., `matches`, `excludeMatches`).
+    *   And if `topFrameMatches` was specified, the **top-level document's origin** matches at least one pattern in `topFrameMatches`.
+    *   And if `excludeTopFrameMatches` was specified, the **top-level document's origin** does *not* match any pattern in `excludeTopFrameMatches`.
+
+
```
carlosjeurissen, line 99:
```suggestion
If any of the match patterns is not supported or understood by the browser, a soft error should be thrown. In `topFrameMatches`, the pattern will be skipped. If `topFrameMatches` is, or thus will end up being empty, no content scripts will be injected.

If `excludeTopFrameMatches` is empty, the property will be ignored and not considered when injecting content scripts.
```

We should still determine what should happen in case invalid patterns show up in excludeTopFrameMatches.

carlosjeurissen (changes requested):

polywock:
@kiaraarose Hi! This PR fell off my radar for a while and I'm circling back to it now. Would you be able to give it a review when you have a moment? Thanks!

Issue #472, opened 2023-10-22
Add support for opening side panel without user interaction


twschiller:
Context
---
- The side panel is an essential surface for showing supplemental information/actions without blocking a user's interaction with a page
- The `sidePanel.open()` method supports [programmatically opening in response to user interaction in Chrome](https://developer.chrome.com/docs/extensions/reference/sidePanel/#user-interaction)
- For enterprise process excellence/compliance/governance/security browser extensions, requiring user interaction undermines the effectiveness of the extension 
- For enterprise productivity extensions (e.g., contact center agent assist), requiring user interaction increases the amount of training required for employees

Example Use Cases
---
- Automatically show relevant policies related to content on the page (e.g., state/locality-specific policies)
- Automatically show a checklist in the side panel
- Automatically show a decision tree in the side panel
- Automatically open a translation side panel when a foreign language is detected on the page
- Automatically open a reporting form if a trust and safety keyword is detected on a page
- Automatically open a grammar/writing sidebar on specific sites like Google Docs, MS Word, etc.

Options:
---
1. Allow programmatically opening the side panel if the extension has `<all_urls>` permission
2. Always allow programmatically opening the side panel
3. Introduce a new optional permission granting permission to automatically open the side panel

Discussion
---
- Including the capability under the `<all_urls>` permissions could make sense, as compliance/governance/security extensions would likely have that permission already, and it also already triggers a more in-depth store review (at least for Chrome)

Related Features/Permissions/APIs
---
- `action.openPopup` [does not require user interaction in Chrome](https://github.com/w3c/webextensions/issues/160)
- [captureVisibleTab](https://developer.chrome.com/docs/extensions/reference/tabs/#method-captureVisibleTab) is supported via the `<all_urls>` permission

Rob--W:
Firefox's [`sidebarAction.open()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/sidebarAction/open) method requires user interaction. Additionally it is also possible to open the sidebar by default once after install, with the `"open_at_install": true` option in the manifest.

Sophie1142:
Upvoting this request! We're building a productivity extension and it hinders its usefulness if users have to remember to open the extension first.

SpaceK33z:
Would like to add to this request, that starting with Chrome 118+ it is possible to programmatically open the popup (`chrome.action.openPopup`). For this **no user interaction is required**.

[...]

Issue #160, opened 2022-02-06
Ensure consistency of `action.openPopup` API across browsers


oliverdunk:
A Chromium bug was [recently opened](https://bugs.chromium.org/p/chromium/issues/detail?id=1245093) sharing an action.openPopup API, and a basic implementation has shipped restricted to the dev channel. This is super exciting since it's the first time Chrome extensions look set to get this ability.

Since Firefox has already implemented [browserAction.openPopup](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/openPopup), I'm curious on a few things:

- Would Mozilla be willing to either support the new namespace as an alias or try to work with Google to encourage use of the existing API naming? In an ideal world I would rather see Chromium implement the Firefox API, since that has already shipped and seems to work well. However, they seem to be wanting to diverge from the private API they have already implemented so I'm not sure how feasible that is.
- One big step the Chromium API has taken is removing the requirement for a user action. The API overview doc says that they "do not currently plan any mitigations for attacks of annoyance" because "extensions can already do far worse than this with the tabs API and window creation." Would Mozilla consider the same? It immediately solves some of the native messaging use cases raised at the start of #15.

---

ℹ️  Update: The initial questions here have been resolved and this issue is now tracking more nuanced behaviour differences between browsers. See the [current behaviours](https://github.com/w3c/webextensions/issues/160#issuecomment-1127042759) and [proposed standard behaviours](https://github.com/w3c/webextensions/issues/160#issuecomment-1149112209).

oliverdunk:
@zombie / @Rob--W, it would be great to get your input from the Mozilla side.

@dotproto, I'm not sure if there's anything you can share on behalf of Google? Would you be open to discussions about using the same API namespace as Mozilla or is that off the cards given that you already use that namespace for an internal API? The crbug seems quiet unfortunately.

zombie:
> * Would Mozilla be willing to either support the new namespace as an alias or try to work with Google to encourage use of the existing API naming?

We already added support for the `action` manifest key/namespace for MV3 in bug [Bug 1706398](https://bugzilla.mozilla.org/show_bug.cgi?id=1706398).

> * One big step the Chromium API has taken is removing the requirement for a user action. The API overview doc says that they "do not currently plan any mitigations for attacks of annoyance" because "extensions can already do far worse than this with the tabs API and window creation." Would Mozilla consider the same?

[...]

PR #817, opened 2025-04-24
Proposal: Focus management API

dotproto:
This proposal introduces a `tab.focus()` method and supporting `tab` namespace properties.

**Status**

This is the first draft of the proposal. It is not entirely contiguous, but initial feedback is welcome.

### DIFF
--- a/proposals/tab-focus-api.md
+++ b/proposals/tab-focus-api.md
@@ -210,9 +210,7 @@ N/A
 
 ### Ancillary UI surfaces
 
-THere are other notable UI surfaces that are extensions are not currently able
-to focus such as the user agent’s address bar or bookmarks bar. If so desired,
-`FocusArea` could be expanded to support these surfaces.
+There are other notable UI surfaces that are extensions are not currently able to focus such as the user agent’s address bar or bookmarks bar. If so desired, `FocusArea` could be expanded to support these surfaces.
 
 ### New UI surfaces
 
--- a/proposals/tab-focus-api.md
+++ b/proposals/tab-focus-api.md
@@ -1,0 +1,223 @@
+# Proposal: Focus management API
+
+**Summary**
+
+Allow extensions to change focus between browser-provided web UI surfaces with a new protected runtime API.
+
+**Document Metadata**
+
+**Author:** @dotproto
+
+**Sponsoring Browser:** TBD
+
+**Contributors:** \<other contributors emails or GitHub handles\>
+
+**Created:** 2025-04-24
+
+**Related Issues:**
+
+* [Issue 693](https://github.com/w3c/webextensions/issues/693): Add API to switch focus from sidePanel to main document to enable accessibility features
+
+## Motivation
+
+### Objective
+
+The **primary objective** of this proposal is to provide extensions with a way to move focus between the main browsing context in a tab and an extension’s sidebar.
+
+A **secondary objective** is to establish an extensible interface to manage focus within a user agent.
+
+#### Use Cases
+
+##### Transitioning between multiple UI surfaces during a workflow
+
+Allow users to trigger an extension feature in a secondary UI surface and immediately have subsequent interactions take place in the main browsing context of a document.
+
+For example, an extension could respond to a user clicking a button in the extension's sidebar by focusing the main document and using web platform APIs to focus an input element within that document. This in turn allows the user’s subsequent keyboard input to occur within the main document as the user expects.
+
+### Known Consumers
+
+Extensions that want to bridge user flows between the sidebar and web page in a given tab. Specific extensions that have indicated they will adopt this capability when available include:
+
+* Tango - [https://tango.ai](https://tango.ai)
+
+## Specification
+
+### Schema
+
+#### TypeScript Definition
+
+```javascript
+declare namespace browser.tabs {
+  /* METHODS */
+  export function focus(options: FocusOptions): Promise<FocusResponseCodeValue>;
+
+  /* PROPERTIES */
+  export enum FocusArea {
+    main = 'main',
+    sidebar = 'sidebar',
+  }
+
+  export enum FocusResponseCode {
+    success = 'success',
+    invalid = 'invalid',
+    unsupported = 'unsupported',
+    unavailable = 'unavailable',
+  }
+
+  /* TYPES */
+  export interface FocusOptions {
+    windowId?: number;
+    tabId?: number;
+    area: FocusAreaValue;
+  }
+
+  export interface FocusResponse {
+    code: FocusResponseCodeValue;
+    message?: string;
+  }
+
+  // The following types map a given string enum to a union of it's values. This
+  // allows us to work around one of the limitations of TypeScript enums:
+  // literals that strictly equal an enum's value cannot be used in places that
+  // expect the enum. For example:
+  //
+  //   enum Shape { square = 'square' };
+  //   declare function isShape(shape: Shape): void;
+  //   isShape('square');
+  //   // Error: Argument of type '"square"' is not assignable to parameter of
+  //   // type 'Shape'.(2345)
+  export type FocusAreaValue = `${FocusArea}`;
+  export type FocusResponseCodeValue = `${FocusResponseCode}`;
+}
+```
+
+#### tabs.focus()
+
+Makes a request to the user agent to focus the area of the user interface specified by the `options` parameter.
+
+At least one of the optional properties must be specified. When multiple properties are specified, all properties must be valid and the combined set of properties must match a focusable surface in the user agent that the extension is allowed to access. If they do not, the promise returned by this method will be reject with an appropriate error.
+
+##### `options`
+
+**Type:** `FocusOptions`
+
+Required. An object that specifies which portion of the agent’s user interface should receive focus.
+
+###### `windowId`
+
+**Type:** `number`
+
+Optional. The ID of the window associated with the UI surface being targeted. If omitted, fall back to the most recently focused window accessible to the extension.
+
+###### `tabId`
+
+**Type:** number
+
+Optional. The ID of the tab associated with the UI surface being targeted. If omitted, fall back to the most recently focused window accessible to the extension.
+
+###### `area`
+
+**Type:** `tabs.FoucsArea`
+
+Optional. The UI surface associated with a given window and tab to focus. If omitted, fall back to `tabs.FocusArea.main`.
+
+##### Return value
+
+**Type**: Promise\<undefined\>
+
+The returned promise rejects in the following scenarios:
+
+If the provided set of ID properties (`windowId`, `tabId`, etc.) do not resolve to a concrete focusable area, the promise MUST be rejected with a `ReferenceError()`. The agent SHOULD set a human readable error message that specifies which reference(s) were invalid.
+
+If the caller provides an `area` value that is not exposed on `tabs.FocusArea`, the promise MUST reject with a `TypeError`. The error’s `message` SHOULD identify the invalid area specified.
+
+#### tabs.FocusArea
+
+**Type:** Enumeration (string)
+Defines the set of areas (sub-regions) of an agent’s user interface that can be focused by an extension.
+
+`main` (value: `"main"`) – The content area of a given window and tab combination. This area is used to display the web page tab. In user agents that support multiple documents per tab, this value refers to the primary or first document in the tab.
+
+`sidebar` (value: `"sidebar"`) – The extension’s `sidePanel` or `sidebarAction` page. The extension’s sidebar page for a given window and tab. When focusing this area, the extension's sidebar will be activated if it is open. If not, the focus() call will fail.
+
+#### tabs.FocusResponse
+
+**Type:** Object
+
+### Behavior
+
+If the user agent does not support an area, does not allow the area to be focused, or the area is disabled by the user’s UI settings, \`focus()\` calls targeting the area will fail.
+
+### New Permissions
+
+No permissions changes are in scope for this initial proposal.
+
+Future iterations on focus management capabilities may warrant the introduction of a permission with a warning. See Security and Privacy for details.
+
+### Manifest File Changes
+
+N/A
+
+## Security and Privacy
+
+Focus management presents a moderate risk to end users. While changing the user agent’s current focus does not grant access to any data in itself, it greatly increases the risk that a user will perform an unintended action. When combined with [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation), this capability can be used to receive escalated permissions or to perform a restricted operation without the user’s express permission.
+
+As such, this proposal intentionally does not entertain the possibility of allowing extensions to focus user agent UI surfaces such as an extension’s action button. When considering future expansion of focus management capabilities, the authors recommend gating the ability to focus dangerous such UI surfaces behind a permission with a warning.
+
+### Exposed Sensitive Data
+
+N/A
+
+### Abuse Mitigations
+
+N/A
+
+### Additional Security Considerations
+
+The current design of this proposal allows extensions to infer stateful information about the user agent that is possible otherwise (I think?). For example, if the sidebar cannot be focused because the user agent is currently prompting the user with an uninterruptable modal, the `focus()` call will reject and the error returned will indicate that the target area is currently unavailable.
+
+## Alternatives
+
+### Existing Workarounds
+
+N/A
+
+### Open Web API
+
+The web has two relevant mechanisms for managing focus on a website: the [HTMLOrSVGElement](https://html.spec.whatwg.org/multipage/dom.html#htmlorsvgelement) [focus()](https://html.spec.whatwg.org/multipage/interaction.html#dom-focus) method and the [Window](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window) [focus()](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) method. The former is used to control focus within a page while the latter is used to move focus between HTML documents. As mentioned [in a note](https://html.spec.whatwg.org/multipage/interaction.html#focus-management-apis:dom-window-focus) under the definition of the Windows interface’s focus() and blur() algorithms:
+
+> Historically, the focus() and blur() methods actually affected the system-level focus of the system widget (e.g., tab or window) that contained the navigable, but hostile sites widely abuse this behavior to the user's detriment.
+
+Today, a web page’s ability to move focus between documents is limited to moving focus from a parent window to a child window in response to a user activation.
+
+### Modify Web APIs
+
+The [HTML specification](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) notes that, “historically, the \[\] `focus()` and `blur()` methods actually affected the system-level focus of the system widget (e.g., tab or window) that contained the [navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#navigable), but hostile sites widely abuse this behavior to the user's detriment.”
+
+An alternative approach would be to restore this historical behavior for `window` instances exposed to extension extension contexts. This could either be granted to all extensions contexts or limited to secure contexts. Both approaches have notable drawbacks.
+
+WRITING NOTES:
+
+Unprivileged contexts - capability would need to be exposed in the page’s process. Primarily enables attacks annoyance, but capability was removed from the open web for a reason.
+
+Only privileged contexts - less consonant developer experience, more difficult to use correctly.
+
+## Implementation Notes
+
+N/A
+
+## Future Work
+
+### Ancillary UI surfaces
+
+THere are other notable UI surfaces that are extensions are not currently able
+to focus such as the user agent’s address bar or bookmarks bar. If so desired,
+`FocusArea` could be expanded to support these surfaces.
+
+### New UI surfaces
+
+As user agent interfaces and their affordances for extensions evolve, it may become necessary to expand the set of focusable UI surfaces supported by `FocusArea`. For example, if more browsers add support for the concept of a “split view” (displaying multiple web pages in a single tab), we may need to introduce a mechanism of selecting between `”main”`areas in a given tab.
+
+### Additional `FocusOptions` targeting properties
+
+In addition to the current `windowId` and `tabId` properties, it may be desirable to support other common methods of targeting document content, namely `frameId` and  `documentId`. These were omitted from the initial draft in part because it was not immediately clear how to distinguish between a desire to focus a specific sub-frame and simply focusing the page itself.


hanguokai:
Thanks for the proposal! Besides the method `tab.focus()`, I think we also need a method to determine whether the current area has obtained the focus or where the current focus is.

dotproto:
When writing the first draft of this proposal I didn't have a clear picture for how we should handle errors in `foucs()` calls. As a result, I accidentally mixed up the concepts of rejecting with an Error and resolving with an object that signaled that the operation could not be completed. I'd appreciate feedback or recommendations on how to approach this.

dotproto:
> I think we also need a method to determine whether the current area has obtained the focus or where the current focus is.

My intent was that the promise returned by the `focus()` call should signal the success/failure of the request. As mentioned in my last comment, though, the current draft mixes up rejecting with an error and resolving with a `code` to signal success/failure.

@hanguokai, can you describe a use case for wanting to query the browser's current focus? I didn't see a need for that in the use case that motivated the proposal.

qupig:
@dotproto
> When writing the first draft of this proposal I didn't have a clear picture for how we should handle errors in `foucs()` calls. As a result, I accidentally mixed up the concepts of rejecting with an Error and resolving with an object that signaled that the operation could not be completed. I'd appreciate feedback or recommendations on how to approach this.
>
> My intent was that the promise returned by the focus() call should signal the success/failure of the request. As mentioned in my last comment, though, the current draft mixes up rejecting with an error and resolving with a code to signal success/failure.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

I think in the Async and Promise model, "rejected" means error/failed, "resolved/fulfilled" means success, you should never return an error code in the "resolved results" to signal failure.

If there must be an error code there, put it in the "rejected reason".

qupig:
@dotproto Preliminary review of the proposal, I'm sorry I object to the API design.

Such a call is very strange and cumbersome:

```js
browser.tabs.focus({area: "sidebar"})
```

The main reasons are:
- The [`browser.tabs`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs) namespace should not be used to manipulate `sidebar` interactions; it is not intuitive and makes it difficult for developers to locate the API.
- Especially when we have a corresponding [`browser.sidebarAction`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/sidebarAction) namespace.
- The `sidebar` area (i.e. the secondary UI you mentioned) is not actually part of the `tab` scope, it is not exactly related or corresponding to the current page.

I think possible alternative designs are:
- Use a new namespace, such as `browser.focus` instead of `browser.tabs.focus`.
- Implement the `focus` method in various namespaces, namely: `browser.tabs.focus()`, `browser.sidebarAction.focus()`, `chrome.sidePanel.focus()`, `browser.omnibox.focus()` etc.

I tend to the latter because it's intuitive enough and simple enough that you don't actually need to pass in any parameters.

The second problem has to be mentioned here, that is, we should not pass in options such as `windowId`, `tabId` in this API, because that is the job of [`tabs.update()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update) or [`windows.update()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/update). The API should only be used for focus switching on the current tab of the current window.

The third point is that above I also mentioned `browser.omnibox.focus()`, cause I think that in addition to `tab/page`, `sidebar`, we might also need to shift the focus to [`omnibox`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/omnibox), but I don't have a specific use case for that at the moment.

proposals/tab-focus-api.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Manifest File Changes
+
+N/A
+
+## Security and Privacy
+
+Focus management presents a moderate risk to end users. While changing the user agent’s current focus does not grant access to any data in itself, it greatly increases the risk that a user will perform an unintended action. When combined with [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation), this capability can be used to receive escalated permissions or to perform a restricted operation without the user’s express permission.
+
+As such, this proposal intentionally does not entertain the possibility of allowing extensions to focus user agent UI surfaces such as an extension’s action button. When considering future expansion of focus management capabilities, the authors recommend gating the ability to focus dangerous such UI surfaces behind a permission with a warning.
```
qupig, line 165: The security concern here is a misinterpretation of `user activation`.

Because we actually already have the relevant APIs for opening the extensions `popup` or `sidebar`:
- [`action.openPopup()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/openPopup)
- [`sidebarAction.open()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/sidebarAction/open)

qupig:
Another issue is like what I mentioned in my initial comment in https://github.com/w3c/webextensions/issues/693#issuecomment-2641898793, some of the current APIs automatically shift focus when we call them, and we should probably add options to those APIs to prevent focus shifting.

Although we could for example do so (to keep focus still on `sidebar`):

```js
await browser.tabs.update(tabId, { active: true }); // focus automatically shifts to `tab` (page)
await browser.sidebarAction.focus(); // manually shift the focus back to `sidebar`
```

But it has two focus transfer, which can be unexpected or delayed, ultimately reducing the user experience.

So a better alternative might be add a `focus` property in [`updateProperties`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#updateproperties):

```js
browser.tabs.update(tabId, { active: true, focus: false });
```

> [!NOTE]
> [`{active: true}`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#active) will not make `tab` to gain focus if the current window does not have focus.
> Therefore, there are differences in the uses and purposes of the two.

And we may want to add corresponding property to [`tabs.Tab`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab), which may fulfill the needs of, say, @hanguokai.

And I think this maybe a better alternative to the above `browser.tabs.focus()`, because it represents the concrete `tab` rather than the vague plural notion of `tabs`. But one issue that needs attention and consideration is that when set `{focus: true}`, `{active: true}` may also should be automatically specified at the same time, otherwise the result may be unexpected.

proposals/tab-focus-api.md
qupig: And these APIs have the `user activation` restriction applied already:
```js
browser.action.openPopup();
Uncaught (in promise) Error: action.openPopup may only be called from a user input handler
browser.sidebarAction.open();
Uncaught (in promise) Error: sidebarAction.open may only be called from a user input handler
```

carlosjeurissen:
Currently, calling `sidebar.open()` does not directly focus the opened sidebar. If we do not want to change this default behaviour for opening a sidebar, we can consider passing focus as an option to the open method as such:

```js
browser.sidebar.open({
  focus: true,
});
```

This has a number of benefits:
- Assures the sidebar will be open (i.e. no error when calling while the sidebar is closed)
- `sidePanel.open` and `sidebarAction.open` currently already require user activation
- Already comes with tabId/windowId and a lot of related logic
- Solves awkward chaining of calls to `sidebar.open()`, followed by `tabs.focus({area: "sidebar"})`
- Will not be awkwardly under the tabs namespace

Considering `action.openPopup()` already allows focussing like this I do not see any new attack vectors.

It could even be argued focussing the sidebar after `sidebar.open()` should be the default and expected behaviour. Which, for potential backwards compatibility, we could make the default in a unified `browser.sidebar.open()` method.

qupig:
@carlosjeurissen

> Assures the sidebar will be open (i.e. no error when calling while the sidebar is closed)

Why does it cause an error when sidebar is closed? Do you want to express that it is called again when it is already opened?

> Already comes with tabId/windowId and a lot of related logic

What are you referring to?

> Solves awkward chaining of calls to `sidebar.open()`, followed by `tabs.focus({area: "sidebar"})`

Actually I don't think it's awkward:

```js
await browser.sidebarAction.open();
await browser.sidebarAction.focus();
```

But I don't object to adding options to the `open()` method, it has some convenience.

> It could even be argued focussing the sidebar after sidebar.open() should be the default and expected behaviour. 

I can see the logic of the argument, but I can also immediately think of some negative examples, such as if sidebar is just used to show some information, then it does not need to have to get focus.

proposals/tab-focus-api.md
dotproto: There have been a few discussions in this group about removing the user activation requirement from `action.openPopup()`, `sidebarAction.open()`, and `sidePanel.open()`. To the best of my knowledge browsers are aligned on removing the requirement from `action.openPopup()`, but I'm not sure there's consensus around the sidebar concepts.

* https://github.com/w3c/webextensions/issues/472
* https://github.com/w3c/webextensions/issues/160

To be clear, toggling the visibility of the sidebar and popup are not the concern here. User interactions with browser controlled extension UI surfaces apply a user activation (also refereed to as a user gesture) flag in the event handler. Many web and extension APIs and are also gated behind user activation, such as invoking `permissions.request()` to request an optional (or disabled) permission.

More importantly, though, user interactions with browser controlled extension UI surfaces also grant `activeTab`. This gives the extension temporary host permission access to the page on which the grant occured, which includes the ability to inject scripts into a page. In situations where an extension doesn't request `activeTab` but  does request broad host permissions, users can use browser settings to configure the extension to only "run on click." In this mode, extensions behave similarly to the `activeTab`: they will only inject content scripts when an appropriate  trigger is invoked by the user and the host permissions granted are revoked after navigation.

dotproto:
> I think in the Async and Promise model, "rejected" means error/failed, "resolved/fulfilled" means success, you should never return an error code in the "resolved results" to signal failure.
>
> If there must be an error code there, put it in the "rejected reason".

The W3C TAG's [Writing Promise-Using Specifications](https://www.w3.org/2001/tag/doc/promises-guide) states that "rejection reasons must be Error instances" ([4.1.2.](https://www.w3.org/2001/tag/doc/promises-guide#reasons-should-be-errors)) and "rejections must be used for exceptional situations" ([4.1.3.](https://www.w3.org/2001/tag/doc/promises-guide#rejections-should-be-exceptional)). That provides some pretty strong guidance on how rejections should be handled. 

The thing I am unsure about is what failure cases should be considered exceptional. I think it makes sense to reject if the provided `tabId` doesn't match any open tabs or the `area` provided isn't supported by the browser. I'm less sure about rejecting when the target can't be focused because the browser has a modal open. I'm not sure if there are other cases I haven't yet considered where the UI surface isn't targeted but rejecting isn't appropriate.

> @dotproto Preliminary review of the proposal, I'm sorry I object to the API design.

No need to apologize. We're here to share and discuss feedback :)

I agree with the critiques you made of exposing this capability on the `tabs` namespace. As @rdcronin noted a few years back, that namespace is currently a bit of a junk drawer and we should try to avoid dumping more stuff into it.

I had this method in `tabs` because I was originally targeting moving focus between a sidebar and the main document of a tab. Since both concepts are directly linked to a given tab, it seemed appropriate enough namespace to start. As I iterated on it and the method became more generic, I agree that this no longer feels like the right namespace. 

I hadn't given the idea of creating different `focus()` methods for each namespace much consideration. I idea did come up briefly, but at the time I was thinking about designing for the future and I wasn't confident that the pattern of having distinct namespaces for each matching UI surface would hold. 

That said, the longer I sit with this idea the more I like it. It also addresses something that's bothered me about my currently proposed approach: if the options object grows to accommodate the needs of arbitrary UI surfaces, it may become very difficult for developers (and maintainers) to reason about. Namespace specific `focus()` methods would greatly simply the common set of properties and make it much easier to tailor new targeting controls to specific UI surfaces. 

> The second problem has to be mentioned here, that is, we should not pass in options such as `windowId`, `tabId` in this API, because that is the job of [`tabs.update()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update) or [`windows.update()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/update). The API should only be used for focus switching on the current tab of the current window.

I was envisioning the currently proposed `tabs.focus()` method in part as a more direct way focus a given window and tab. Personally, I find it a bit annoying that to focus a different window and tab you currently have to do something like this:

```js
const focusTab = async (tabId) => {
  const tab = await browser.tabs.update(tabId, { active: true });
  browser.windows.update(tab.windowId, { focused: true });
}
```

If we have namespace specific methods that only operate on the current window and tab, this would pattern would become more cumbersome:

```js
const focusTabSidebar = async (tabId) => {
  const tab = await browser.tabs.update(tabId, { active: true });
  browser.windows.update(tab.windowId, { focused: true });
  if (browser.sidebarAction) {
    await browser.sidebarAction.open();
    await browser.sidebarAction.focus();
  } else {
    await browser.sidePanel.open();
    await browser.sidePanel.focus();
  }
}
```

This could be simplified in the future if we align on a common `sidebar` namespace as has been floated in previous discussion, but it's still a decent number of steps for what should be a relatively simple operation. And this doesn't include error handling. 

> The third point is that above I also mentioned `browser.omnibox.focus()`, cause I think that in addition to `tab/page`, `sidebar`, we might also need to shift the focus to [`omnibox`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/omnibox), but I don't have a specific use case for that at the moment.

I touched on this idea in the Future Work >  Ancillary UI surfaces section of the proposal. We also briefly discussed it in today's WECG meeting. @Rob--W raised serious concerns about allowing extensions to focus the address bar, but we didn't have time to begin exploring those concerns.

proposals/tab-focus-api.md
```diff
@@ -0,0 +1,223 @@
...
+
+N/A
+
+## Future Work
+
+### Ancillary UI surfaces
+
+THere are other notable UI surfaces that are extensions are not currently able
+to focus such as the user agent’s address bar or bookmarks bar. If so desired,
+`FocusArea` could be expanded to support these surfaces.
```
dotproto:
```suggestion
There are other notable UI surfaces that are extensions are not currently able to focus such as the user agent’s address bar or bookmarks bar. If so desired, `FocusArea` could be expanded to support these surfaces.
```

dotproto:
@carlosjeurissen, I see how adding a focus option to `browser.sidebar.open()` could be useful. That said, I worry that using this as the main way of focusing the sidebar could lead to accidental navigation of the sidebar page. That in turn could lead to unexpected refreshes, loss of interaction state, or even data loss. I think we'd still need a way to focus the sidebar non-destructively. 

---

> > Assures the sidebar will be open (i.e. no error when calling while the sidebar is closed)
> 
> Why does it cause an error when sidebar is closed? Do you want to express that it is called again when it is already opened?

That's basically the behavior described in the current proposal (highlight added for clarity):

> #### tabs.FocusArea
> ...
> `sidebar` (value: `"sidebar"`) – The extension’s `sidePanel` or `sidebarAction` page. The extension’s sidebar page for a given window and tab. <mark>When focusing this area, the extension's sidebar will be activated if it is open. If not, the focus() call will fail.</mark>

---

> > Already comes with tabId/windowId and a lot of related logic
> 
> What are you referring to?

The [`sidePanel.open()`](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-open) and [`sidePanel.setOptions()`](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-setOptions) methods allow callers to set per-tab sidebar pages. That's why the current proposal includes the optional `windowId` and `tabId` properties in `tabs.focus()`'s `options` object.

qupig:
> The W3C TAG's ...
> The thing I am unsure about is what failure cases should be considered exceptional. I think it makes sense to reject if the provided `tabId` doesn't match any open tabs or the `area` provided isn't supported by the browser. I'm less sure about rejecting when the target can't be focused because the browser has a modal open. I'm not sure if there are other cases I haven't yet considered where the UI surface isn't targeted but rejecting isn't appropriate.

Thanks for citing the specification, but I think it is exactly in line with my understanding.

In this particular case, when you call the `focus()` method, if the desired object successfully gains focus, it is resolved successfully (regardless of whether it has already gained focus before), and if it does not get focus, it is rejected. I think this is clear to me. Maybe my previous expression was not clear enough.

Otherwise, it is a nightmare for developers, both "to catch error when throw" and "to check the return value when resolved" to confirm whether `foucs` is successful.

> I was envisioning the currently proposed `tabs.focus()` method in part as a more direct way focus a given window and tab. Personally, I find it a bit annoying that to focus a different window and tab you currently have to do something like this...

Simply put, I agree. 
But I was mean, unless you are going to design a new and well-developed API set and intend to **replace** and **deprecate** some old ones, there are multiple overlapping methods that may not be a good result.

> If we have namespace specific methods that only operate on the current window and tab, this would pattern would become more cumbersome...

I think there are several problems here:
- We may need to retain flexibility, such as focusing a part of the window but not focusing the window itself, then when the user actively switches to the window, the focus will be activated in the correct position.
- While I can see the benefits of simplifying these steps, and for simple purposes I like it, it is important to consider that some operations are not strictly bound.
- The misalignment of `browser.sidebarAction` and `chrome.sidePanel` is a significant problem here, and I'm not sure if there are other issues that track this.

> I touched on this idea in the Future Work > Ancillary UI surfaces section of the proposal. We also briefly discussed it in today's WECG meeting. @Rob--W raised serious concerns about allowing extensions to focus the address bar, but we didn't have time to begin exploring those concerns.

Yes I saw this in later reading. I really didn't think of a use case at the moment, but it is still necessary to consider this possibility in design. So maybe it's not worth too much attention before fully considering the use cases and risks.

---

> That's basically the behavior described in the current proposal (highlight added for clarity):

Thanks, this explains my question, which makes sense for the `focus` operation, I was limited to the `open` method before.

> The [`sidePanel.open()`](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-open) and [`sidePanel.setOptions()`](https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-setOptions) methods allow callers to set per-tab sidebar pages. That's why the current proposal includes the optional `windowId` and `tabId` properties in `tabs.focus()`'s `options` object.

OK, thanks and that explains part of it.

I think the concept of user intuition is that one interaction (such as a click) will only get a certain part/element visible in the current screen to focus, or switch to another view.
In other words, it would sound weird if a click focuses on a part/element in a currently invisible view.
I hope this explains why I was expected it to only work on the visible part of the current window (tab, sidebar etc.).

In fact, I am disappointed with the processing of the first click of the current non-activated window, different performances in different applications and systems. 
In some cases, the click will only activate the window instead of focusing elements and triggering events, while in others, it will not, even in the same window of the same application, there still different processing situations.
In other words, when I click on an element in a non-activated window, I don't know if I could to click once to interact, or if I have to click a second time to make a difference.

So I can see the benefits of this for powerful automation and convenience. And if this is the case, you may need to design all the advanced and complex focus management systems separately in `browser.focus` namespace and may require explicit permission from users because it sounds unexpected to most users.

But I think this seems a little bit beyond what we are currently trying to solve.

proposals/tab-focus-api.md
qupig: Okay, that explains the broader considerations.
Because it initially sounds like preventing the open/focus popup (action button).
It should be mentioned that if intend to focus only the tab-top-window, it is unnecessary to obtain host permissions.

jsuiker:
*** waiting with bated breath ***

carlosjeurissen:
> In this particular case, when you call the `focus()` method, if the desired object successfully gains focus, it is resolved successfully (regardless of whether it has already gained focus before), and if it does not get focus, it is rejected. I think this is clear to me. Maybe my previous expression was not clear enough.
> 
> Otherwise, it is a nightmare for developers, both "to catch error when throw" and "to check the return value when resolved" to confirm whether `focus` is successful.

An example of a web API which resolves even on some errors is the Fetch API. Which can resolve while the server responds with a 404 error.

In extension space, APIs like `permissions.request()` resolve even when the user rejected said permissions.

As for if this API should reject or resolve when attempting to focus, we can look at existing behaviour in other APIs like `action.openPopup()`. Currently in Chrome, when a sign-in prompt like "sign in to x.com with google.com" is open, calling `action.openPopup()` rejects with a generic "Failed to open popup." Error. So rejecting on failure of .focus() seems to make sense compared to existing behaviour.

In general it seems blocking UI patterns like that needs some more consideration and documentation. For example, with these Google sign-in prompts, extension popups can not be opened not even when the user clicks an extension icon. Which might be a bug @oliverdunk.

qupig:
@carlosjeurissen Thanks for providing more references.

> An example of a web API which resolves even on some errors is the Fetch API. Which can resolve while the server responds with a 404 error.

In this case, fetching data about the error page may be expected behavior, so reject is obviously inappropriate. It's not a error of the `fetch` API, it completes the "fetching" task, doesn't it?

> In extension space, APIs like `permissions.request()` resolve even when the user rejected said permissions.

I think in this case it's because it will be called multiple times during the extension lifecycle, it's more of a query to return rather than a task to resolve. In particular, it only actually pops up the authorization dialog to the user the first time.

And the `focus()` here is a specific one-time task, and I couldn't see a situation where it fails without the need to rejected.

In any case, thanks you all for putting forward these references and I will think more carefully and critically about expressing my views in the future.

dotproto:
> In this case, fetching data about the error page may be expected behavior, so reject is obviously inappropriate. It's not a error of the fetch API, it completes the "fetching" task, doesn't it?

That's true. On the other hand, if you try try to perform a cross origin request and the response doesn't have the right CORS headers set the `fetch()` call will throw at `TypeError`. You can see in action by navigating to https://example.com, then opening the browsers developer tools and executing the following snippet in the console: `await fetch('https://example.org')`.

---

Between the discussion here and some offline followup conversations I had with browser engineers, it seems like the best path forward is to revise this proposal to use per-namespace `focus()` methods. I'm hoping to tackle this in the next few days.

qupig:
> Between the discussion here and some offline followup conversations I had with browser engineers, it seems like the best path forward is to revise this proposal to use per-namespace `focus()` methods. I'm hoping to tackle this in the next few days.

Thank you for all your work and it's great to see it continue to be pushed forward.

I think "focus management" will have more advanced and meticulous improvements in the future, not only at the browser level, but also at the system/window level.

But as far as solving the problem we need right now, per-namespace `focus()` methods (except `tabs.focus()`) would be a relatively easy and gentle path.

Note that `tabs.focus()` should still probably be avoided because it is not clear which `tab` is, and `tabs` does not represent the current `page` part. If the double-column `page` is implemented in the future, the problem will be more obvious.

For this, I still think that using `focus` property in [`updateProperties`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#updateproperties) is a viable alternative in the `tabs` namespace, which is more in line with current API logic. https://github.com/w3c/webextensions/pull/817#issuecomment-2828746041

qupig:
I'm now more of the opinion that adding a `focus` property in [`updateProperties`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#updateproperties) and the `focused` property to [`tabs.Tab`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab) is a good alternative in for the `browser.tabs` namespace.

Currently we could check if a stored `tabId` is now focused in two ways:

```js
const tabId = 1;
const focusedTab = await browser.tabs.query({active: true, lastFocusedWindow: true})
if(focusedTab.id === tabId) {
	const tab = await browser.tabs.get(tabId);
	// tab...
}
```

or

```js
const tabId = 1;
const tab = await browser.tabs.get(tabId);
const win = await browser.windows.get(tab.windowId);
if (win.focused && tab.active) {
	// tab...
}
```

They're all a bit troublesome, and that doesn't include error handling.

Why can't we directly:

```js
const tabId = 1;
const tab = await browser.tabs.get(tabId);
if (tab.focused) {
	// tab...
}
```

More to the point, it solves the problems I mentioned above without API breaking changes: https://github.com/w3c/webextensions/pull/817#issuecomment-2828746041

dotproto:
> Note that `tabs.focus()` should still probably be avoided because it is not clear which `tab` is, and `tabs` does not represent the current `page` part. If the double-column `page` is implemented in the future, the problem will be more obvious.
> 
> For this, I still think that using `focus` property in [`updateProperties`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#updateproperties) is a viable alternative in the `tabs` namespace, which is more in line with current API logic. [#817 (comment)](https://github.com/w3c/webextensions/pull/817#issuecomment-2828746041)

If we're going for set of API changes and only introducing a `focus()` method on the `sidebarAction`/`sidePanel` namespace, then I can see `tabs.update()` + a new property on `updateProperties` making sense. I think introducing `tabs.focus()` makes more sense once we have 2 or more other namespaces with focus methods. At that point we start having a consistency issue where focusing a specific document in a tab uses a different pattern than everything else. 

At the moment, I think the path we take for the first revision is contingent on how likely we are to either introduce a `focus()` method on at least one other namespace APIs in the next ~3-5 years or want to expose more advanced focus control for a given tab (target sub-frames or another document in the same tab). If that seems reasonably likely, then we should lean into the pattern now. If it seems less than moderately, then we should tweak the `tabs.update()` method to better serve the known use cases.

> More to the point, it solves the problems I mentioned above without API breaking changes: [#817 (comment)](https://github.com/w3c/webextensions/pull/817#issuecomment-2828746041)

Sorry, I didn't follow this comment. How does introducing a `focus()` method break create breaking changes?

qupig:
@dotproto Thanks for the reply.

> a _`focus`_ property in [`updateProperties`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/update#updateproperties)

First of all, I'm sorry for the point I wrote in all the comments above. I should use `focused` instead of `focus` property for consistency with the existing properties in the `browser.windows` namespace.

> I think introducing `tabs.focus()` makes more sense once we have 2 or more other namespaces with focus methods. At that point we start having a consistency issue where focusing a specific document in a tab uses a different pattern than everything else.

I admit this creates inconsistencies with other namespaces, but this is due to the complexity of the `browser.tabs` APIs itself. And we've already done this in the [`browser.windows`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/update#updateinfo) namespace:

```js
await browser.windows.update( windowId, { focused: true } );
```

So do we also need to implement `browser.windows.focus()` to replace it to maintain consistency? (This would be a potential breaking change)

I think there is inconsistency of `tabs` and `windows` because of the current demand, they are already more complex than other namespaces.

If other namespaces have similar complexity and have appropriate places to place, I think maybe similar tweaks make sense, for example I've seen that `firefox` now has a double column `sidebar` (vertical tabs and extension panel), and to be honest I'm not sure how it will handle focus.

> or want to expose more advanced focus control for a given tab (target sub-frames or another document in the same tab). If that seems reasonably likely, then we should lean into the pattern now. If it seems less than moderately, then we should tweak the `tabs.update()` method to better serve the known use cases.

In fact, if we want to use `tabs.focus()` for more complex purposes, we will have consistency issue anyway. Because for other namespaces, we are not considering any complex use cases at the moment.

More importantly, if we want to consider more meticulous focus management rather than just switching focus between different blocks (tab, sidebar, etc.), that is, if we want to consider focus management inside the block (like frames or even elements you said), I strongly believe that we should use the brand new namespace `browser.focus` to specialize in all focus management tasks. It fits your original proposal, just not under the `tabs` namespace anymore.

Another point is that I'm not sure if a use case like focus sub-frames is currently something to consider or necessary. 
Because I think once we get the `tab` block focused, we could always control the focus within the `tab/page` by executing the [content scripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts). (e.g. `iframe.contentWindow.focus();`)

I think the current focus is mainly to switch `focus` between different blocks (tab, sidebar, etc.), which is something we currently lack. It is something we cannot do at the moment.

In other words, I think what we need to urgently address at this point should just be refining the current APIs to provide a way to switch focus between blocks (tab, sidebar, etc.), whether it's `_namesapce_.focus()` or `tabs.update(tabId, {focused: true})`.

In fact I don't even mind having both `tabs.focus()` and `tabs.update(tabId, {focused: false})`, these are all mainly solving the issues existing in the current APIs.

After this is done, we can consider using the new namespace `browser.focus` to implement any advanced and unified focus management and gradually replace existing APIs in the future, and achieve all the desired consistency, which will be a huge proposal that requires more consideration. Unless you want to do it now, I don't think it's realistic at the moment. I think advanced focus management may be demanding in the future, such as the development of `MCP`. It should be in the brand new namespace and requires a separate **permission**.

> Sorry, I didn't follow this comment. How does introducing a `focus()` method break create breaking changes?

Still the use case I mentioned above, how do we use `tabs.focus()` to prevent the now-existing force focus switch to the `tab` block in `tabs.update()`? How do we prevent the call to `tabs.update()` from making sure the `focus` remains at `sidebar` instead of doing one or more `focus` switches? I can't see how to do this without doing breaking changes besides from adding new properties in `updateProperties`, this is the easiest way to solve the current problem I think.

PR #826, opened 2025-05-16
Proposal: Allow extensions to track their own disable/uninstall events without "management" permission

jaissam10:
Proposal: Allow extensions to track their own disable/uninstall events without "management" permission

### DIFF
--- a/proposals/proposal-extension-self-management-events.md
+++ b/proposals/proposal-extension-self-management-events.md
@@ -8,7 +8,7 @@ Currently, extensions cannot track when users disable or uninstall them without 
 
 **Author:** [sameerjaiswal](https://github.com/jaissam10)
 
-**Sponsoring Browser:** Chrome
+**Sponsoring Browser:** TBD
 
 **Contributors:** 
 
--- a/proposals/proposal-extension-self-management-events.md
+++ b/proposals/proposal-extension-self-management-events.md
@@ -1,0 +1,168 @@
+# Proposal: Allow extensions to track their own disable/uninstall events without "management" permission
+
+**Summary**
+
+Currently, extensions cannot track when users disable or uninstall them without requesting the broad "management" permission. This proposal suggests modifying the behavior of `chrome.management.onDisabled` and `chrome.management.onUninstalled` events to allow extensions to track their own state changes without requiring the "management" permission.
+
+**Document Metadata**
+
+**Author:** [sameerjaiswal](https://github.com/jaissam10)
+
+**Sponsoring Browser:** Chrome
+
+**Contributors:** 
+
+**Created:** 2025-05-16
+
+**Related Issues:** https://issues.chromium.org/issues/417868528
+
+## Motivation
+Extensions need to track when users disable or uninstall them for legitimate purposes like:
+- Analytics and user behavior tracking
+- Understanding user retention
+- Improving extension quality based on user behavior
+
+Currently, this requires the broad "management" permission, which is inconsistent with other self-reference methods like `management.getSelf()` that don't require permissions.
+### Objective
+To enable extensions to track their own disable/uninstall events without requiring the broad "management" permission, while maintaining the existing permission model for tracking other extensions' events.
+
+### Use Cases
+1. **Analytics Tracking**
+   - Track when users disable/uninstall the extension
+   - Understand user retention patterns
+   - Measure impact of updates on user behavior
+2. **User Feedback**
+   - Trigger feedback surveys when users disable/uninstall
+   - Collect reasons and improve extension based on user feedback
+3. **State Management**
+   - Clean up extension data when disabled/uninstalled
+   - Handle graceful shutdown of extension services
+   - Manage user preferences and settings
+
+### Known Consumers
+- Adobe Acrobat Extension
+- Other extensions that need to track user behavior and retention
+
+## Specification
+
+### Schema
+### Current Behavior
+- `chrome.management.onDisabled` and `chrome.management.onUninstalled` require "management" permission
+- This permission is meant for managing other extensions
+- Extensions cannot track their own state changes without this broad permission
+
+### Proposed Changes
+Two implementation approaches are proposed:
+
+#### Option 1: Single API with Permission-based Behavior
+- Keep existing `onDisabled` and `onUninstalled` events
+- If "management" permission is present: trigger for all extensions
+- If no permission: only trigger for the calling extension
+- Maintains backward compatibility
+
+Example:
+```javascript
+// Without permission - only triggers for self
+chrome.management.onDisabled.addListener((info) => {
+  if (info.id === chrome.runtime.id) {
+    // Handle self disable and this will get called without permission
+  }
+});
+
+// With permission - triggers for all extensions
+chrome.management.onDisabled.addListener((info) => {
+  // Handle any extension disable if anything generic needs to be done but that needs permision
+});
+```
+
+#### Option 2: Separate API Methods
+- Add new methods specifically for self-events:
+  - `chrome.management.onSelfDisabled`
+  - `chrome.management.onSelfUninstalled`
+- Keep existing methods for all-extension events (requiring permission)
+- Clearer API separation
+
+Example:
+```javascript
+// New methods for self-events (no permission needed)
+chrome.management.onSelfDisabled.addListener(() => {
+  // Handle self disable
+});
+
+// Existing methods (requires permission)
+chrome.management.onDisabled.addListener((info) => {
+  // Handle any extension disable
+});
+```
+
+
+### Backward Compatibility
+- Option 1 maintains full backward compatibility
+- Option 2 requires new API methods but doesn't break existing code
+
+## Examples
+### Current Usage
+```javascript
+// Requires "management" permission
+chrome.management.onDisabled.addListener((info) => {
+  if (info.id === chrome.runtime.id) {
+    // Handle self disable
+  }
+});
+```
+
+### Proposed Usage (Option 1)
+```javascript
+// No permission needed for self-events
+chrome.management.onDisabled.addListener((info) => {
+  if (info.id === chrome.runtime.id) {
+    // Handle self disable
+  }
+});
+```
+
+### Proposed Usage (Option 2)
+```javascript
+// New dedicated method for self-events
+chrome.management.onSelfDisabled.addListener(() => {
+  // Handle self disable
+});
+```
+
+### New Permissions
+No new permissions are required.
+
+### Manifest File Changes
+No new manifest fields are required.
+
+## Security and Privacy
+### Exposed Sensitive Data
+- No sensitive data is exposed
+- Only the extension's own state changes are tracked
+- No access to other extensions' information without permission
+
+### Abuse Mitigations
+- Self-events only trigger for the calling extension
+- No ability to track other extensions without permission
+- Consistent with existing security model
+
+
+### Additional Security Considerations
+
+
+## Alternatives
+### Existing Workarounds
+
+### Open Web API
+No equivalent functionality exists in the Open Web API.
+
+## Implementation Notes
+
+- No changes to existing permission model
+- Clear documentation needed for behavior differences
+- Implementation should be consistent across browsers
+
+## Future Work
+1. Consider adding reason codes for disable/uninstall events
+2. Explore adding similar functionality for other self-reference events
+
--- a/proposals/proposal-extension-self-management-events.md
+++ b/proposals/proposal-extension-self-management-events.md
@@ -1,8 +1,12 @@
-# Proposal: Allow extensions to track their own disable/uninstall events without "management" permission
+# Proposal: Allow extensions to track their own disable/uninstall events
 
 **Summary**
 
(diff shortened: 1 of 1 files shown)


fregante:
Duplicate of https://github.com/w3c/webextensions/pull/792?

jaissam10:
> Duplicate of #792?

No, that looks different.

fregante:
I don’t think so. The management API is only available in chrome-extension:// pages, all of which are immediately unloaded the moment the user disables or uninstalls the extension. 

So that leaves only content scripts behind. That event is the "context invalidated" event with `reason: disabled` or `reason: uninstalled`

jaissam10:
@fregante
Got it.  
I reviewed the proposal again and it seems like it also accounts for uninstall and disable scenarios, as well as cases like Local File Access Toggle, updates, and transitions where the service worker is auto-switched from off to on, browser restart.

**Just to confirm—this API can be used within the main `service_worker` file, including usage from content scripts, correct? And there can be cases where no content script is there and we need these events.**  
From what I understand, the method I was suggesting would only work in the `background_script`, as the `management` API isn’t available in content scripts.

That said, Proposal #792 appears to handle this in a more comprehensive and scalable way if it is available in background scipt too.

jaissam10:
@fregante Can you please confirm that will this be available in background scripts too (like main service_worker file)?

fregante:
As mentioned before, by the time "on uninstalled" is fired, the background script has been unloaded already, regardless of which API "wins".

You can request a specification of such behavior in the linked PR. The most you could hope for is support for `navigator.sendBeacon()` in this event, which probably satisfies your needs.

_Side note: I'm just a commenter here, I'm not a w3c contributor, my comments are "IMHO"._

dotproto:
@jaissam10, as a quick process note, we typically discuss concerns in an issue before creating a proposal for a specific solution. That process helps us align on use cases, engineering concerns, and the general shape of a potential solution. From there we create proposal PRs that flesh out a proposal in detail.

The ipr bot has [flagged this PR](https://labs.w3.org/repo-manager/pr/id/w3c/webextensions/826) because it can't detect that you've agreed to the W3C's [intellectual property rights](https://www.w3.org/copyright/intellectual-rights/) agreement. Before we can merge a PR, you will need to [join](https://www.w3.org/community/webextensions/join) our group and [link your GitHub and W3C accounts](https://www.w3.org/users/myprofile/connectedaccounts).

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Related Issues:** https://issues.chromium.org/issues/417868528
+
+## Motivation
+Extensions need to track when users disable or uninstall them for legitimate purposes like:
+- Analytics and user behavior tracking
+- Understanding user retention
+- Improving extension quality based on user behavior
+
+Currently, this requires the broad "management" permission, which is inconsistent with other self-reference methods like `management.getSelf()` that don't require permissions.
```
dotproto: For clarity, are you saying that `management.onUninstalled` fires inside an extension that is uninstalled? If so, that sounds like a bug to me. I would expect that event to fire after an uninstall occurs;  by definition the extension wouldn't be installed to receive it's on uninstalled event.

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Summary**
+
+Currently, extensions cannot track when users disable or uninstall them without requesting the broad "management" permission. This proposal suggests modifying the behavior of `chrome.management.onDisabled` and `chrome.management.onUninstalled` events to allow extensions to track their own state changes without requiring the "management" permission.
+
+**Document Metadata**
+
+**Author:** [sameerjaiswal](https://github.com/jaissam10)
+
+**Sponsoring Browser:** Chrome
```
dotproto: Change requested: I don't see a comment from a Chrome rep on this PR or in [the original Chromium issue](https://issues.chromium.org/issues/417868528) indicating that they intend to implement this proposal. Until we have a signal like that, it would be best to mark this field "TBD".

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
...
+To enable extensions to track their own disable/uninstall events without requiring the broad "management" permission, while maintaining the existing permission model for tracking other extensions' events.
+
+### Use Cases
+1. **Analytics Tracking**
+   - Track when users disable/uninstall the extension
+   - Understand user retention patterns
+   - Measure impact of updates on user behavior
+2. **User Feedback**
+   - Trigger feedback surveys when users disable/uninstall
+   - Collect reasons and improve extension based on user feedback
```
dotproto: As far as I know this is the primary use case for [`runtime.setUninstallURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL).

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
...
+3. **State Management**
+   - Clean up extension data when disabled/uninstalled
+   - Handle graceful shutdown of extension services
+   - Manage user preferences and settings
```
dotproto: I don't understand this set of use cases. An extension's local state is only useful to the extension itself. When the extension is removed, so too is it's state. There's no need for the extension to clean up anything. 

As for remote state, extensions should use the same general approach as you would for web development, where page sessions are ephemeral.

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
...
+
+Currently, this requires the broad "management" permission, which is inconsistent with other self-reference methods like `management.getSelf()` that don't require permissions.
+### Objective
+To enable extensions to track their own disable/uninstall events without requiring the broad "management" permission, while maintaining the existing permission model for tracking other extensions' events.
+
+### Use Cases
+1. **Analytics Tracking**
+   - Track when users disable/uninstall the extension
+   - Understand user retention patterns
+   - Measure impact of updates on user behavior
```
dotproto: I believe these are secondary use cases for [`runtime.setUninstallURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL).

proposals/proposal-extension-self-management-events.md
```diff
@@ -0,0 +1,168 @@
+# Proposal: Allow extensions to track their own disable/uninstall events without "management" permission
+
+**Summary**
+
+Currently, extensions cannot track when users disable or uninstall them without requesting the broad "management" permission. This proposal suggests modifying the behavior of `chrome.management.onDisabled` and `chrome.management.onUninstalled` events to allow extensions to track their own state changes without requiring the "management" permission.
```
dotproto: Change request: We prefer to use the browser neutral `browser` global namespace rather than the vendor-specific `chrome` namespace where reasonable.

dotproto (changes requested):

dotproto:
We discussed this PR in today's WECG meeting. Consensus among attendees was that [`runtime.setUninstallURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL) should address the use cases in this proposal as we understand them. Barring clarification on problems this proposal is seeking to address, I'm inclined to close this PR.

proposals/proposal-extension-self-management-events.md
jaissam10: Thanks for pointing out, have updated it to "TBD"

proposals/proposal-extension-self-management-events.md
jaissam10: I understand that `setUninstallURL` can be used for this purpose, and we are already using it. However, there are additional scenarios to consider—such as when a user disables the extension and later uninstalls it. We'd like to be able to track those users as well.  
Unfortunately, there’s currently no way for us to capture data when a user simply disables the extension.

proposals/proposal-extension-self-management-events.md
jaissam10: Currently, the only available mechanism is `setUninstallURL`, which allows registering a page to collect user feedback or send analytics to internal servers for tracking extension activity.

However, this solution doesn't fully address a few scenarios:

1. When the extension is disabled  (not uninstalled)
2. When it is uninstalled after being disabled  
3. In the case of admin-installed extensions, opening a feedback page may not be appropriate. Since the installation isn't user-initiated, triggering the page for all users at once could be disruptive or confusing.

proposals/proposal-extension-self-management-events.md
jaissam10: Updated, thanks.

proposals/proposal-extension-self-management-events.md
jaissam10: Thanks for pointing out, my initial assumption was it is for tracking all extensions including self. But i rechecked and its only for other extensions. I've updated the proposal.

proposals/proposal-extension-self-management-events.md
jaissam10: Updated. 
I was considering the scenario where a DNR (Declarative Net Request) rule is registered. In such cases, does the rule also get removed when the extension is disabled or uninstalled?

jaissam10:
> We discussed this PR in today's WECG meeting. Consensus among attendees was that [`runtime.setUninstallURL()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL) should address the use cases in this proposal as we understand them. Barring clarification on problems this proposal is seeking to address, I'm inclined to close this PR.

The current `setUninstallURL` method only partially addresses this need:
- It works only during direct uninstall events
- Cannot silently log analytics without opening a page
- It does not support disable events
- It is not ideal for enterprise/admin-installed extensions, where showing a feedback page can feel intrusive or confusing as its uninstalled by admin.

proposals/proposal-extension-self-management-events.md
dotproto: Yes, DNR rules are tied to the enabled and installation state of the extension, as are event listeners, browser settings, content script registrations, etc.

proposals/proposal-extension-self-management-events.md
dotproto: My read on each of these scenarios is rather different, so I'll go through each in turn.

> 1. When the extension is disabled (not uninstalled)

Based on our discussion in the last public meeting, I don't think there's much interest in exposing disabled/enabled events for an extension to that extension. A disabled event suffers from the same lifetime concerns as uninstallation. An enabled event doesn't have that issue, but I don't see that getting traction unless it addresses an end user need or solves a platform design issue.

From my point of view, the core issue is that data collection in itself is not a compelling use case. A user disabling an extension is one of many different reasons why an extension may go unused. Others include the device's battery ran out, the user is on vacation, the user switched web browsers, the user uses different profiles (or OS level accounts) for different tasks, the user "power washed" their computer and cleared out their old browser profile, the user happened to stay in a private browsing session for longer than usual, the user is testing multiple browser extensions, the user upgraded their computer, etc. Any of these could lead to an extended gap in usage akin to the extension being disabled.

Given that disabling an extension is a small subset of these cases, my question then is what problem would having enabled/disabled events solve? Right now I don't see a demonstrable benefit to adding them. My impression is that you're better off using other signals to project user engagement and product stability. 

> 2. When it is uninstalled after being disabled

This feels like something we should probably discuss in the next public meeting. Since we already have `setUninstallURL()`, I think there's a solid argument to make that the last uninstall URL set should be opened when an extension uninstalled, even if the extension is disabled. I'd recommend adding an agenda topic suggestion to https://github.com/w3c/webextensions/issues/835 to discuss this specific case in our next meeting.

> 3. In the case of admin-installed extensions, opening a feedback page may not be appropriate. Since the installation isn't user-initiated, triggering the page for all users at once could be disruptive or confusing.

For clarity, is this a practical or hypothetical concern?

I don't know how Chrome and Firefox handle this offhand, but my expectation is that if an extension is uninstalled in response to an administrator action, the extension's uninstall URL should not open in a new tab when the extension is uninstalled. My rationale here is that the uninstall URL is a way for the extension to get feedback from the end user about why they chose to part ways with the extension. Since this the removal did not occur in response to a user action, they can't meaningfully provide that feedback. This might also be good to raise as an agenda item for discussion purposes.

jaissam10:
Hey @dotproto ,
Can you please revisit the proposal and take another look at the points I raised, particularly those that aren’t addressed by `setUninstallURL`.
Would it be possible for us to review it again?

dotproto:
I meant to close this [last week](https://github.com/w3c/webextensions/pull/826#issuecomment-2901792349) as browser vendors raised concerns about this specific proposal during our last public meeting. @jaissam10, if you're interested in continuing to explore the concerns that motivated this proposal, please create an issue for discussion purposes.

jaissam10:
> I meant to close this [last week](https://github.com/w3c/webextensions/pull/826#issuecomment-2901792349) as browser vendors raised concerns about this specific proposal during our last public meeting. @jaissam10, if you're interested in continuing to explore the concerns that motivated this proposal, please create an issue for discussion purposes.

Sure, I've created an [issue](https://github.com/w3c/webextensions/issues/836) for the discussion.
I created the proposal directly without raising an issue here as I've been asked to create proposal in the Chromium bug comment.

Issue #521, opened 2024-01-12
`sidePanel` API: `sidePanel.close()` and `sidePanel.toggle()`


fregante:
The sidePanel can be opened via `chrome.sidePanel.open()` but the API isn't symmetrical, there is no `.close()` API.

It would also be useful (maybe even more so) to just `.toggle()` the sidebar instead, using any trigger other than the action click, for example:

- `menus.onClick`
- any `<button>` added on the page
- keyboard shortcuts
- command palettes in the content script

## Alternative to `chrome.sidePanel.close()` (✅  it works)

As [@dotproto pointed out](https://github.com/w3c/webextensions/issues/515#issuecomment-1887701389), the sidebar can close itself via `window.close()`, so in other contexts you would have to set up messaging.

```js
// background.js
function open () {
  chrome.sidePanel.open()
}
function close() {
  chrome.runtime.sendMessage('closeSidePanel');
}
```
```js
// sidepanel.js
chrome.runtime.onMessage.addListener(message => {
  // Might not be as easy if there are multiple side panels open
  if (message === 'closeSidePanel') {
    window.close();
  }
})
```

You can also use `chrome.sidePanel.setOptions({enabled: false})` but that has undesired side effects (it removes it from the side panels list)

## Alternative to `chrome.sidePanel.toggle()` (⛔️ it doesn't work)

This is more complicated because there's no easy way to determine whether the sidebar is open, so you'd have to add another listener and hope that this delay won't cause you to lose the user gesture permission:

> [!warning]
> I just tested this and it doesn't work: <code>Error: `sidePanel.open()` may only be called in response to a user gesture</code>

```js
function isSidePanelOpen() {
  try {
    return await chrome.runtime.sendMessage('isSidePanelOpen');
  } catch {
    return false
  }
}
async function toggle() {
  if (await isSidePanelOpen()) {
    chrome.runtime.sendMessage('closeSidePanel');
  } else {
    chrome.sidePanel.open()
  }
}
```

```js
// sidepanel.js
chrome.runtime.onMessage.addListener(message => {
  // Might not be as easy if there are multiple side panels open
  if (message === 'isSidePanelOpen') {
    return true
  }

  if (message === 'closeSidePanel') {
    window.close();
  }
})
```

sublimator:
The sidePanel api is a bit weird really! 
[...]

Issue #128, opened 2021-11-01
Proposal: agree on a unified `sidebar` API for mv3


carlosjeurissen:
Since `sidebar_action` is not implemented in Chromium we have the opportunity to still define what it should look like in mv3.

In mv2, this has been implemented by Mozilla Firefox, Opera and Naver Whale. Yet they are different across browsers. See:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/sidebar_action
https://developers.whale.naver.com/api/extensions/sidebarAction/
https://dev.opera.com/extensions/sidebar-action-api/

One of the differences is the support of `sidebar_action` next to another action in Firefox and Opera, but not in Naver Whale. Another is the lack of `open_at_install` in Opera. And Firefox uses `default_panel` while Naver Whale uses `default_page`.

Going forward, we can define a merged API covering the aspects in each of these browsers. Have a meeting about it in which we invite both Opera, Naver Whale and potentially Vivaldi (they have non-extension web panels) to discuss directions and come to an agreement.

aam1r:
Hi there,

We recently shipped a new feature in our Chrome Extension leveraging the Side Panel API. 

Initially, we were confused by the fact that this functionality seemed exclusive to Chrome Extensions. Our searches, such as "side panel Firefox," didn't yield the expected results. However, after discovering this issue, we realized that similar capabilities exist in other browsers but under different names.

Given this context, I'm curious if there are any plans to standardize the terminology used across different browsers for this feature. Harmonizing the nomenclature could bring several benefits:

- **Discoverability:** Enhanced discoverability of the API which could lead to broader adoption. Our team's experience is a testament to the challenges posed by inconsistent naming.
- **Simpler Documentation:** Simplified cross-browser documentation comparison, enabling developers to quickly understand what's supported in each browser.
- **Better DX:** Easier comparison of API functionalities for Sidebars, making development and planning easier.

Thank you

carlosjeurissen:
@Rob--W following the discussion in https://github.com/w3c/webextensions/issues/884. It seems the `sidePanel.setBehavior({openPanelOnActionClick})` would better fit the `action` namespace.

oleksiilevzhynskyi:
Hey. Reposting "call for unifying" sidepanel api from "WECG at TPAC 2025" topic  - https://github.com/w3c/webextensions/issues/843#issuecomment-3507735922

[...]

PR #837, opened 2025-05-28
Proposal: close() Method for the sidePanel API

hharshas:
This proposal introduces a sidePanel.close() method (among work items for [Summer of Code](https://summerofcode.withgoogle.com/)).
While not fully fleshed out, it outlines the core idea. Initial feedback is welcome. I would also like your thoughts on the need for a user gesture for this. #521 [#chromium](https://issues.chromium.org/issues/403765214)

### DIFF
--- a/proposals/sidepanel-close-api.md
+++ b/proposals/sidepanel-close-api.md
@@ -92,9 +92,10 @@ namespace sidePanel {
 - If `windowId` or `tabId` is invalid, rejects with an error.
 - If both `windowId` and `tabId` are provided, the method will verify
   that the tab belongs to the specified window. If not, it rejects with an error.
-- If a side panel is opened globally for a given `windowId` and the method is
-  called with both `windowId` and `tabId`, the side panel will be closed
-  for all tabs in that window, not just the specified tab.
+- If both `windowId` and `tabId` are provided:
+    1. If a contextual side panel is visible on the specified `tabId`, it will be closed.
+    2. If a global side panel is visible on the given `tabId`,
+       it will be closed for all tabs in that window, not just the specified tab.
 
 ### New Permissions
 
--- a/proposals/sidepanel-close-api.md
+++ b/proposals/sidepanel-close-api.md
@@ -25,7 +25,8 @@ What does this API enable?
 - sidePanel.close() method will allow the sidebar to close programmatically.
 
 Why do we need it?
-- Currently, extensions cannot directly close their own sidebar from the background script, 
+- Currently, extensions cannot directly close their own sidebar from the
+ background script, 
 developers have to:
   1. Send a message from a background script to the side panel.
   2. Call window.close() from the side panel window on message delivery.
@@ -35,11 +36,14 @@ This event aims to give developers an easier control.
 #### Use Cases
 
 Programmatic Control: 
-- An extension could automatically close the panel after completing an action or when certain conditions are met. e.g. Closing the sidePanel after longer inactivity.
+- An extension could automatically close the panel after completing an action or
+ when certain conditions are met. e.g. Closing the sidePanel after longer
+  inactivity.
 
 ### Known Consumers
 
-[The Chromium bug](https://issues.chromium.org/issues/403765214) has a significant amount of developer interest and the disscussions on the issue shows the same.
+[The Chromium bug](https://issues.chromium.org/issues/403765214) has a significant
+ amount of developer interest and the discussions on the issue shows the same.
 
 ## Specification
 
@@ -63,9 +67,9 @@ namespace sidePanel {
     // `options`: Specifies the context in which to close the side panel.
     // `callback`: Called when the side panel has been closed.
     static void close(
-        CloseOptions options,
-        optional VoidCallback callback);
-    };
+      CloseOptions options,
+      optional VoidCallback callback);
+  };
 };
 ```
 
@@ -73,14 +77,18 @@ namespace sidePanel {
 
 - Type: Promise<void>
 
-- Resolves when the panel has been closed. If the panel is already closed, still resolves successfully.
+- Resolves when the panel has been closed. If the panel is 
+already closed, still resolves successfully.
 
 ### Behavior
-- The operation will only close side panels that belong to the calling extension (extension ID must match).
-- If the panel is already closed or does not exist in the given context, the method does nothing.
+- The operation will only close side panels that belong to the 
+calling extension (extension ID must match).
+- If the panel is already closed or does not exist in the given 
+context, the method does nothing.
 - If neither windowId nor tabId is provided, rejects with an error 
 - If windowId or tabId is invalid, rejects with an error.
-- If both windowId and tabId are provided, the method will verify that the tab belongs to the specified window. If not, it rejects with an error.
+- If both windowId and tabId are provided, the method will verify 
+that the tab belongs to the specified window. If not, it rejects with an error.
 ### New Permissions
 N/A.
 
@@ -110,7 +118,8 @@ N/A
 
 ### Open Web API
 
-Not applicable to the open web as it only complements the existing `open()` functionality.
+Not applicable to the open web as it only complements the existing `open()` 
+functionality.
 
 ## Implementation Notes
 
@@ -118,4 +127,6 @@ N/A.
 
 ## Future Work
 
-This implementation will align with [#779](https://github.com/w3c/webextensions/pull/779), to ensure proper event handling on panel closure. (Open for discussion)
+This implementation will align with 
+[#779](https://github.com/w3c/webextensions/pull/779), to ensure proper event
+ handling on panel closure. (Open for discussion)
--- a/proposals/sidepanel-close-api.md
+++ b/proposals/sidepanel-close-api.md
@@ -30,6 +30,9 @@ Why do we need it?
   background script, developers have to:
   1. Send a message from a background script to the side panel.
   2. Call `window.close()` from the side panel window on message delivery.
+- Further, this method has a big problem: when `window.close()`
+  is used in a tab-specific side panel, it also closes any global side panels
+  that might be open on other tabs.
 
 This event aims to give developers an easier control.
 
--- a/proposals/sidepanel-close-api.md
+++ b/proposals/sidepanel-close-api.md
@@ -47,13 +47,14 @@ The Chromium bug has a significant amount of developer interest and the disscuss
 ```typescript
 namespace sidePanel {
   dictionary CloseOptions {
-    // The window in which to close the side panel. If omitted,
-    // operates on the current window.
+    // The window in which to close the side panel.
     long? windowId;
 
     // The tab for which to close the side panel. If specified,
     // only closes tab-specific side panels.
     long? tabId;
+
+    // At least one of these must be provided, otherwise rejects with an error
   };
 
   interface Functions {
@@ -76,9 +77,9 @@ namespace sidePanel {
 ### Behavior
     - The operation will only close side panels that belong to the calling extension (extension ID must match).
 
-    - If called with no options, closes the panel in the current window.
-
     - If the panel is already closed or does not exist in the given context, the method does nothing.
+
+    - If neither windowId nor tabId is provided, rejects with an error 
 
     - If windowId or tabId is invalid, rejects with an error.
 
--- a/proposals/sidepanel-close-api.md
+++ b/proposals/sidepanel-close-api.md
@@ -1,6 +1,7 @@
 # New close() Method for the sidePanel API
 
 **Summary**
+
 Proposal to introduce a programmatic close method for the side panel API.
 
 **Document Metadata**
@@ -9,27 +10,26 @@ Proposal to introduce a programmatic close method for the side panel API.
 
 **Sponsoring Browser:** Chromium
 
-**Contributors:** [hharshas](https://github.com/hharshas) (
-  [Summer Of Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
+**Contributors:** [hharshas](https://github.com/hharshas) ([Summer Of
+Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
 
 **Created:** 2025-05-27
 
 **Related Issues:** [#521](https://github.com/w3c/webextensions/issues/521),
- [#chromium](https://issues.chromium.org/issues/403765214)
+[#chromium](https://issues.chromium.org/issues/403765214)
 
 ## Motivation
 
 ### Objective
 
 What does this API enable?
-- sidePanel.close() method will allow the sidebar to close programmatically.
+- `sidePanel.close()` method will allow the sidebar to close programmatically.
 
 Why do we need it?
-- Currently, extensions cannot directly close their own sidebar from the
- background script, 
-developers have to:
+- Currently, extensions cannot directly close their own side panel from the
+  background script, developers have to:
   1. Send a message from a background script to the side panel.
-  2. Call window.close() from the side panel window on message delivery.
+  2. Call `window.close()` from the side panel window on message delivery.
 
 This event aims to give developers an easier control.
 
@@ -37,13 +37,13 @@ This event aims to give developers an easier control.
 
 Programmatic Control: 
 - An extension could automatically close the panel after completing an action or
- when certain conditions are met. e.g. Closing the sidePanel after longer
+  when certain conditions are met. For example, closing the side panel after longer
   inactivity.
 
 ### Known Consumers
 
 [The Chromium bug](https://issues.chromium.org/issues/403765214) has a significant
- amount of developer interest and the discussions on the issue shows the same.
+amount of developer interest and the discussions on the issue shows the same.
 
 ## Specification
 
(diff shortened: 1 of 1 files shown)


proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+  2. Call window.close() from the side panel window on message delivery.
+
+This event aims to give developers an easier control.
+
+#### Use Cases
+
+Programmatic Control: 
+- An extension could automatically close the panel after completing an action or
+ when certain conditions are met. e.g. Closing the sidePanel after longer
+  inactivity.
```
dotproto: Two minor issues with this block:

1. Several tech style guides recommend against using "e.g." in place of "for example" ([Apple](https://support.apple.com/guide/applestyleguide/a-apsg3acde405/1.0/web/1.0#apde9a26a57d4df4), [Google](https://developers.google.com/style/abbreviations#dont-use), [Microsoft](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg)). While this group does not follow a specific style guide, it's a good practice to be aware of when writing technical content.

2. This line (and other parts of the document) use inconsistent spacing prefixes. This can cause issues as Markdown parsers do not necessarily handle single line breaks and varying indentation levels consistently. It's a bit long, but [this section](https://github.github.com/gfm/#motivation) of the GFM spec discusses some related concerns.

```suggestion
- An extension could automatically close the panel after completing an action or
  when certain conditions are met. For example, closing the sidePanel after longer
  inactivity.
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Known Consumers
+
+[The Chromium bug](https://issues.chromium.org/issues/403765214) has a significant
+ amount of developer interest and the discussions on the issue shows the same.
```
dotproto: Minor: Trim leading white space.

```suggestion
[The Chromium bug](https://issues.chromium.org/issues/403765214) has a significant
amount of developer interest and the discussions on the issue shows the same.
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
+# New close() Method for the sidePanel API
+
+**Summary**
+Proposal to introduce a programmatic close method for the side panel API.
```
dotproto: Add a new line after `**Summary**` to prevent the sentance that follows from rendering on the same line. This rendering issue is visible [here](https://github.com/hharshas/webextensions/blob/255679a4b0073b51fe47cfe675d743b4f8c4da79/proposals/sidepanel-close-api.md).
```suggestion
**Summary**

Proposal to introduce a programmatic close method for the side panel API.
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Document Metadata**
+
+**Author:** [hharshas](https://github.com/hharshas)
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** [hharshas](https://github.com/hharshas) (
+  [Summer Of Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
```
dotproto: Placing the opening parentheses (`(`) on one line and the content inside it on the next inserts a space between the the parentheses and its content. (Visible [here](https://github.com/hharshas/webextensions/blob/255679a4b0073b51fe47cfe675d743b4f8c4da79/proposals/sidepanel-close-api.md))

```suggestion
**Contributors:** [hharshas](https://github.com/hharshas) ([Summer Of
Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Created:** 2025-05-27
+
+**Related Issues:** [#521](https://github.com/w3c/webextensions/issues/521),
+ [#chromium](https://issues.chromium.org/issues/403765214)
```
dotproto: Minor: Trim leading space.

```suggestion
**Related Issues:** [#521](https://github.com/w3c/webextensions/issues/521),
[#chromium](https://issues.chromium.org/issues/403765214)
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+
+What does this API enable?
+- sidePanel.close() method will allow the sidebar to close programmatically.
+
+Why do we need it?
+- Currently, extensions cannot directly close their own sidebar from the
+ background script, 
+developers have to:
+  1. Send a message from a background script to the side panel.
+  2. Call window.close() from the side panel window on message delivery.
```
dotproto: Fixes a rendering bug where "developers have to:" appeared as a paragraph separating the unordered list that precedes it from the ordered list that follows.

Also styles "window.close()" as monospace to signal this text is a snippet of code.

```suggestion
Why do we need it?
- Currently, extensions cannot directly close their own sidebar from the
  background script, developers have to:
  1. Send a message from a background script to the side panel.
  2. Call `window.close()` from the side panel window on message delivery.
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+context, the method does nothing.
+- If neither windowId nor tabId is provided, rejects with an error 
+- If windowId or tabId is invalid, rejects with an error.
+- If both windowId and tabId are provided, the method will verify 
+that the tab belongs to the specified window. If not, it rejects with an error.
+### New Permissions
+N/A.
+
+### Manifest File Changes
+N/A
```
dotproto: This suggestion addresses a few inconsistencies.

- Separate headers and contentent before/after them with a single blank line.
- Terminate all list items with period for consistency.
- Use inline monospace annotations to clearly signal which words are direct code references, such as an object's properties.
- Trim trailing whitespace.
- Normalize list item wrapping so the first character of a wrapped line aligns with the first character of content in the list item. 

```suggestion
### Behavior

- The operation will only close side panels that belong to the calling extension
  (extension ID must match).
- If the panel is already closed or does not exist in the given context, the
  method does nothing.
- If neither `windowId` nor `tabId` is provided, rejects with an error.
- If `windowId` or `tabId  is invalid, rejects with an error.
- If both `windowId` and `tabId` are provided, the method will verify that the
  tab belongs to the specified window. If not, it rejects with an error.

### New Permissions

N/A

### Manifest File Changes

N/A
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+* Calling window.close() from the panel context.
+
+### Open Web API
+
+Not applicable to the open web as it only complements the existing `open()` 
+functionality.
+
+## Implementation Notes
+
+N/A.
```
dotproto: For consistency with the rest of the document, don't terminate a paragraph that only contains "N/A" with a period. 

```suggestion
N/A
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Future Work
+
+This implementation will align with 
+[#779](https://github.com/w3c/webextensions/pull/779), to ensure proper event
+ handling on panel closure. (Open for discussion)
```
dotproto: Trim leading whitespace. 

```suggestion
This implementation will align with 
[#779](https://github.com/w3c/webextensions/pull/779), to ensure proper event
handling on panel closure. (Open for discussion)
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,132 @@
...
+### Additional Security Considerations
+
+N/A
+
+## Alternatives
+
+### Existing Workarounds
+
+* Message passing between background and panel contexts.
+* Calling window.close() from the panel context.
```
dotproto: Style inline code snippet as monospace text.

```suggestion
* Calling `window.close()` from the panel context.
```

dotproto (changes requested):
All changes requested in this review are either internal styling consistency or GFM rendering fixes.

dotproto:
> I would also like your thoughts on the need for a user gesture for this.

Developers can currently close the document in the sidebar or sidepanel ("sidebar" for simplicity) by calling `window.close()` inside the sidebar document in Firefox and Chrome. This flow does not require user activation. Introducing a user activation requirement in a new `sidebar.close()` method would create an unexpected point of friction compared to the current solution. 

There are times when it might make sense to diverge or to change existing behavior to also require a user activation, but I don't think this is such a case. As I see it, there's minimal risk to allowing an extension to close it's own sidebar, especially since both supporting browsers currently require user activation to open a sidebar. The main risk I see is that an extension might try to close a side panel right before a user clicks in order to cause the user to click something they didn't intend to in the main document, but I'm not sure how practical that kind of attack would be. 

At the moment I'm comfortable with not requiring user activation, but I'll defer to @Rob--W for a more official position.

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,134 @@
...
+  2. Call `window.close()` from the side panel window on message delivery.
+
+This event aims to give developers an easier control.
+
+#### Use Cases
+
+Programmatic Control:
+- An extension could automatically close the panel after completing an action or
+  when certain conditions are met. For example, closing the side panel after longer
+  inactivity.
```
oliverdunk: I don't think inactivity is a compelling use case - could we think of another one? One that comes to mind is an extension that does some work / operates on a page and then closes when its work is finished.

oliverdunk (commented):
I left one comment, but otherwise this seems like a good proposal.

kiaraarose (approved):

hharshas:
@carlosjeurissen While preparing the proposal, I was not in favor of throwing an error when the side panel is already closed for the given context. To keep things simple, an error should only be thrown if the context is invalid. Otherwise, it should simply do nothing and resolve, indicating that the panel is already closed for that context. If the side panel is opened globally for a window and we try to close it for just one tab, it will close for all tabs in that window.

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,136 @@
...
+### Behavior
+
+- The operation will only close side panels that belong to the
+  calling extension.
+- If the panel is already closed or does not exist in the given
+  context, the method does nothing.
+- If neither `windowId` nor `tabId` is provided, rejects with an error.
+- If `windowId` or `tabId` is invalid, rejects with an error.
+- If both `windowId` and `tabId` are provided, the method will verify
+  that the tab belongs to the specified window. If not, it rejects with an error.
```
oliverdunk, line 94: Adding a thread here to make sure we add a note on the behavior when there is a global side panel open and both `windowId` and `tabId` are provided.

oliverdunk (approved):
We should make sure to align on the behavior when both `windowId` and `tabId` are specified. Otherwise, LGTM.

carlosjeurissen:
1) as mentioned in https://github.com/w3c/webextensions/pull/838#issuecomment-3032409017, it would make a ton of sense to align on a unified sidebar API as proposed in https://github.com/w3c/webextensions/issues/128

2) We should clarify the difference between hiding vs closing the sidebar and the implications.

3) As was discussed briefly in the last call, the behaviour if both `windowId` and `tabId` and if a global side panel is visible on the given `tabId`. Why should it close the sidePanel for all tabs instead of simply throwing an exception? The intention of the developer is to close it for a specific tab. Closing the global sidePanel can be an unrequested side-effect. If an exception is thrown, a developer can always call close without a tabId to get this kind of behaviour. Yet it seems this should be optional, not enforced.

hharshas:
> 2. We should clarify the difference between hiding vs closing the sidebar and the implications.

Currently, the focus is on using "close" when the intention is to remove the side panel from the tab registry entirely. The concept of "hide"—making the side panel inactive but not removing it—could be considered for future enhancements.

> 3. As was discussed briefly in the last call, the behaviour if both `windowId` and `tabId` and if a global side panel is visible on the given `tabId`. Why should it close the sidePanel for all tabs instead of simply throwing an exception? The intention of the developer is to close it for a specific tab. Closing the global sidePanel can be an unrequested side-effect. If an exception is thrown, a developer can always call close without a tabId to get this kind of behaviour. Yet it seems this should be optional, not enforced.

What I thought was - 
1. If a developer wants to close the side panel for a specific tab, they should provide the tabId. This action should only affect the tab-specific side panel on that particular tab.
2. If a developer wants to remove all side panels from a specific tab—regardless of whether they are global or tab-specific—they should provide both windowId and tabId.
@oliverdunk  WDYT?

oliverdunk:
> 2\. If a developer wants to remove all side panels from a specific tab—regardless of whether they are global or tab-specific—they should provide both windowId and tabId.

If `tabId` is specified, I don't think we should change behavior (other than basic validation that they match) depending on whether `windowId` is present. That would conflict with other APIs like `sidePanel.open()` where `windowId` is currently ignored if `tabId` is set.

If there is a global panel open, and you call close() with a `tabId`, to me it feels like the best option is to match the behavior of the user clicking the close icon in the UI. With that in mind, we would close the global panel everywhere. I feel less strongly on this though and think we could do whatever there is most alignment on.

@kiaraarose Do you have a preference from the Safari side?

kiaraarose:
> @kiaraarose Do you have a preference from the Safari side?

I agree with @carlosjeurissen that the behavior of closing a global sidePanel for all tabs if both `windowId` and `tabId` is a bit odd. If the intent is to close a global side panel, then not specifying the tabId would make sense. I can follow up with a dev who hacked on this for Safari (as a project, not shipped) to see they align.

Update: For Safari, we agree that having the sidePanel close for all tabs in this scenario wouldn't be ideal. I think an error for this case is sufficient.

oliverdunk:
> Update: For Safari, we agree that having the sidePanel close for all tabs in this scenario wouldn't be ideal. I think an error for this case is sufficient.

Thanks for following up. To confirm:

- If only `tabId` is specified, we can still close tab specific side panels. We don't need to require `windowId`.
- If `tabId` is specified, but the open side panel for that tab is a global side panel, throw an error.

Is that correct?

Also, an additional question: If you try to open a side panel for a `tabId`, but you only have a global side panel, should we also throw an error? I expect so but would want to check with Chrome folk as this would be a breaking change we would need to make.

oliverdunk:
Hi @kiaraarose / @carlosjeurissen - we have been [looking at](https://chromium-review.googlesource.com/c/chromium/src/+/7004729) updating `sidePanel.open` to remove the fallback to a global panel if `tabId` is specified but there is no tab-specific panel. However, in testing this breaks many of the popular extensions I tried including Grammarly, Phantom, Sider and QuillBot. With that in mind, it seems like this would be both a non-trivial breaking change and go against the expectations developers seem to naturally have.

Let's plan on discussing next steps at TPAC? Based on this I am once again leaning towards supporting a fallback to the global panel for both the `open()` and `close()` methods.

Rob--W:
I'm taking over the review from @zombie on behalf of Mozilla, since I have been actively involved in discussions on the design of sidepanel/sidebar APIs at TPAC 2025. Those meeting notes will be published soon.

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,142 @@
...
+- If both `windowId` and `tabId` are provided:
```
carlosjeurissen, line 95:
```suggestion
- If `tabId` is provided:
```

Calling only with a tabId should be possible

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Resolves when the panel has been closed. If the panel is
+  already closed, still resolves successfully.
+
+### Behavior
+
+- The operation will only close side panels that belong to the
+  calling extension.
+- If the panel is already closed or does not exist in the given
+  context, the method does nothing.
```
carlosjeurissen, line 90:
```suggestion
```

Move this to the two different scenarios, in which the method is called with a `tabId` or without a `tabId`

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,142 @@
...
+- If neither `windowId` nor `tabId` is provided, rejects with an error.
+- If `windowId` or `tabId` is invalid, rejects with an error.
+- If both `windowId` and `tabId` are provided, the method will verify
+  that the tab belongs to the specified window. If not, it rejects with an error.
+- If both `windowId` and `tabId` are provided:
+    1. If a contextual side panel is visible on the specified `tabId`, it will be closed.
+    2. If a global side panel is visible on the given `tabId`,
+       it will be closed for all tabs in that window, not just the specified tab.
```
carlosjeurissen, line 98:
```suggestion
    1. If no contextual side panel is configured for the specified `tabId`, reject with an error.
    2. If the contextual side panel is currently not open, the method does nothing.
    3. If the contextual side panel is currently open, it will close the contextual side panel.
```

proposals/sidepanel-close-api.md
```diff
@@ -0,0 +1,142 @@
...
+
```
carlosjeurissen, line 99:
```suggestion
- If `tabId` is not provided:
    1. If no global side panel is currently open in the window, this method does nothing.
    2. If a global side panel is currently open in the window, close just the global side panel.
```

carlosjeurissen (changes requested):
The suggested changes apply what has been discussed during TPAC2025. There is one open end:
Should we specify a global side panel will be reopened when closing a contextual one when switching tabs and returning to the tab with contextual side panel? As this is current Chrome behaviour.

Issue #833, opened 2025-05-22
Side Panel API: query position: left/right, close(), and focus(): tab to and from side panel


solomonkinard:
High level ideas for discussion:
1. query position: left/right
2. close()
3. focus(): tab to and fro side panel

dotproto:
From today's meeting, these items are under consideration by Chrome as work items for [Summer of Code](https://summerofcode.withgoogle.com/).

Given that someone may be working on `focus()` sooner than later, I'm going to prioritize the next revision of [PR 817](https://github.com/w3c/webextensions/pull/817).

carlosjeurissen:
**Sidebar close() method**
Issue: https://github.com/w3c/webextensions/issues/521
PR: https://github.com/w3c/webextensions/pull/837

**Sidebar focus() method**
Issue: https://github.com/w3c/webextensions/issues/693
PR: https://github.com/w3c/webextensions/pull/817

**Sidebar query position**
Issue: missing, could be this issue
PR: https://github.com/w3c/webextensions/pull/838

hharshas:
Chrome 140 now supports sidePanel.getLayout [#CL](https://chromium-review.googlesource.com/c/chromium/src/+/6576001).
The documentation of the new method is at https://developer.chrome.com/docs/extensions/reference/api/sidePanel#method-getLayout.

carlosjeurissen:
The `sidePanel.getLayout` may not return the actual side in some Chromium-based browsers like Vivaldi. See:
https://forum.vivaldi.net/topic/114346/sidepanel.getlayout-incorrectly-always-returns-side-right

PR #838, opened 2025-05-28
Proposal: API to Query Side Panel Layout

hharshas:
This proposal adds a method to detect the side panel's current alignment (left/right) as set in chrome://settings/appearance.  Feedback is appreciated! #833 [#chromium](https://issues.chromium.org/issues/406511291)  [Summer of Code](https://summerofcode.withgoogle.com/)

### DIFF
--- a/proposals/getposition-sidepanel.md
+++ b/proposals/getposition-sidepanel.md
@@ -1,4 +1,4 @@
-# Add API to Query Side Panel Position
+# Add API to Query Side Panel Layout
 
 **Summary**
 
@@ -47,14 +47,14 @@ namespace sidePanel {
     enum Side {
         LEFT = "left",
         RIGHT = "right"
-    }
+    };
 
-    interface PanelPosition {
+    interface PanelLayout {
         side: Side;
-    }
+    };
 
     // Example function declarations
-    function getPosition(): Promise<PanelPosition>;
+    function getLayout(): Promise<PanelPosition>;
 }
 ```
 
@@ -112,8 +112,8 @@ N/A
 ## Basic uses
 
 ```typescript
-chrome.sidePanel.getPosition().then(position => {
-  console.log(`Panel is on the ${position} side`);
-  applyPositionSpecificStyles(position);
+chrome.sidePanel.getLayout().then(layout => {
+  console.log(`Panel is on the ${layout.side} side`);
+  applyPositionSpecificStyles(layout.side);
 });
 ```
--- a/proposals/getposition-sidepanel.md
+++ b/proposals/getposition-sidepanel.md
@@ -1,0 +1,120 @@
+# Add API to Query Side Panel Position
+
+**Summary**
+Proposal to introduce a programmatic way to query the Side panel position.
+
+**Document Metadata**
+
+**Author:** [hharshas](https://github.com/hharshas)
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** [hharshas](https://github.com/hharshas) ([Summer Of Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
+
+**Created:** 2025-05-27
+
+**Related Issues:** [#chromium](https://issues.chromium.org/issues/406511291)
+
+## Motivation
+
+### Objective
+
+- At chrome://settings, it is possible to change the "Side panel position" which affects which side of the browser the side panel UI opens on. This can include panels registered by the developer using the sidePanel API.
+- We should add a way to query this so the developer can adjust their UI based on the position.
+
+#### Use Cases
+
+Programmatic Control: 
+- For example having phrasing or visuals that depend on the relative position of the page.
+
+### Known Consumers
+
+The Chromium bug has some amount of developer interest. This item is under consideration by Chrome among work items for Summer of Code.
+
+## Specification
+
+### Schema
+
+```typescript
+namespace sidePanel {
+ namespace sidePanel {
+  // Possible side panel positions
+  enum PanelPosition {
+    LEFT = "left",
+    RIGHT = "right"
+  };
+
+  interface Functions {
+    // Returns the current position of the side panel
+    // Returns: Promise that resolves with the current position
+    static Promise<PanelPosition> getPosition();
+    
+    ////
+    // Alternative approach for future extensibility (open for discussion)
+    dictionary PanelSettings {
+      PanelPosition position;
+      // Future settings could be added here
+    };
+    
+    static Promise<PanelSettings> getSettings();
+  };
+};
+```
+
+### Return value
+
+- Type: Promise<void>
+
+- Resolves when it returns the panel position.
+
+### Behavior
+    - Returns the current position as set in browser settings. 
+    - No context option is needed as the setting is same for all the context.
+
+### New Permissions
+N/A.
+
+### Manifest File Changes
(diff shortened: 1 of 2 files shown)


proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Created:** 2025-05-27
+
+**Related Issues:** [#chromium](https://issues.chromium.org/issues/406511291)
+
+## Motivation
+
+### Objective
+
+* At `chrome://settings`, it is possible to change the "Side panel position"
```
oliverdunk: We should be careful about when we reference Chrome specifically, since we want to motivate why this makes sense across all browsers. I would just mention this as one approach, for example:

> Browsers may offer users the ability to change which side of the browser the side panel opens on. In Chrome, this is exposed at chrome://settings.

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+The Chromium bug has some amount of developer interest. This item is under
+consideration by Chrome among work items for Summer of Code.
+
+## Specification
+
+### Schema
+
+```typescript
+namespace sidePanel {
+ namespace sidePanel {
```
oliverdunk:
```suggestion
```

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+    LEFT = "left",
+    RIGHT = "right"
+  };
+
+  interface Functions {
+    // Returns the current position of the side panel.
+    static Promise<PanelPosition> getPosition();
+    
+    // Alternative approach for future extensibility (open for discussion).
+    dictionary PanelSettings {
```
oliverdunk: It looks like there is a mix of TypeScript and IDL-like syntax here. Can you adjust this to be valid TypeScript, for example removing the Functions interface and using a type or interface for PanelSettings? Feel free to reach out to me or Solomon if you need assistance.

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+ namespace sidePanel {
+  // Possible side panel positions
+  enum PanelPosition {
+    LEFT = "left",
+    RIGHT = "right"
+  };
+
+  interface Functions {
+    // Returns the current position of the side panel.
+    static Promise<PanelPosition> getPosition();
```
oliverdunk: In the future, Chrome or other browser vendors might have different information to expose about the position. Regardless of if we use `PanelSettings`, let's also make `PanelPosition` a dictionary and then we can call the enum something like `Side` which can be a single key in that.

So for example:

```ts
enum Side {
    LEFT = "left",
    RIGHT = "right"
}

interface PanelPosition {
    side: Side;
}
```

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Additional Security Considerations
+
+N/A
+
+## Alternatives
+
+### Existing Workarounds
+
+So far, nothing works. The DOM of the side panel and the
```
oliverdunk: I don't think one is going to be found so let's be more definitive.

```suggestion
There are currently no workarounds. The DOM of the side panel and the
```

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,130 @@
...
+content script are separate, so even if we detect the resize
+event when the side panel is opened, we can't determine the
+panel's position. It will be interesting to see other's
+approaches in the comments of this PR.
```
oliverdunk: Let's remove "It will be interesting to see other's approaches in the comments of this PR." since this document will live far beyond the relevance of this PR :)

oliverdunk (changes requested):
Thanks! I've left a few initial comments here.

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,119 @@
...
+        LEFT = "left",
+        RIGHT = "right"
+    }
+
+    interface PanelPosition {
+        side: Side;
+    }
+
+    // Example function declarations
+    function getPosition(): Promise<PanelPosition>;
```
oliverdunk:
```suggestion
    function getLayout(): Promise<PanelLayout>;
```

As discussed offline, let's change this to `getLayout()`. Also open to other suggestions, but I think getPosition() is a little too specific.

oliverdunk (approved):
LGTM. Adding others to review.

hharshas:
We thought about using `getLayout()` as a more general name.

carlosjeurissen:
A few thoughts on this PR:

1) First of all, how do we as WECG want to deal with proposals which include aspects not all browsers are aligned on yet. In this situation, there is no unified sidebar API (See https://github.com/w3c/webextensions/issues/128). So creating a proposal which only considers the current implementation in one browser could be problematic.

2) Do we want to consider RTL layouts and read direction positions? As in, provide extension developers both with a physical location (left/right), as well as a reading position (end/start).

3) Do we want to allow different sidebars (for example, different tab specific sidebars), to have different positions on an API level? This would require a different API design.

4) Do we want to be flexible for the future and allow other kind of positions (like a panel at the bottom/top on specific foldable devices)? Or restrict ourselves to left/right?

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,119 @@
...
+**Author:** [hharshas](https://github.com/hharshas)
+
+**Sponsoring Browser:** Chromium
+
+**Contributors:** [hharshas](https://github.com/hharshas)
+([Summer Of Code](https://summerofcode.withgoogle.com/)), mentor: oliverdunk, solomonkinard
+
+**Created:** 2025-05-27
+
+**Related Issues:** [#chromium](https://issues.chromium.org/issues/406511291)
```
Rob--W:
```suggestion
**Related Issues:** https://issues.chromium.org/issues/406511291
```

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Motivation
+
+### Objective
+
+* Browsers may offer users the ability to change which side of the browser
+  the side panel opens on. In Chrome, this is exposed at `chrome://settings`.
+  This can include side panels registered by the extensions using the sidePanel API.
```
xeenon:
```suggestion
  This can include side panels registered by the extensions using the `sidePanel` API.
```

xeenon (approved):

proposals/getposition-sidepanel.md
```diff
@@ -0,0 +1,119 @@
...
+    };
+
+    // Example function declarations
+    function getLayout(): Promise<PanelPosition>;
+}
+```
+
+### Return value
+
+* Type: Promise<void>
```
Rob--W:
```suggestion
* Type: `Promise<PanelPosition>`
```

Rob--W (approved):
Other than the few nitpicks, I don't have objections to the proposal here, especially because we don't have an implementation and the change here is straightforward.

The loose naming convention that we have is to start the document name with the namespace that the proposal applies to, so that when sorted alphabetically, related proposals are close to each other. Could you rename the documents from your PRs, e.g. getposition-sidepanel.md` to `sidepanel-getposition.md`?

oliverdunk (approved):
Thanks Harsh!

PR #842, opened 2025-06-06
Document split tabs extension proposal

agale123:


carlosjeurissen:
Related Microsoft Edge issue:
https://github.com/microsoft/MicrosoftEdge-Extensions/issues/296

hanguokai:
What is the concept of "split" here? Is it a concept in browsers or operating systems? Do all browsers or operating systems support this concept? What conditions can cause an operation to fail? Is the behavior the same across platforms? 

Will this new feature introduce a breaking change? For example, will something that used to succeed now fail?

agale123:
Thanks for the link to the related Microsoft Edge issue. I've updated the proposal to reference that. 

As for the concept of a split, this is referring to a view within a single browser window where you can see two different web contents at once. On different browsers it is called things like Split Screen or Split View. Not all browsers support it currently but I know several already do (like Microsoft Edge and Arc) and Chrome is implementing it now. 

I can't speak to the other browser's implementations but within Chrome we are planning to fail gracefully if any conflicts occur. If extensions manipulate tabs in a way that is inconsistent with a split view then the split will be removed but the tabs will remain. For example, if tabs 0 and 1 are in a split view and a tab is moved between them then we will remove the split so the tab move can be executed.

hanguokai:
Thanks for the explanation. "Multiple tabs/pages visible at the same time in one browser window or one browser tab" is a big change. I tried Edge and Arc. Edge supports only two pages in one tab, while Arc supports multiple visible tabs in one window. I haven't understood the full behavior yet. Depends on the specific functional design, there may be impacts on various APIs details.

Rob--W:
On naming, after speaking with @sarah-clements (a tech lead on the Firefox desktop frontend team), she expressed the preference for the name to contain "view", for clarity what a split means.

WDYT of the naming `splitId` -> `splitViewId`, for example? Open to other ideas.

agale123:
@Rob--W I actually really like that suggestion since it helps with the confusion between split being a verb versus a noun. Updated the proposal.

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+        ...
+      }
+    ]
+  }
+]
+```
+
+### Behavior
+
+If a tab isn’t currently in a Split View, the splitViewId` will be -1, otherwise it
```
dotproto:
```suggestion
If a tab isn’t currently in a Split View, the `splitViewId` will be -1, otherwise it
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+elements of the tab strip and then verify if their actions worked as expected
+or handle any errors thrown when attempting to do things like move or close
+tabs.
+
+### Open Web API
+
+This isn’t adding a new API method, just updating an existing one to be more
+complete.
+
+
```
dotproto: Normalize spacing around headers.

```suggestion
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+* An extension wants to move a tab within the tab strip. With this API, it can
+detect if the moved tab is part of a Split View and can move the tabs within a
+Split View together or choose not to move the tab.
+* An extension wants to close a tab within the tab strip. With this API, it can
+detect that the closed tab was part of a Split View and can decide to close
+everything in the Split View or not to close the tab to preserve the Split
+View.
+
+
```
dotproto: Normalize spacing around headers.

```suggestion
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+unsplitting the view. So if a tab is moved in between two tabs in a Split View,
+that Split View will be removed and the two tabs will be separated. If one tab
+in a Split View is closed, the Split View will be removed and the one tab will
+be closed.
+
+## Future Work
+
+In the future, it would be nice to have an API to create, update, and remove
+Split Views. It would also be helpful to have an API to query/update details of
+a Split View like whether it is a vertical or horizontal.
```
dotproto: Newline at EOF

```suggestion
a Split View like whether it is a vertical or horizontal.

```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+  {
+    "namespace": "tabs",
+    "description": "Use the <code>chrome.tabs</code> API to interact with the browser's tab system. You can use this API to create, modify, and rearrange tabs in the browser.",
+    "types": [
+      {
+        "id": "Tab",
+        "type": "object",
+        "properties": {
+          ...
+          "splitViewId`": {
```
dotproto: Typo

```suggestion
          "splitViewId": {
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+      {
+        "name": "query",
+        "type": "function",
+        "description": "Gets all tabs that have the specified properties, or all tabs if no properties are specified.",
+        "parameters": [{
+   "type": "object",
+   "name": "queryInfo",
+   "properties": {
+     ...
+     "splitViewId`": {
```
dotproto: Typo

```suggestion
     "splitViewId": {
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+        "description": "Fired when a tab is updated.",
+        "parameters": [
+          {"type": "integer", "name": "tabId", "minimum": 0},
+          {
+            "type": "object",
+            "name": "changeInfo",
+            "description": "Lists the changes to the state of the tab that was updated.",
+            "properties": {
+              ...
+              "splitViewId`": {
```
dotproto: Typo

```suggestion
              "splitViewId": {
```

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,192 @@
...
+### Known Consumers
+
+We don’t have specific known consumers but we expect any consumers of the
+“groupId” property would also be interested in this new property.
+
+## Specification
+
+### Schema
+
+We will add a new property called “splitViewId`” to the Tab object. This will behave
```
dotproto: Typo

```suggestion
We will add a new property called “`splitViewId`” to the Tab object. This will behave
```

dotproto (changes requested):
This review only contains typo fixes and related polish.

proposals/split_tabs_proposal.md
agale123: Done

proposals/split_tabs_proposal.md
agale123: Done

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,191 @@
...
+### Schema
+
+We will add a new property called “`splitViewId`” to the Tab object. This will behave
+similarly to the “groupId” field that already exists.
+
+```
+[
+  {
+    "namespace": "tabs",
+    "description": "Use the <code>chrome.tabs</code> API to interact with the browser's tab system. You can use this API to create, modify, and rearrange tabs in the browser.",
```
Rob--W:
```suggestion
    ...
```

This is existing boilerplate, let's remove it so that it's clear that it is not a change in this proposal.

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,191 @@
...
+        "parameters": [{
+   "type": "object",
+   "name": "queryInfo",
+   "properties": {
+     ...
+     "splitViewId": {
+       "type": "integer",
+       "minimum": -1,
+       "optional": true,
+       "description": "The ID of the Split View that the tabs are in, or $(ref:tabs.SPLIT_TAB_ID_NONE) for tabs that aren't in a Split View."
```
Rob--W:
```suggestion
       "description": "The ID of the Split View that the tabs are in, or $(ref:tabs.SPLIT_VIEW_ID_NONE) for tabs that aren't in a Split View."
```

`SPLIT_TAB_ID_NONE` -> `SPLIT_VIEW_ID_NONE`? This constant is not defined in this document, let's do it for clarity.

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,191 @@
...
+      }
+    ]
+    ...
+    "functions": [
+      {
+        "name": "query",
+        "type": "function",
+        "description": "Gets all tabs that have the specified properties, or all tabs if no properties are specified.",
+        "parameters": [{
+   "type": "object",
```
Rob--W: nit: the indentation of the JSON here is off, which makes it harder to read where we are at. Could you fix up the formatting?

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,191 @@
...
+    ]
+  }
+]
+```
+
+### Behavior
+
+If a tab isn’t currently in a Split View, the `splitViewId` will be -1, otherwise it
+will have a unique value that represents the Split View. All tabs in the Split
+View will have the same identifier.
```
Rob--W: Are these IDs persistent across browser sessions?

agale123:
Any other feedback before landing this? Thanks!

Rob--W (commented):
I noticed that I have unsubmitted review comments that I wrote a long time ago; submitting them now.

proposals/split_tabs_proposal.md
agale123: In our implementation, we are planning to make these persistent across browser sessions but I'm not sure if that will be guaranteed elsewhere.

proposals/split_tabs_proposal.md
agale123: Done

proposals/split_tabs_proposal.md
agale123: Done

proposals/split_tabs_proposal.md
Rob--W: For reference:
- windowId, tabId are not persisted across browser sessions.
- groupId is not persisted across browser sessions in Chrome, in Firefox it is.
- splitViewId is apparently persisted across browser sessions in Chrome, as proposed.

In Firefox, the ID for a split view concept would also persist across sessions.

I'd be supportive of documenting these as persistent across browser sessions, because that would be valuable for extension developers who need to associate state to it.

Additional questions:
- If a window is closed and re-opened, what are the expected values of splitViewId?
- Is splitViewId integrated in the `sessions` API (the extension API that exposes re-opening closed tabs/windows).  This API is already supported in Firefox and Chrome, documentation at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/sessions/Session and https://developer.chrome.com/docs/extensions/reference/api/sessions#type-Session

agale123: From the Chrome perspective, we want to avoid locking ourselves into a session restore implementation that persists split IDs across sessions until after the initial launch when we are more confident that nothing key to the design will change. At this time, we suspect that it is fine to persist split IDs even though we don't persist group IDs but we want to validate those assumptions over time. For that reason, we just want to guarentee that the splitViewId will be the same between all tabs in a split view and won't be the same between two tabs not in a split view. 

That is a good question about the Sessions API. We aren't planning to implement Tab Restore right away but that is something we would like to do in the future. I don't want to land the spec for it until we are closer to landing the functionality so I added that to the Future Work section.

agale123:
I've responded to questions received thus far and made updates as needed. Let me know if there are any further concerns before landing this.

Rob--W (approved):
The proposal here looks good to me, and we don't need to wait for other changes before merging.

It would be nice to follow up with a PR that expands the proposal with a description of special circumstances in existing `tabs` APIs.

For example, when a tab is moved to a different index, a split is most likely going to unsplit. But the `tabs.move` extension API has the ability to move multiple tabs at once, so in that case it would be preferable to maintain the split.

agale123:
I'm not sure why Github is showing a remaining requested change. I think I've resolved all feedback from dotproto. Let me know if I'm missing anything. Hoping to get this landed.

proposals/split_tabs_proposal.md
```diff
@@ -0,0 +1,199 @@
...
+
+**Author:** agale123@
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:** N/A
+
+**Created:** 2025-06-06
+
+**Related Issues:** https://github.com/microsoft/MicrosoftEdge-Extensions/issues/296
```
Rob--W: Is there a crbug to link here?

Here is a Firefox bug that you can link: https://bugzilla.mozilla.org/show_bug.cgi?id=1993037

Rob--W:
From my perspective this should be merged. @xeenon Could you take a look?

proposals/split_tabs_proposal.md
agale123: Thanks for the Firefox bug. Also added the Chromium bug

proposals/split_tabs_proposal.md
dotproto: This is a non-blocking suggestion.

It seems like we're aligned on the idea that it would be desirable to persist split IDs across sessions, but there are reasonable concerns about committing to that behavior too early. In situations like this, I think it would be reasonable to specify the desired behavior using [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119)'s SHOULD add a non-normative note explaining why MUST was not used and state that we intend to change this in the future.

Since we're using plain markdown for these proposals, I suggest we use a [note blockquote](https://github.com/orgs/community/discussions/16925) for the non-normative note.

agale123: Yeah, that sounds reasonable. Made the update.

xeenon (approved):
Looks good to me.

dotproto (approved):

dotproto:
All reviewers have approved, so I am merging this pull request.

Dustin4444 (approved):

PR #852, opened 2025-07-17
Proposal: browser.test API

zombie:
As suggested by [WPT folks](https://github.com/web-platform-tests/rfcs/pull/219), we're moving the already approved `browser.test` API specification to our own repo.

### DIFF
--- a/proposals/browser_test_api.md
+++ b/proposals/browser_test_api.md
@@ -30,22 +30,24 @@ write and run a cross-browser test suite for their own extensions.
 
 ### Known Consumers
 
-All browsers already have this API namespace, and the goal of specifying it
-here is to align existing implementations around common behaviors.
+All browsers already use this API namespace internally, with several shared
+methods and some inconsistencies between implementations.  The intention of
+this proposal is to specify a desired common behavior and then introduce
+changes to align existing implementations around it.
 
 ## Specification
 
 ### Schema
 
 **`browser.test.assertTrue(condition, message)`**
-Asserts that a given condition is `true`. If the condition is `false`, the test fails.
+Asserts that a given condition is `true`, otherwise the test fails.
 
 **Parameters**
 - `condition` (boolean) – The condition to assert.
 - `message` (string, optional) – A message describing the assertion.
 
 **`browser.test.assertFalse(condition, message)`**
-Asserts that a given condition is `false`. If the condition is `true`, the test fails.
+Asserts that a given condition is `false`, otherwise the test fails.
 
 **Parameters**
 - `condition` (boolean)
--- a/proposals/browser_test_api.md
+++ b/proposals/browser_test_api.md
@@ -122,7 +122,7 @@ This API does not provide any additional avenue for abuse because
 ### Existing Workarounds
 
 Current ways of testing extensions use proprietary methods and non-standard
-protocols like CDP.  As a result, testing extensions is hard and as a result,
+protocols like CDP.  As a result, testing extensions is hard and especially,
 automated testing of extensions across browsers is very rare.
 
 ### Open Web API
@@ -141,7 +141,6 @@ Existing implementation have additional methods with different names and
 behaviors, and we plan to align and specify more of them, for example:
 
  - `assertRejects`
- - `assertThrows`
  - `log`
  - `sendMessage`
  - `withHandlingUserInput`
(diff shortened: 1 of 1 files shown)


kiaraarose (approved):
Thanks for moving this over!

mukul-p (approved):

xeenon (approved):

proposals/browser_test_api.md
```diff
@@ -0,0 +1,173 @@
...
+
+All browsers already have this API namespace, and the goal of specifying it
+here is to align existing implementations around common behaviors.
+
+## Specification
+
+### Schema
+
+**`browser.test.assertTrue(condition, message)`**
+Asserts that a given condition is `true`. If the condition is `false`, the test fails.
```
Rob--W: true or truthy?

Similarly below, false or falsey?

In Firefox, we treat them as truthy/falsey - we coerce the value to a boolean before comparing to true/false ([source](https://searchfox.org/mozilla-central/rev/1838f847aa3bf909c3d34a94a8f0cd7e37fca086/toolkit/components/extensions/child/ext-test.js#192,195,288-294)).

In Chrome, the input must be a boolean (or even stranger - a string...!) - https://source.chromium.org/chromium/chromium/src/+/main:extensions/renderer/resources/test_custom_bindings.js;l=94-177;drc=abe36fd704986c2e4f54b4bd35c291d6beb0e28c

(this piece of feedback doesn't block landing this PR, but I felt that it was worth calling out the discrepancy between implementations.)

Rob--W (approved):

proposals/browser_test_api.md
kiaraarose: For Safari, our implementation aligns with Firefox.

proposals/browser_test_api.md
```diff
@@ -0,0 +1,173 @@
...
+#### Use Cases
+
+The primary use case is testing WebExtensions implementations using
+Web Platform Tests.  Secondary use case is enabling developers to
+write and run a cross-browser test suite for their own extensions.
+
+### Known Consumers
+
+All browsers already have this API namespace, and the goal of specifying it
+here is to align existing implementations around common behaviors.
```
rdcronin: this is doing more than just specifying existing implementations / common behaviors; lots of this doesn't work in Chromium today.  That's okay (we've acknowledged there's work to be done here and this won't work out of the box), but I think it's worth highlighting that the intention of this proposal *is* to introduce behavioral changes in some browsers.

proposals/browser_test_api.md
```diff
@@ -0,0 +1,173 @@
...
+
+Each test function is treated as an individual test case. Tests run in order. When one completes,
+the next begins. If a test fails (by throwing or rejecting), execution of that test stops and the
+next one begins.
+
+The promise:
+- **Resolves** if all tests pass.
+- **Rejects** if any fail.
+
+Tests pass when they return `undefined` or a resolved promise. They fail if they:
```
rdcronin:
```suggestion
Tests pass when they either return `undefined` or when a promise returned by the test resolves.  They fail if they:
```

proposals/browser_test_api.md
```diff
@@ -0,0 +1,173 @@
...
+- Throw an exception
+- Return a rejected promise
```
rdcronin:
```suggestion
- Return a promise that rejects
```

(or similar.  A returned promise may not be settled at the point it is returned)

rdcronin (approved):
Thanks!  LGTM, but same comment as before that this is a diversion from Chromium's current implementation.

proposals/browser_test_api.md
zombie: I'll leave it as explicitly `true`, since that's a common denominator, and we can start writing tests that manually coerce to boolean, which should work the same everywhere.  Separately we can discuss whether we want to align on `truthy` here or introduce a separate `test.assert` or `test.ok`.

proposals/browser_test_api.md
zombie: Good point, I added a more explicit statement that this will require implementation changes.

Rob--W:
There was a follow-up question from a Googler on the shape of the `assertThrows` API, but the comment thread is not easy to find in Github. I quoted the original message and added a comment at: https://github.com/w3c/webextensions/commit/e08950c581969e45f6953e378e25abdb4ec2e410#commitcomment-186116825

Issue #94, opened 2021-10-02
Support for wildcards in externally_connectable.matches


minfrin:
Right now, there are two men-in-the-middle between a webpage and a webextension native app:

- The content script; or optionally on Chrome
- An external website limited to the top two levels of a domain name.

It must be possible for a webpage and a background script of an extension to be communicated with without a content script being present, and without being limited to an explicit external set of websites.

## content script

In order for a webpage to contact a webextension's native app via a content script, the communication flow is as follows:

**webpage -[1]-> content-script -[2]-> background-script -[3]-> native-app.**

Starting from the back:

[3] The native app sends messages to the background script over stdin/stdout (or API call for Safari). There is no man in the middle possible here.

[2] The background script is sent messages by the content script using browser.runtime.connect(). The content script in some cases can contact any background script, and there is no strong verification whether the name has been faked, thus a potential man-in-the-middle and https://github.com/w3c/webextensions/issues/76. In other cases (Safari, to my knowledge) the content script can only contact it's own background script, meaning that man in the middle is prevented. The content script however has "root" access to the webpage, and so has the same security level as an old Netscape plugin.

[1] The webpage sends messages to either all contact scripts of all extensions with document.dispatchEvent(), or to all open tabs with window.postMessage(). Here all extensions are a potential man-in-the-middle, or all open tabs are men-in-the-middle, thus https://github.com/w3c/webextensions/issues/78

## external website

On Chrome, In order for a webpage to contact a webextension's native app via an external website, the communication flow is as follows:

**webpage -[1]->  webextension-owner-webpage -[2]-> background-script -[3]-> native-app**

Starting from the back:

[3] The native app is contacted by the background script over stdin/stdout (or API call for Safari). There is no man in the middle possible here.

[2] The background script is contacted by the webextension-owner-webpage using chrome.runtime.connect() and externally_connectable. The externally_connectable option only works if a man-in-the-middle website is specified as a subdomain of https://publicsuffix.org/list/. Extensions wishing to be general purpose and work on all websites are forced to introduce the men-in-the-middle as specified above under "content script".

[1] The webpage is forced to pass information across to a website controlled by the extension owner., the webextension-owner-webpage is the man in the middle.

## fixing the problem

The simplest fix for the problem is to standardise `externally_connectable` while lifting the restriction that 'Patterns cannot include wildcard domains nor subdomains of (effective) top level domains;'.

[...]

Issue #22, opened 2021-06-24
request: standardize `externally_connectable` and support optional/dynamic matches


twschiller:
**Context**

Chromium browsers support a `externally_connectable` manifest key that enables websites to message the extension directly and securely

However, the list of `externally_connectable` sites is static in the manifest. Therefore, the primary current use case is connecting to the website associated with the extension

As there are many non site-specific extensions, making the list of `externally_connectable` sites to be dynamic (similar to optional hosts) would allow deeper extensibility between sites and these extensions. In this world, browser extension authors could provide a companion JS library that website developers would use to make their site customizable/extensible via the browser extension

**Request**

- Include support for `externally_connectable` in the web extension standard
- Allow the list to be modified using `browser.permissions` or a similar API
- Provide a way for a website to signal that it supports that browser extension via `externally_connectable`, so that the extension can react even if it doesn't have the activeTab permission yet (E.g., by confirming that the user should see a permissions prompt to give access). Potential methods: an HTML meta tag, an JS API available to the external site, e.g., `browsers.permissions.acceptExtensions(ids)`

**References**

* https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/externally_connectable
* A "polyfill" using `document.defaultView.postMessage` which requires access to the page to inject a content script: http://github.com/pixiebrix/pixiebrix-extension/blob/9acb556f36e5bf99d35daeb3dc1d77c0c9565287/src/contentScript/externalProtocol.ts#L191-L191

carlosjeurissen:
@twschiller work on specification can be found here:
https://github.com/w3c/webextensions/pull/873

PR #873, opened 2025-09-18
Add specification for `externally_connectable`.

kiaraarose:
Here's a first draft of a spec for `externally_connectable`. I'm not sure if the way it's written is spec-y enough so open to all feedback for improvements

***
<a href="https://pr-preview.s3.amazonaws.com/kiaraarose/webextensions/pull/873.html" title="Last updated on Aug 5, 2026, 7:36 PM UTC (69844f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webextensions/873/1fbcd29...kiaraarose:69844f4.html" title="Last updated on Aug 5, 2026, 7:36 PM UTC (69844f4)">Diff</a>

### DIFF
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -8,6 +8,7 @@ URL: https://w3c.github.io/webextensions/specification/index.html
 Editor: Mukul Purohit, Microsoft Corporation https://www.microsoft.com, mpurohit@microsoft.com
 Editor: Tomislav Jovanovic, Mozilla https://www.mozilla.org/, tjovanovic@mozilla.com
 Editor: Oliver Dunk, Google https://www.google.com, oliverdunk@chromium.org
+Editor: Kiara Rose, Apple https://www.apple.com, kiara_rose@apple.com
 Abstract: [Placeholder] Abstract.
 Markup Shorthands: markdown yes
 </pre>
@@ -143,7 +144,19 @@ This key may be present.
 
 ### Key `externally_connectable`
 
-This key may be present.
+The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
+
+This key may be present and may include the following optional keys:
+
+#### Key `ids`
+
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions, include the pattern `*` in the list. If left empty or omitted, no extensions can connect.
+
+#### Key `matches`
+
+A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
+
+Any [=match patterns=] that include wildcard domains, or wildcard subdomains of a top-level domain, must be ignored.
 
 ### Key `devtools_page`
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -146,11 +146,13 @@ This key may be present.
 
 The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
 
+A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
+
 This key may be present and may include the following optional keys:
 
 #### Key `ids`
 
-A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions, include the pattern `*` in the list. If left empty or omitted, no extensions can connect.
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions to connect, include the wildcard pattern `*`. If left empty or omitted while the `externally_connectable` key is present, no extensions can connect.
 
 #### Key `matches`
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -152,7 +152,7 @@ The <a href="#key-externally_connectable">`externally_connectable`</a> key decla
 
 A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
 
-These events are not available to web pages, only in privileged extension contexts.
+These events are not available to web pages, only in a [=privileged extension context=].
 
 This key may be present and may include the following optional keys:
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -152,6 +152,8 @@ The <a href="#key-externally_connectable">`externally_connectable`</a> key decla
 
 A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
 
+These events are not available to web pages, only in privileged extension contexts.
+
 This key may be present and may include the following optional keys:
 
 #### Key `ids`
@@ -162,7 +164,7 @@ A [=list=] of [=extension IDs=] that specifies which extensions can communicate 
 
 A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
 
-Any [=match patterns=] that include wildcard domains, or wildcard subdomains of a top-level domain, must be ignored.
+Note: Only the document's own URL is used for matching. There is no fallback to the URL of the document that created it.
 
 ### Key `devtools_page`
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -8,6 +8,7 @@ URL: https://w3c.github.io/webextensions/specification/
 Editor: Mukul Purohit, Microsoft Corporation https://www.microsoft.com, mpurohit@microsoft.com
 Editor: Tomislav Jovanovic, Mozilla https://www.mozilla.org/, tjovanovic@mozilla.com
 Editor: Oliver Dunk, Google https://www.google.com, oliverdunk@chromium.org
+Editor: Kiara Rose, Apple https://www.apple.com, kiara_rose@apple.com
 Abstract: [Placeholder] Abstract.
 Repository: w3c/webextensions
 Markup Shorthands: markdown yes
@@ -147,7 +148,19 @@ This key may be present.
 
 ### Key `externally_connectable`
 
-This key may be present.
+The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
+
+This key may be present and may include the following optional keys:
+
+#### Key `ids`
+
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions, include the pattern `*` in the list. If left empty or omitted, no extensions can connect.
+
+#### Key `matches`
+
+A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
+
+Any [=match patterns=] that include wildcard domains, or wildcard subdomains of a top-level domain, must be ignored.
 
 ### Key `devtools_page`
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -150,11 +150,13 @@ This key may be present.
 
 The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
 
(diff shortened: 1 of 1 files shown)


specification/index.bs
```diff
@@ -143,7 +143,19 @@ This key may be present.
 
 ### Key `externally_connectable`
 
-This key may be present.
+The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
```
kiaraarose, line 151: i'm not sure if this is the correct way to reference these two APIs. also, will the documentation for these live in this doc or are we planning to have separate docs for each API namespace? if the latter, then i'm not sure what's the best way to reference them here

Rob--W: We should at least mention that these keys control what is allowed to trigger the `runtime.onMessageExternal` and `runtime.onConnectExternal` events. This is a worthy callout because ordinarily, the `runtime.connect` and `runtime.sendMessage` trigger `runtime.onMessage` and `runtime.onConnect` instead.

specification/index.bs
```diff
@@ -143,7 +143,19 @@ This key may be present.
...
+
+This key may be present and may include the following optional keys:
+
+#### Key `ids`
+
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions, include the pattern `*` in the list. If left empty or omitted, no extensions can connect.
```
Rob--W: Question:
Should we emphasize that there are two ways to allow ANY extension to connect:

- `externally_connectable` is omitted altogether
- `externally_connectable` is specified and `ids: ["*"]` is set

You already mentioned in the introduction that omitting `externally_connectable` altogether allows any extension to connect, but it might be too easy to overlook that and conclude that nobody can connect when they read the "If left empty or omitted, no extensions can connect." here.

specification/index.bs
```diff
@@ -143,7 +143,19 @@ This key may be present.
...
+
+#### Key `matches`
+
+A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
```
Rob--W, line 165: There is an edge case where document URLs cannot be described by a match pattern. For these cases, we specified an algorithm to determine the URL to use for matching content scripts, at [specification/index.bs: "Determine the URL for matching a document"](https://github.com/w3c/webextensions/blob/6d2086f01a4d149bcd1ec024eb40e8b08b3e2072/specification/index.bs#L301-L331).

Can we point to this?

@oliverdunk Could you confirm that this is what Chrome does (or intends to use)?

specification/index.bs
Rob--W: Since these messaging APIs are web-exposed, I wonder whether we should make an effort to write down the namespace and methods that are expected to be supported, part of https://github.com/w3c/webextensions/blob/main/specification/window.browser.bs (or a separate doc, linked from there)

specification/index.bs
```diff
@@ -143,7 +143,19 @@ This key may be present.
...
+
+Any [=match patterns=] that include wildcard domains, or wildcard subdomains of a top-level domain, must be ignored.
```
MicahZoltu: Wildcard domains is, arguably, the most important thing that `externally_connectable` *should* support, and people have been asking for it for ~8 years.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1319168 for a long history, but the TL;DR is:

An extension can extend the built-in browser API via `externall_connectable` by enabling a wildcard `externally_connectable`.  Websites can then probe to see if APIs of interest are exposed by any installed extensions and communicate with the extension without the extension needing to inject content scripts into the page which is currently the only way to achieve this behavior.

carlosjeurissen:
Existing issue requesting specification: https://github.com/w3c/webextensions/issues/22

specification/index.bs
kiaraarose: would modifying that sentence to "If the list is left empty or omitted, no extensions can connect"? that way it's more clear that this is referring to `ids` and not `externally_connectable`

kiaraarose: verified with Rob that this sentence is more clear

specification/index.bs
xeenon: I'm not sure pointing to that algorithm would be helpful, since it is more about `content_scripts` and `match_origin_as_fallback` and `match_about_blank` which don't exist for `externally_connectable `.

specification/index.bs
kiaraarose: i think that would contradict this statement in the `window.browser` spec
> The specifics of what APIs within `window.browser` are at present intentionally undefined, and are left up to User Agent implementation.

specification/index.bs
```diff
@@ -143,7 +144,21 @@ This key may be present.
 
 ### Key `externally_connectable`
 
-This key may be present.
+The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
+
+A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
```
Rob--W, line 153:
```suggestion
A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
These events are not available to web pages, only in privileged extension contexts.
```

specification/index.bs
rdcronin: This is inaccurate today (for Chrome) -- we allow wildcard domains / wildcard subdomains.

specification/index.bs
Rob--W: I checked Chrome's source code for relevant behavior, which is at https://source.chromium.org/chromium/chromium/src/+/main:extensions/browser/api/messaging/message_service.cc;l=538-540;drc=373d7fdb08be061501a27cf72cb0ff846fbd0899

So it looks like these APIs can only be used from documents whose URL matches the match pattern. In practice, http(s): and file:-URLs. Not URLs like about:blank, blob:-, etc.

So let's resolve this thread by editing the section to emphasize that the match pattern should match the document's URL.

specification/index.bs
dotproto: I'd like to discuss this further, but I feel like a PR comment isn't the right place for a bigger conversation about this capability.

I shared some thoughts in https://github.com/w3c/webextensions/issues/94#issuecomment-4994833020. Please hold off on resolving this thread until we align in #94.

specification/index.bs
oliverdunk: This sounds like a good approach to me.

specification/index.bs
kiaraarose: We can update our behavior in Safari to match Chrome. At the time of writing this PR, their spec had `'Patterns cannot include wildcard domains nor subdomains of (effective) top level domains;'.` which is what we based our implementation on. 

We can remove this restriction and match Chrome's current behavior.

specification/index.bs
```diff
@@ -143,7 +144,23 @@ This key may be present.
...
+
+These events are not available to web pages, only in privileged extension contexts.
```
Rob--W: With #905 merged, we now have a definition for a privileged extension context that we can refer to:

```suggestion
These events are not available to web pages, only in [=privileged extension contexts=].
```

Rob--W (approved):
Small suggestion now that we have a definition of "privileged extension context". The preview is broken, if you rebase or merge the main branch, the preview should work again.

kiaraarose:
@oliverdunk @rdcronin i finally got around to updating this proposal. can y'all take another look when you get the chance?

specification/index.bs
```diff
@@ -147,7 +148,23 @@ This key may be present.
...
+
+These events are not available to web pages, only in a [=privileged extension context=].
+
+This key may be present and may include the following optional keys:
+
+#### Key `ids`
+
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions to connect, include the wildcard pattern `*`. If left empty or omitted while the `externally_connectable` key is present, no extensions can connect.
+
+#### Key `matches`
```
oliverdunk: Looking at the diff link added by the bot to your PR description, adding this heading has broken some of the links in the content scripts section (`matches` for content scripts now links here instead of to the right place). I'm not sure if the root problem is caused by this PR or earlier mistakes (in which case we don't necessarily need to block) but might be worth a quick look?

Image

oliverdunk (approved):
LGTM with nit - let's also get Devlin's review here.

specification/index.bs
kiaraarose: the issue here was that the new defintion of "Key matches" for externally connectable appears first in the spec so it was using that reference for all references to "Key matches". i fixed this by making the headings more specific (i.e. `#key-content_scripts-matches` and #key-externally_connectable-matches`

i think this would break any existing spec links to content script matches section but since `matches` is a common key for a lot of APIs, I think being specific here will prevent this issue from happening in the future.

oliverdunk: Got it - thanks!

specification/index.bs
```diff
@@ -147,7 +148,23 @@ This key may be present.
 
 ### Key `externally_connectable`
 
-This key may be present.
+The <a href="#key-externally_connectable">`externally_connectable`</a> key declares which extensions and web pages can establish connections to the extension using [=runtime.connect()=] and [=runtime.sendMessage()=]. If omitted, all extensions may connect, but no web pages can connect.
+
+A connection from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension triggers the [=runtime.onMessageExternal=] event listener to fire.
```
rdcronin, line 153: "A connection" is vague, and messages can be sent via a port or sendMessage.  Suggestion below.

```suggestion
A call to [=runtime.connect()=] from an external web page or extension triggers the [=runtime.onConnectExternal=] event listener to fire. Similarly, a message sent by an external web page or extension with [=runtime.sendMessage()=] triggers the [=runtime.onMessageExternal=] event listener to fire.
```

specification/index.bs
```diff
@@ -147,7 +148,23 @@ This key may be present.
...
+
+These events are not available to web pages, only in a [=privileged extension context=].
```
rdcronin, line 155:
```suggestion
These events are only available in a [=privileged extension context=], not in web pages or less-privileged contexts like [=content scripts=].
```

specification/index.bs
```diff
@@ -147,7 +148,23 @@ This key may be present.
...
+
+This key may be present and may include the following optional keys:
+
+#### Key `ids` #### {#key-externally_connectable-ids}
+
+A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions to connect, include the wildcard pattern `*`. If left empty or omitted while the `externally_connectable` key is present, no extensions can connect.
```
rdcronin, line 161:
```suggestion
A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions to connect, include the wildcard pattern `"*"`. If left empty or omitted while the `externally_connectable` key is present, no extensions can connect.
```

rdcronin, line 161: IMO, this is a very subtle (and regrettable!) detail:

* If externally_connectable is *not specified*, _all_ extensions can connect by default.
* If externally_connectable _is_ specified, but `ids` is not, _no_ extensions can connect.

That's alluded to here, but I think it's worth calling out more explicitly.

```
NOTE: The default of whether an extension can connect changes based on whether the `externally_connectable` key is present. If the key is omitted, the default is to allow all extensions to connect. If the key is present, no extensions can connect unless specified in `"ids"`, even if the `"ids"` key is ommitted.
```

(I think we have different styling classes for "ugly implementation note"?)

rdcronin (approved):
Thanks, Kiara!  LG % nits.

specification/index.bs
w4rw2dssxw-netizen: Hi

specification/index.bs
MicahZoltu: Did this conversation get resolved elsewhere, or accidentally dropped?

It appears that the final specification still doesn't allow `externally_connectable: '*'` (unless I'm missing it), despite Chrome's current behavior doing it and allegedly Safari indicating they would match Chrome's behavior.

Note the wording about allowed extensions allowing wildcards but allowed pages does not:
```
A [=list=] of [=extension IDs=] that specifies which extensions can communicate with the extension. To allow all extensions to connect, include the wildcard pattern `"*"`.
```
```
A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
```

oliverdunk: It seems like we should remove this line. @kiaraarose, would you like to open a PR? Otherwise I'm happy to.

MicahZoltu: I believe this line was removed, but the new set of lines still implies the same behavior (that there are no wildcards for websites, but there are wildcards for extensions).

Proposed change to latest version would be to change:
```diff
-A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.
+A [=list=] of [=match patterns=] that specifies which web pages can communicate with the extension. If left empty or omitted, no web pages can connect.  To allow all web pages to connect, include the wildcard pattern `"*"`.
```

Issue #882, opened 2025-10-02
Proposal: `WebRequest.securityInfo` API for receiving TLS certificate from web requests.


vkrot-cell:
## Summary

An API proposal to extend the `webRequest` API to provide TLS/QUIC certificate information for a given request.

## Objective

This API enables extensions to inspect the server certificate of a web request after a secure connection has been established. This fulfills a long-standing developer need for TLS introspection, allowing for the creation of advanced security and debugging tools.
Currently, Firefox has a similar [getSecurityInfo extensions API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/SecurityInfo), so this proposal mostly targets other browsers.

## Related Issues:
* [Mozilla Bug 1322748](https://bugzilla.mozilla.org/show_bug.cgi?id=1322748)
* [Similar Chromium Proposal](https://github.com/EFForg/webrequest-tlsinfo-api/blob/master/proposal.md)

## Use Cases

The lack of TLS introspection prevents the development of a number of security-focused browser extensions. This API would enable extensions to:
* Replicate the functionality of security scoring tools.
* Provide more granular or prominent certificate chain introspection than the browser's built-in UI.
* Attempt to detect Man-in-the-Middle (MITM) attacks.
* Assist in debugging TLS deployment and configuration issues.

## Schema

This API adds a new optional field, `securityInfo`, to the `details` object of the `onHeadersReceived`.

```typescript

export enum OnHeadersReceivedOptions {
   // .. old fields,

   // If used, then securityInfo will be provided.
   "securityInfo",
   // The option is used to provide securityInfo with raw bytes of a server certificate.
   "securityInfoRawDer"
}

export interface Fingerprint {
    /**
     * SHA-256 hash of the certificate.
     */ 
    sha256: string;
}

export interface CertificateInfo {
    fingerprint: Fingerprint;
    rawDER: Uint8Array;
}

/**
 * Represents the state of the network connection.
 */
export enum ConnectionState {
    /**
     * The TLS handshake failed (for example, the certificate has expired).
     */
    Broken = "broken",

    /**
     * The connection is not a TLS connection.
     */
    Insecure = "insecure",

    /**
     * The connection is a secure TLS connection.
     */
    Secure = "secure"
}

export interface SecurityInfo {
    /**
     * The array contains a single CertificateInfo object, for the server certificate, or empty in case 
     * the connection state is "insecure" (http).
     */
    certificates: CertificateInfo[];
[...]

PR #899, opened 2025-11-05
Proposal: `SecurityInfo` in `WebRequest`

vkrot-cell:
The proposal following discussion at #882

### DIFF
--- a/proposals/security_info_web_request.md
+++ b/proposals/security_info_web_request.md
@@ -1,0 +1,193 @@
+# Proposal: `SecurityInfo` in `webRequest` 
+
+**Summary**
+
+An API proposal to extend the `webRequest` API to provide TLS/QUIC certificate information and connection state for a given request.
+
+**Document Metadata**
+
+**Author:** [Vlad Krot](https://github.com/vkrot-cell)
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:** <other contributors emails or GitHub handles>
+
+**Created:** 2025-11-05
+
+**Related Issues:**
+* [Mozilla Bug 1322748](https://bugzilla.mozilla.org/show_bug.cgi?id=1322748)
+* [Previous abandoned Chromium Proposal](https://github.com/EFForg/webrequest-tlsinfo-api/blob/master/proposal.md)
+
+## Motivation
+
+### Objective
+
+This API enables extensions to inspect the server certificate of a web request after a secure connection (TLS/QUIC) has been established. This fulfills a long-standing developer need for TLS introspection, allowing for the creation of advanced security and debugging tools.
+
+#### Use Cases
+
+The lack of TLS introspection prevents the development of a number of security-focused browser extensions. This API would enable extensions to:
+* Replicate the functionality of security scoring tools.
+* Provide more granular or prominent certificate chain introspection than the browser's built-in UI.
+* Attempt to detect Man-in-the-Middle (MITM) attacks.
+* Assist in debugging TLS deployment and configuration issues.
+* Enforce custom security policies by blocking requests with insecure or non-compliant certificates.
+
+### Known Consumers
+
+This API would facilitate the migration of existing Firefox extensions to Chrome and other browsers that adopt this specification. 
+A notable example is [IndicateTLS](https://addons.mozilla.org/en-US/firefox/addon/indicatetls/), which relies on similar functionality.
+
+## Specification
+
+### Schema
+
+This proposal introduces a new optional field, `securityInfo`, to the details object of the `webRequest.onHeadersReceived` event. Its presence is controlled by new `OnHeadersReceivedOptions`.
+
+```typescript
+
+export enum OnHeadersReceivedOptions {
+   // .. old fields,
+
+   // If used, then securityInfo will be provided.
+   "securityInfo",
+   // The option is used to provide securityInfo with raw bytes of a server certificate.
+   "securityInfoRawDer"
+}
+
+export interface Fingerprint {
+    /**
+     * SHA-256 hash of the certificate.
+     */ 
+    sha256: string;
+}
+
+export interface CertificateInfo {
+    fingerprint: Fingerprint;
+    rawDER: Uint8Array;
+}
+
+/**
+ * Represents the state of the network connection.
+ */
+export enum ConnectionState {
+    /**
+     * The TLS handshake failed (for example, the certificate has expired).
+     */
+    Broken = "broken",
+
+    /**
+     * The connection is not a TLS connection.
+     */
+    Insecure = "insecure",
+
+    /**
+     * The connection is a secure TLS connection.
+     */
+    Secure = "secure"
+}
+
+export interface SecurityInfo {
+    /**
+     * The array contains a single CertificateInfo object, for the server certificate, or empty in case 
+     * the connection state is "insecure" (http).
+     */
+    certificates: CertificateInfo[];
+    /**
+     * State of the connection.
+     */
+    state: ConnectionState;
+}
+
+export interface OnHeadersReceivedDetails {
+    // ... existing fields
+    /**
+     * Information about the security of the connection. This is only
+     * present if "securityInfo" is passed in the extraInfoSpec parameter
+     * and the request used a secure connection.
+     */
+    securityInfo?: SecurityInfo;
+}
+
+```
+
+### Behavior
+
+* New `onHeadersReceivedOptions` are necessary for performance reasons. They specify whether ssl data must be kept for as long as the web request is alive. Without them, unnecessary data can be kept for longer. Which is very undesirable especially in post quantum cryptography, since certificates can take a significant amount of memory space. 
+Note that Firefox has no such an option, because it only supports blocking web requests call while there’s still ssl data.
+
+* A new `securityInfo` object can be obtained in the `onHeadersReceived` event listener.
+
+* To receive this information, an extension **must** include `"securityInfo"` or `"securityInfoRawDer"` in the `extraInfoSpec` array when calling `addListener`. This opt-in design prevents performance overhead for the majority of extensions that don't need this data.
+
+* The `securityInfo` object will only be populated for requests made over a secure protocol (e.g., HTTPS, WSS) where the TLS/QUIC handshake has successfully completed or also in case of certificate errors.
+Browsers interrupt connections when there's a certificate error, unless user has explicitely allowed it in the browser UI, only in this case it is possible to have SecurityInfo with `state = "broken"`.
+
+* `certificates` - will contain only the leaf server certificate. This is done for future extensibility and Firefox API compatibility, because there they also provide a leaf if [certificateChain](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/getSecurityInfo#certificatechain) is not included in getSecurityInfo options.
+
+* `state` - State of the connection. One of:
+    * `"broken"`: the TLS handshake failed (for example, the certificate has expired)
+    * `"insecure"`: the connection is not a TLS connection
+    * `"secure"`: the connection is a secure TLS connection
+    * Note that Firefox extension API has a `“weak”` state of connection, which does not exist in Chrome.
+
+* The `CertificateInfo.rawDER` field contains the raw certificate bytes in DER format, which can be parsed by the extension using a third-party library. This field is only provided if `extraInfoSpec` array includes `"securityInfoRawDer"`. The reason for it is a performance optimization to not pass raw bytes when
+it is not necessary, and compatibility with Firefox extensions API.
+
+### Example
+
+Here is an example of an extension listening for web requests to log the server's certificate fingerprint.
+
+```javascript
+chrome.webRequest.onHeadersReceived.addListener(
+  (details) => {
+    if (details.securityInfo && details.securityInfo.state == "secure") {
+      console.log(`Received certificate for ${details.url}. The fingerprint is ${details.securityInfo.certificates[0].fingerprint.sha256}`);
+    }
+  },
+  // Filter for requests.
+  { urls: ["<all_urls>"] },
+  // Opt-in to receive the security info.
+  ["securityInfo"]
+);
+```
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+This API exposes the server's leaf certificate for a given connection. This is public information that is exchanged as part of any standard TLS handshake.
+So, no sensitive data is exposed.
+
+## Alternatives
+
+### Existing Workarounds
+
+Currently, extensions can only approximate this functionality by capturing a request's URL and sending it to a separate backend service. This service must then perform its own TLS handshake with the destination server to retrieve the certificate. This approach has significant drawbacks:
+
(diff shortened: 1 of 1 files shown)


vkrot-cell:
@Rob--W Hi, any comments on the proposal?

Rob--W:
> @Rob--W Hi, any comments on the proposal?

I'm still in Japan following TPAC last week. I also have not had the chance to query for API usage (which I mentioned in https://github.com/w3c/webextensions/blob/main/_minutes/2025-11-06-wecg.md ), but that does not have to block my review.

The most important part is to verify that similar names and properties have equivalent meanings across browsers. I am not available until mid December, but if this is really pressing I can ask another team member to assist in the verification (or do that myself when idle). On what timeframe do you need feedback?

vkrot-cell:
@Rob--W It is fine for me to wait a week or two.

vkrot-cell:
Hi @Rob--W
Can you take a look this week?

Rob--W:
> Hi @Rob--W
> Can you take a look this week?

I'll try to do so, yes. I wrote "mid December" before to have a bit of buffer, but I actually put this issue on my todo list for this week.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+     */
+    Secure = "secure"
+}
+
+export interface SecurityInfo {
+    /**
+     * The array contains a single CertificateInfo object, for the server certificate, or empty in case 
+     * the connection state is "insecure" (http).
+     */
+    certificates: CertificateInfo[];
```
Rob--W: In Firefox `certificates` is omitted when `state` is `insecure` (http).
But from the API design perspective I would also be okay with using an empty array.

Do you have strong preferences either way?

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * State of the connection.
+     */
+    state: ConnectionState;
```
Rob--W, line 103: As commented in https://github.com/w3c/webextensions/issues/882#issuecomment-3657931985, just `state` is not enough to tell the difference between trusted and untrusted. Can we add at least `isUntrusted`?

I see that you have specified the "broken" behavior for the scenario of "user has explicitely allowed it in the browser UI", below. This is not what Firefox has currently implemented. "broken" is a really broken state; if the certificate is good other than the expiry date, and the user accepted it, then the state is considered "secure" in Firefox.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+}
+
+export interface Fingerprint {
+    /**
+     * SHA-256 hash of the certificate.
+     */ 
+    sha256: string;
+}
+
+export interface CertificateInfo {
```
Rob--W, line 68: Firefox includes more properties in `certificateInfo`. In theory, all properties could be parsed from the certificate information with a JS library, except for [isBuiltInRoot](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/CertificateInfo#isbuiltinroot). This bit is not inherently a part of the certificate itself, but of the browser.

Question: Did you consciously choose to omit the other documented fields? Was there any particular reason to do so?

extraInfoSpec is specified to accept `securityInfo` and `securityInfoRawDer`, but without other properties the `securityInfo` option is of limited use. Most extensions don't pass the `rawDER` option

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+export enum OnHeadersReceivedOptions {
+   // .. old fields,
+
+   // If used, then securityInfo will be provided.
+   "securityInfo",
+   // The option is used to provide securityInfo with raw bytes of a server certificate.
+   "securityInfoRawDer"
+}
+
+export interface Fingerprint {
```
Rob--W, line 61: Note: Firefox also includes the `sha1` fingerprint in addition to `sha256`, but the use of sha1 is no longer recommended, and I'd be fine with omitting it.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+* Replicate the functionality of security scoring tools.
+* Provide more granular or prominent certificate chain introspection than the browser's built-in UI.
+* Attempt to detect Man-in-the-Middle (MITM) attacks.
+* Assist in debugging TLS deployment and configuration issues.
+* Enforce custom security policies by blocking requests with insecure or non-compliant certificates.
+
+### Known Consumers
+
+This API would facilitate the migration of existing Firefox extensions to Chrome and other browsers that adopt this specification. 
+A notable example is [IndicateTLS](https://addons.mozilla.org/en-US/firefox/addon/indicatetls/), which relies on similar functionality.
```
Rob--W: The IndicateTLS example does not use certificates as proposed in this proposal; it relies on other fields that are missing. For a write-up of known consumers, see https://github.com/w3c/webextensions/issues/882#issuecomment-3657931985

On the Chromium side, there have been two feature requests for "security info" functionality, at:
- https://issues.chromium.org/issues/41264310 Extend webRequests onCompleted callback to include information about TLS connections
- https://issues.chromium.org/issues/40754456 Feature request: webRequest.getSecurityInfo()

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+     */
+    Insecure = "insecure",
+
+    /**
+     * The connection is a secure TLS connection.
+     */
+    Secure = "secure"
+}
+
+export interface SecurityInfo {
```
Rob--W, line 94: This proposal only includes `certficates`, but that is not enough to fulfill the stated objective of "TLS introspection". Connection-specific information that are useful includes:
- Cipher suite (`cipherSuite`)
- key exchange algorithm (`keaGroupName`)
- key (well key size at least - `secretKeyLength`)
- protocol version (`protocolVersion`)

There are more properties, see https://github.com/w3c/webextensions/issues/882#issuecomment-3657958537 for an example of the actual API result,
and https://github.com/w3c/webextensions/issues/882#issuecomment-3657931985 for a write-up and actual API usage.

The properties are currently listed in "future work". Even if you're going to implement this in two steps, it would still be nice for the proposal to be complete. Are you considering to implement additional fields, and if so, which?

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+     */
+    securityInfo?: SecurityInfo;
+}
+
+```
+
+### Behavior
+
+* New `onHeadersReceivedOptions` are necessary for performance reasons. They specify whether ssl data must be kept for as long as the web request is alive. Without them, unnecessary data can be kept for longer. Which is very undesirable especially in post quantum cryptography, since certificates can take a significant amount of memory space. 
+Note that Firefox has no such an option, because it only supports blocking web requests call while there’s still ssl data.
```
Rob--W, line 121: This part can be dropped now. We agreed to include it in the event details, and Firefox could also implement that.

```suggestion
```

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+export interface Fingerprint {
+    /**
+     * SHA-256 hash of the certificate.
+     */ 
+    sha256: string;
+}
+
+export interface CertificateInfo {
+    fingerprint: Fingerprint;
+    rawDER: Uint8Array;
```
Rob--W:
```suggestion
    // Only included if "securityInfoRawDer" is passed to extraInfoSpec
    rawDER?: Uint8Array;
```

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+Note that Firefox has no such an option, because it only supports blocking web requests call while there’s still ssl data.
+
+* A new `securityInfo` object can be obtained in the `onHeadersReceived` event listener.
+
+* To receive this information, an extension **must** include `"securityInfo"` or `"securityInfoRawDer"` in the `extraInfoSpec` array when calling `addListener`. This opt-in design prevents performance overhead for the majority of extensions that don't need this data.
+
+* The `securityInfo` object will only be populated for requests made over a secure protocol (e.g., HTTPS, WSS) where the TLS/QUIC handshake has successfully completed or also in case of certificate errors.
+Browsers interrupt connections when there's a certificate error, unless user has explicitely allowed it in the browser UI, only in this case it is possible to have SecurityInfo with `state = "broken"`.
```
Rob--W, line 128: I have commented elsewhere; in Firefox `state` is still "secure"` if the user intentionally accepted an expired certificate.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+
+* `certificates` - will contain only the leaf server certificate. This is done for future extensibility and Firefox API compatibility, because there they also provide a leaf if [certificateChain](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/getSecurityInfo#certificatechain) is not included in getSecurityInfo options.
+
+* `state` - State of the connection. One of:
+    * `"broken"`: the TLS handshake failed (for example, the certificate has expired)
+    * `"insecure"`: the connection is not a TLS connection
+    * `"secure"`: the connection is a secure TLS connection
+    * Note that Firefox extension API has a `“weak”` state of connection, which does not exist in Chrome.
```
Rob--W: MDN documents `weak`, but it is never set in practice. It can be ignored.

When `webRequest.getSecurityInfo` was introduced (Firefox 62), that state was never set in practice. It was introduced internally to signal RC4 weaknesses, but that deprecation & removal was completed long before.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+
+* A new `securityInfo` object can be obtained in the `onHeadersReceived` event listener.
+
+* To receive this information, an extension **must** include `"securityInfo"` or `"securityInfoRawDer"` in the `extraInfoSpec` array when calling `addListener`. This opt-in design prevents performance overhead for the majority of extensions that don't need this data.
+
+* The `securityInfo` object will only be populated for requests made over a secure protocol (e.g., HTTPS, WSS) where the TLS/QUIC handshake has successfully completed or also in case of certificate errors.
+Browsers interrupt connections when there's a certificate error, unless user has explicitely allowed it in the browser UI, only in this case it is possible to have SecurityInfo with `state = "broken"`.
+
+* `certificates` - will contain only the leaf server certificate. This is done for future extensibility and Firefox API compatibility, because there they also provide a leaf if [certificateChain](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/getSecurityInfo#certificatechain) is not included in getSecurityInfo options.
```
Rob--W, line 130: FYI: the majority of extensions calling `getSecurityInfo` also set the `certificateChain` option to true. Not including all certificate information may be too limiting in practice.

proposals/security_info_web_request.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Implementation Notes
+
+This proposal is heavily inspired by Firefox's `webRequest.getSecurityInfo()` to promote cross-browser compatibility. However, it adapts the design to fit Chrome's non-blocking, event-driven architecture (Manifest V3):
+
+* **Event-Based vs. Function Call**: Instead of a separate `getSecurityInfo(requestId)` function (as in Firefox), this API injects the data into the existing `onHeadersReceived` event.
+* **Rationale**: In a non-blocking model, the internal network and certificate data associated with a request is often discarded immediately after the request pipeline advances. A separate asynchronous getSecurityInfo call would be racy and unreliable. By "opting in" with extraInfoSpec, the extension signals the browser to hold onto this data just long enough to deliver it with the onHeadersReceived event, ensuring data availability without blocking.
+
+
+## Future Work
```
Rob--W, line 190: Would it make sense to also support `securityInfo` in `webRequest.onErrorOccurred`? This would allow extensions to measure connection failures and debug it.

Rob--W (commented):
Thanks for the patience; this feedback took longer than expected because I wanted to complete the API usage analysis first (shared in the issue at https://github.com/w3c/webextensions/issues/882#issuecomment-3657931985).

proposals/security_info_web_request.md
OV-ETN: yes please

OV-ETN:
thanks for this

proposals/security_info_web_request.md
vkrot-cell: I do not have a strong preference about it, so let's do it as Firefox already done (meaning the field is optional). Thanks for noting it.

proposals/security_info_web_request.md
vkrot-cell: "Did you consciously choose to omit the other documented fields? Was there any particular reason to do so?" - So I only included fields necessary to accomplish "google internal" goal. I did not evaluate deeply how extension devs would use it. Thank you for this research.

proposals/security_info_web_request.md
vkrot-cell: Let's omit it, it is deprecated certificate algorithm.

proposals/security_info_web_request.md
vkrot-cell: Thanks for giving more insight on what is weak in Firefox. So, I would just remove this line with "weak" state.

proposals/security_info_web_request.md
vkrot-cell: I don't have final answer on this, but internally including certificate info in onErrorOccurred had a pushback. There was a concern that giving ability to devs to use "bad" certificates is not good. 
I will add this in the Future work section, as to me this addition makes sense.

proposals/security_info_web_request.md
vkrot-cell: I answered in https://github.com/w3c/webextensions/issues/882#issuecomment-3705600431

proposals/security_info_web_request.md
vkrot-cell: I agree with you on this: extraInfoSpec is specified to accept securityInfo and securityInfoRawDer, but without other properties the securityInfo option is of limited use. Most extensions don't pass the rawDER option

However, rawDER is absolute must to include for "google-internal" needs. And yes, securityInfo then is not super useful, only fingerprint and state. 

Is it a blocker to not have enough fields?

proposals/security_info_web_request.md
vkrot-cell: I totally agree, but still I would prefer to leave this for another iteration of the API, which was developed to be compatible with this option.

proposals/security_info_web_request.md
vkrot-cell: Thanks, I have added this feature requests to Related Issues, and corrected sentence a bit about IndicateTLS.

proposals/security_info_web_request.md
Rob--W: > Is it a blocker to not have enough fields?

In the context of this thread, it is okay for parsed certificate fields to be missing, because these can be derived from the certificate with JS libraries. Even the sha256 fingerprint can be computed. So if you don't have a purpose for that field in your use case, and want a minimal API, you could also leave it off altogether.

In the context of the broader proposal, I question the completeness of `securityInfo` when fields for non-certificate information are unspecified. I asked about that before at: https://github.com/w3c/webextensions/pull/899/files#r2621143436 and am still interested in an answer.

Issue #920, opened 2025-11-27
Should extension APIs consume user activation?


oliverdunk:
The User Activation [spec](https://html.spec.whatwg.org/multipage/interaction.html#user-activation-gated-apis) defines two relevant types of APIs gated by user activation:

> ...
>
> - **Transient activation-gated APIs**
> These APIs require the [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) state to be true, but they don't [consume](https://html.spec.whatwg.org/multipage/interaction.html#consume-user-activation) it, so multiple calls are allowed per user activation until the transient state [expires](https://html.spec.whatwg.org/multipage/interaction.html#activation-expiry).
> 
> - **Transient activation-consuming APIs**
> These APIs require the [transient activation](https://html.spec.whatwg.org/multipage/interaction.html#transient-activation) state to be true, and they [consume user activation](https://html.spec.whatwg.org/multipage/interaction.html#consume-user-activation) in each call to prevent multiple calls per user activation.
>
> ...

Should extension APIs be activation consuming, and if so, should we do these across all APIs or on a case by case basis?

oliverdunk:
We are supportive of this on the Chrome side and tentatively believe all extension APIs we have today should consume activation.

nasreenkhannam:
Hello everyone 👋
I’m Nasreen Khan, new to the WebExtensions community.

From my understanding, the core question is whether extension APIs should only work after an explicit user activation (like a click), or whether some APIs can run without consuming activation.

I’d like to contribute to this discussion.
Here are some thoughts:

1. Predictability for users:
Requiring user activation for sensitive APIs (e.g., opening windows, modifying tabs, clipboard) improves user trust and reduces the chance of unwanted behavior.


2. Developer flexibility:
Some extension actions (e.g., scheduled background tasks or passive listeners) may need to run without user activation. Making all APIs require activation may break valid workflows.


3. Possible middle ground:

Critical / intrusive APIs → must require user activation

Non-critical APIs → should not require activation

Provide clear documentation of which APIs consume activation




I’m happy to help draft a clearer list of APIs that should consume activation and those that shouldn't.
Please let me know if this direction is useful.

Thank you!

tophf:
> tentatively believe all extension APIs we have today should consume activation

[...]

PR #921, opened 2025-11-29
User activation proposal

nasreenkhannam:
This PR adds a proposal document describing rules for when WebExtension APIs
should consume user activation. The proposal covers:

- Intrusive APIs requiring activation
- Non-intrusive APIs not requiring activation
- Clear documentation needs
- A balanced approach for cross-browser consistency

This proposal is in response to Issue #920.
Feedback is welcome, and I’m happy to refine it further.

### DIFF
--- a/proposals/-user-activation-consumption.md
+++ b/proposals/-user-activation-consumption.md
@@ -1,0 +1,53 @@
+# Proposal: Rules for User Activation Consumption in WebExtensions
+
+This proposal is related to Issue #920: “Should extension APIs consume user activation?”
+
+## 1. Why this proposal?
+Some extension APIs should run only after a user gesture (click, keypress) for safety.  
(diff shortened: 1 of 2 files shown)


tophf:
>  (should consume activation)
> - Opening or closing tabs
> - Modifying tabs (move, update, reload)
> - Opening windows

Absolutely not, because this would destroy a lot of important use cases where extensions are configured to do things on schedule or which are designed to intercept certain URL/headers/content and open a new window/tab in response. Overall the PR looks AI-generated and not based on any analysis of the existing use cases and how the existing extensions are used or should be used in practice.

On the other hand, browser makers might want to work on improving the indication of which extension has opened or modified a tab/window by adding this info either to the menu of the tab/window (e.g. "Extensions modification history") or in the info badge inside the location bar (although popups created by extensions in Chrome don't have one).

PR #928, opened 2025-12-18
Publish minutes of 2025-12-18 meeting

Rob--W:
Generated from https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit using the tool and process from https://github.com/w3c/webextensions/pull/105.

During this meeting we discussed or mentioned issues #925, #531, #927, #926, #353, #617 and PRs #921, #922, #842.

PR #932, opened 2025-12-22
Update `runtime.onLoadedReason` to match `runtime.OnInstalledReason`

carlosjeurissen:
This changes the reasons to the present simple as agreed during the [2025-12-18](https://github.com/w3c/webextensions/pull/928) meeting to align with runtime.onInstalledReason

This PR also adds the `browser_update` and `shared_module_update` reasons.

Alternatively we could also add these reasons to runtime.onInstalledReason and use it for onExtensionLoaded as well.

@oliverdunk when this gets implemented in Chrome, is Google in favour of using `browser_update` instead of `chrome_update` as reason at least in the onExtensionLoaded event?

@oliverdunk once merged we should update https://issues.chromium.org/issues/41486131

Lastly, there is open discussion on what should happen if multiple reasons apply. See:
https://github.com/w3c/webextensions/issues/353#issuecomment-3683960304

@kiaraarose the PR also replaces all `chrome.runtime` instances with `browser.runtime`.

### DIFF
--- a/proposals/runtime_on_load_on_enabled_events.md
+++ b/proposals/runtime_on_load_on_enabled_events.md
@@ -23,12 +23,12 @@ take action on the full “lifecycle” of their extensions.
 
 Enable two new extension lifecycle events for the developer to add listeners for:
 
-#### chrome.runtime.onEnabled
+#### browser.runtime.onEnabled
 Enables a developer to listen to when an extension is re-enabled from a disabled
 state. There is currently no way for a developer to listen for the extension
 being disabled and then enabled.
 
-### chrome.runtime.onExtensionLoaded
+### browser.runtime.onExtensionLoaded
 Enables a developer to listen to when an extension is added to the set of active
 extensions. This can occur from being installed, enabled, or when the browser /
 profile first starts. The listener will be given the cause so they can decide to
@@ -36,7 +36,7 @@ take different actions based on the reason it was added to the set. There is
 currently no generic event that encapsulates all instances of extension install
 (and update), startup, and being enabled from a disabled state.
 
-#### Use Cases `chrome.runtime.onEnabled`
+#### Use Cases `browser.runtime.onEnabled`
 A developer wants to run code once in an extension’s "lifecycle", without storing
 a flag to keep track of this.
 
@@ -48,13 +48,13 @@ complexity.
 
 More use cases can be found [here](https://github.com/w3c/webextensions/issues/353#issuecomment-1490078217).
 
-#### Use Cases `chrome.runtime.onExtensionLoaded`
+#### Use Cases `browser.runtime.onExtensionLoaded`
 An extension automatically does initial bootstrapping, such as creating an
 initial state used throughout the browser session, creating a WebSocket
 connection, checking for things you've missed while the browser was closed etc.
 
 Before this API the developer would have to create multiple listeners for
-`chrome.runtime.onInstalled()`, `onStartup()`, and (above) `onEnabled()`.
+`browser.runtime.onInstalled()`, `onStartup()`, and (above) `onEnabled()`.
 
 With this new API method the developer could write one method that performs the
 same action for all three, potentially with small differences based on the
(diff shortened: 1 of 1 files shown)


proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -95,19 +95,27 @@ namespace runtime {
...
+  // 'enable': The extension was re-enabled from a disabled state.
   //
-  // 'updated': The extension was reloaded after an update.
+  // 'install': The extension was newly installed.
+  //
+  // 'update': The extension was reloaded after an update.
+  //
+  // 'browser_update': The browser was updated to a new version.
+  //
+  // 'shared_module_update': Another extension, which contains a module used by this extension, was updated.
```
oliverdunk: As far as I know, [shared modules](https://developer.chrome.com/docs/extensions/reference/manifest/shared-modules) are a Chrome-only concept (and aren't currently supported for extensions in the Chrome Web Store). @Rob--W / @xeenon: Is there any interest in mentioning those here or should we just make that an additional Chrome-specific reason?

oliverdunk:
> @oliverdunk when this gets implemented in Chrome, is Google in favour of using browser_update instead of chrome_update as reason at least in the onExtensionLoaded event?

I'll reach out to Devlin about this and follow-up. I assume we will be supportive though :)

oliverdunk:
> I'll reach out to Devlin about this and follow-up. I assume we will be supportive though :)

Confirming that we are ok using `browser_update` for new events. We don't plan to change any existing ones in the short term.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -95,19 +95,27 @@ namespace runtime {
...
   //
   // 'startup': The extension is being loaded during browser startup.
   //
   // 'reload': The extension was reloaded (e.g. via `chrome.runtime.reload() or`
```
kiaraarose: `browser.runtime.reload()`

carlosjeurissen: @kiaraarose thanks for noting! Replaced all chrome instances with browser.

proposals/runtime_on_load_on_enabled_events.md
kiaraarose: Not including it makes sense to me.

carlosjeurissen:
@oliverdunk Removed the shared module reason. Also specified only one event should be fired and there is a specific order of importance specified.

We may want to consider removing runtime.onEnabled from the proposal if we proceed with not implementing onEnabled.

proposals/runtime_on_load_on_enabled_events.md
```diff
@@ -135,6 +140,10 @@ namespace runtime {
 
 Described as code comments in schema description.
 
+If multiple reasons apply at the same time, only one event should be fired.
+The reason first in the below chain should be given:
```
Rob--W:
```suggestion
The precedence of reasons is as follows, with the first item having the highest precedence:
```

Feel free to copy-edit further, I wanted to include the word "precedence" for clarity.

And also, should we emphasize that if the event were to fire again due to incognito split (see earlier in the document), that the reason would be "startup"? Or should we introduce yet another reason specifically for that case?

Rob--W (approved):

proposals/runtime_on_load_on_enabled_events.md
carlosjeurissen: Removed the shared modules!

proposals/runtime_on_load_on_enabled_events.md
carlosjeurissen: @Rob--W made the copy change. Thanks for the suggestion.

As for your scenario, seems that is another scenario besides the ones which we may need to resolve from issue 353. We could leave these scenarios out of this PR and discuss them first in 353 so we can get this merged?

Rob--W: Sounds good. I will raise it in #353.

Rob--W: Commented at https://github.com/w3c/webextensions/issues/353#issuecomment-3757271543

oliverdunk (approved):
Thanks Carlos! Approving on behalf of Chrome.

kiaraarose (approved):

Issue #902, opened 2025-11-09
Proposal: `browser.documents` namespace


carlosjeurissen:
### Intro
Now [documentIds](https://github.com/w3c/webextensions/issues/890) are getting common in browsers and APIs like tabGroups are getting their own namespace, it seems odd for documents not to have the same. Turns out by moving many APIs to `browser.documents` you end up with much more semantic and elegant APIs. This issue explores how such namespace could look like.

### Design
```ts
namespace documents {
    // replacing `browser.webNavigation.getFrame`
    function get(documentId: string): Promise<Document>;

    // see https://github.com/w3c/webextensions/issues/77
    function sendMessage(
          documentId: string,
          message: any,
        ): Promise<any>;
    };

    // replacing https://github.com/w3c/webextensions/pull/586
    // replacing webNavigation.getAllFrames()
    function query(queryInfo: {
      documentIds?: string[];
      tabIds?: number[];
      frameIds?: number[];
      windowIds?: number[];
      url?: string;
      documentLifecycle?: DocumentLifecycle;
      status?: DocumentStatus;
      frameType?: FrameType;
    }): Promise<Document[]>;

    // replacing runtime.getDocumentId()
    // see https://github.com/w3c/webextensions/issues/800
    function getId(context: Window | HTMLIFrameElement): Promise<string>;

    // focus management see https://github.com/w3c/webextensions/pull/817
    function focus(documentId: string): Promise<string>;
}
```

In addition, lifecycle events from webNavigation can be moved to browser.documents. With the exception of `onTabReplaced`, which seems to sit best under `browser.tabs`. These events would only be available with the `webNavigation` API. Effectively deprecating the webNavigation namespace (but not the `webNavigation` permission).

As for permissions, it seems for many of these APIs, the `webNavigation` permission is too dramatic, with the exception of the lifecycle events and potentially parts of the `browser.documents.query` method.

tophf:
`document` and DOM elements of a web page are cross-origin to the extension's process and can't be returned by or sent to the API. The only case where it could work is inside the content script and only for the same-origin documents.

A `window` mini-wrapper that's already exposed to cross-origin frames by the web platform, would obviously work in an extension process's tabs/frames but not in the service worker in chromium's MV3. However this mini-wrapper doesn't allow reading anything meaningful anyway, so it doesn't seem useful either.

[...]

Issue #800, opened 2025-03-28
Proposal: `browser.runtime.getDocumentId()`


carlosjeurissen:
Considering Chrome is hesitant to implement https://github.com/w3c/webextensions/issues/12 due to the current information frameIds provide. And as workaround for workarounds for https://github.com/w3c/webextensions/issues/77

Having a way to request the documentId of a specific frame is very useful for dealing with messaging and assumingly many other use-cases. 

This could be achieved by introducing `browser.runtime.getDocumentId()` in a similar fashion as `browser.runtime.getFrameId()` as specified in https://github.com/w3c/webextensions/issues/12#issuecomment-930545112

See example usage: https://github.com/w3c/webextensions/issues/77#issuecomment-2761498371

This would resolve part of https://github.com/w3c/webextensions/issues/469

xeenon:
I am supportive and would be willing to sponsor.

oliverdunk:
Adding supportive following discussion at Berlin face to face meetup.

Rob--W:
Firefox has not implemented `documentId` yet, but I mentioned this issue at https://bugzilla.mozilla.org/show_bug.cgi?id=1891478 so that we don't lose track of it.

brocef:
Would love to see document IDs supported in Firefox

xeenon:
@carlosjeurissen implemented this for Safari in  https://github.com/WebKit/WebKit/commit/c69d6fa3d7d0ec1fed72124725fae8001bbcdb8e .

Issue #12, opened 2021-06-17
request: allow to retrieve a `frameID` from an `<iframe>` element


therealglazou:
One of the problems we faced at Dashlane is dealing with frameIDs. It's currently super tricky to associate a frameID to a given <iframe> element and it's even too complicated to retrieve a frameID at all since the only existing way to do that ATM is to send a message, from the inner context to that iframe, to the background, collect the frameID there and send it back to the origin iframe.

I think there should be an easy way to retrieve a frameID for a given iframe element, from the enclosing context of that iframe element. Similarly, from the JS context inside an iframe, there should be an easy way to retrieve the current frameID without having to ping-pong the background.

therealglazou:
_Side note_: even if `<embed>` is deprecated, some sites still use it (yeah, fools...). Example: https://www2.rxnt.com/phr/

Getting the frameID for a document embedded through `<embed>` is even more complicated because it's not possible to `postMessage()` the inner window.

dotproto:
> … it's even too complicated to retrieve a frameID at all since the only existing way to do that ATM is …

Another way to do this is to use the [webNavigation.getAllFrames()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/getAllFrames), but this requires the webNavigation permission. Offhand I'm not sure what permission prompt other browsers display for this, but Chrome warns users that this permission allows extensions to "[Read your browsing history](https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#:~:text=%22webnavigation%22)."

Can you share a bit more about why the content script needs to know the ID of its host frame? As opposed to, say, the background script using this data.

Exposing a way for the host frame to get it's own frame ID sounds reasonable. We may also want to investigate a more ergonomic way of getting all frame IDs for a given tab than the current webNavigation approach, as that grants access to much more data than necessary to, say, properly target script injection for a given frame.

therealglazou:
> Can you share a bit more about why the content script needs to know the ID of its host frame? As opposed to, say, the background script using thi data.

[...]

PR #934, opened 2026-01-13
Add `runtime.getDocumentId()` proposal

carlosjeurissen:
A more formal proposal for `runtime.getDocumentId()` as discussed during the 2025 berlin meetup and initially proposed as issue in https://github.com/w3c/webextensions/issues/800.

Discussion points:
- Are we all on board with the method rejecting in all cases when a documentId could not be provided?
- Do we want to accept `null` or `undefined` as alias for `window.self` to get the current document, or do we want to stay explicit? (currently we don't, this aligns with `runtime.getFrameId()`)
- Should we specify the exact documentId type, allowing more strict implementation tests?
- The method may make more sense under the documents or webNavigation namespace, see https://github.com/w3c/webextensions/issues/902. Due to initial objections I kept it at runtime in this proposal to match runtime.getFrameId().

The underscore filename was used for the file name to match the other runtime proposals.

### DIFF
--- a/proposals/runtime_get_document_id.md
+++ b/proposals/runtime_get_document_id.md
@@ -1,0 +1,104 @@
+# Proposal: `runtime.getDocumentId()` API
+
+**Summary**
+
+This API can be used to get the `documentId` based on a window object OR a DOM element like `<iframe>`.
+
+**Document Metadata**
+
+**Author:** carlosjeurissen
+
+**Sponsoring Browser:** Safari
+
+**Contributors:** -
+
+**Created:** 2026-01-13
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/800 https://github.com/w3c/webextensions/issues/469 https://github.com/w3c/webextensions/issues/12
+
+## Motivation
+
+### Objective
+
+Provide a way to get the current `documentId` from a content script.
+
+#### Use Cases
+
+The documentIds can be used for more secure messaging and script injection.
+
+### Known Consumers
+
+Password managers (see https://github.com/w3c/webextensions/issues/12)
+Privacy extensions
+AdFiltering extensions
+UI Wrapper extensions
+
+## Specification
+
+### Schema
+
+```typescript
+namespace runtime {
+  function getDocumentId(
+    target: WindowProxy | HTMLIFrameElement | HTMLFrameElement
+  ): string;
+}
+```
+
+### Behavior
+
+If the target has an incorrect type OR if the browser for whatever reason can
+not provide the document id, an exception should be thrown. There is no such
+concept as a `DOCUMENT_ID_NONE`.
+
+### New Permissions
+
+No new permissions are needed. If an extension author already has access to the
+window object or <iframe> element of a frame, knowing the documentId should not
+be of any concern.
+
+### Manifest File Changes
+
+No manifest file changes
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+None.
+
+### Abuse Mitigations
+
+Extensions already have systems in place for frame access. Knowing the
+documentId would not allow you to access the frame without having
+host permissions.
+
+### Additional Security Considerations
+
+Exposing the documentId should not result in any additional security issues.
+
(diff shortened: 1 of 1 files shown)


carlosjeurissen:
@oliverdunk given Chrome supports documentIds and `runtime.getDocumentId()` now has an existing implementation (Safari), can you take a look at this proposal from Googles side?

@zombie @Rob--W Can either of you (mentioning @zombie for due to his `runtime.getFrameId()` proposal (https://github.com/w3c/webextensions/issues/12#issuecomment-930545112) take a look at this proposal?

mnassabain:
Hello, if I may offer my perspective, this feature would work well for my use case.

Our app injects a content script into the root document and into each frame on the page. These scripts currently communicate by sending messages to the background script, which then tries to find the target frame from `webNavigation.getAllFrames`. However, this has some unsupported edge cases, for example if a parent has 2 identical child frames.

Being able to get the target frame's `documentId` directly inside the content script and include it in the message would make the background script's job much more reliable, since the id is unique.

Just to confirm: requesting the `documentId` of a cross-origin child frame wouldn't raise errors, right?

xeenon (approved):

carlosjeurissen:
@mnassabain [said](https://github.com/w3c/webextensions/pull/934#issuecomment-3871810187):
> Just to confirm: requesting the `documentId` of a cross-origin child frame wouldn't raise errors, right?

Correct. In the current webkit implementation (which you can try out in [STP237](https://developer.apple.com/documentation/safari-technology-preview-release-notes/stp-release-237)), using the `runtime.getDocumentId()` API works for cross-origin child frames and does not throw any exceptions. This works both from content scripts running in the top level document and from content scripts running in subframe documents.

proposals/runtime_get_document_id.md
```diff
@@ -0,0 +1,102 @@
+# Proposal: `runtime.getDocumentId()` API>
```
oliverdunk:
```suggestion
# Proposal: `runtime.getDocumentId()` API
```

proposals/runtime_get_document_id.md
```diff
@@ -0,0 +1,102 @@
...
+**Created:** 2026-01-13
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/800 https://github.com/w3c/webextensions/issues/469 https://github.com/w3c/webextensions/issues/12
+
+## Motivation
+
+### Objective
+
+Provide a way to get the current `documentId` from a content script
+What does this API enable?  Why do we need it?
```
oliverdunk:
```suggestion
```

proposals/runtime_get_document_id.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Additional Security Considerations
+
+Exposing the documentId should not result in any additional security issues.
+
+## Alternatives
+
+### Existing Workarounds
+
+Sending a window message to the frame, and having a content script in the frame
```
oliverdunk: I'm not too concerned about this (I'm convinced this is a useful API) but I find this hard to follow. Can we rewrite it and make it clearer what the objective is and how this flow works to accomplish that?

oliverdunk (changes requested):
This generally looks good to me, other than a few notes. This hasn't been through security review on the Chrome side which would be a requirement before we could accept a patch, but I don't think we need to block on that since there is implementor interest from other browsers.

proposals/runtime_get_document_id.md
carlosjeurissen: Sure, updated to be easier to follow. Yet as the workaround in itself is quite complicated, not sure it can be clarified more than currently is. Happy to take suggestions though!

carlosjeurissen:
@oliverdunk Thanks for taking a look and clarifying the current Chrome situation. I have incorporated your suggestions. Thanks!

oliverdunk (approved):
LGTM - thanks Carlos!

zombie:
This looks good, we wanna support it in Firefox.

PR #942, opened 2026-01-23
Proposal: `i18n.getAvailableLanguages()`

carlosjeurissen:
Corresponding issue: https://github.com/w3c/webextensions/issues/995

motivation for separate PR:
While a similar method has been proposed in https://github.com/w3c/webextensions/pull/641/changes#diff-a7a555f6d716353fea0a7a15523a1639b33556d7dee81e1e531c320a4f593fcfR70 as `i18n.getAllLanguages`. I believe splitting it into a separate proposal allows us to move faster for this specific method, considering just this method is already useful without `i18n.setLanguage()` and would be able to be used with the `i18n.withLanguage()` proposal. See https://github.com/w3c/webextensions/issues/274.

A separate PR also allows us to dive deeper in the specifics of this method which seem very valuable.

open ends:
- should any other transformation other than _ to - be applied? ie, iw -> he?
  -> probably not, as this requires more advanced language negotiations in `i18n.setLanguage()` and `i18n.withLanguage()` and removes the use case for using fetch('messages.json').
- what if an invalid locale tag is used, like "_locales/b/messages.json"? Or a
  directory exists which does not contain a messages.json file?
  -> this seems to be mostly already handled when browsers validate the extension.
- what if the browser never intents to support a specific language in
  i18n.setLanguage('iw') or i18n.withLanguage('iw')? Should it be excluded
  in getAvailableLanguages?
  -> seems there are no real reasons a browser would avoid specific language tags with the exception of language tags not following bcp47 standards.
- do we want to reject if the extension is not localised, or return an empty array or return an array with the browser locale?
  -> when the extension is not localised, this method has zero use. It seems most intuitive to return a rejected promise in this case.
- do we want the returned list to be sorted alphabetically? Potentially put the `default_locale` at the top?
  -> sorting alphabetically by default seem to be enough to have same behaviour cross-browser. On the other hand, extension developers can easily use .sort() if requested. And may want to eventually sort the list based on human name.
- what about the name? @hanguokai suggested `i18n.getAllLanguages()`, @oliverdunk suggested `i18n.getParsedLocalesDirectories` before.
  -> other alternatives could be `i18n.getAvailableResourceLanguages()`. `All` and `Available` could be ambiguous.

### DIFF
--- a/proposals/i18n-get-available-languages.md
+++ b/proposals/i18n-get-available-languages.md
@@ -1,0 +1,148 @@
+# Proposal: `i18n.getAvailableLanguages()`
+
+**Summary**
+
+An API which allows extensions to get a list of BCP47 language tags containing
+the languages the browser can use and the extension provides resource files for.
+
+**Document Metadata**
+
+**Author:** carlosjeurissen
+
+**Sponsoring Browser:** TBD
+
+**Contributors:** hanguokai
+
+**Created:** 2025-01-23
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/995, https://github.com/w3c/webextensions/issues/258, https://github.com/w3c/webextensions/issues/274
+
+## Motivation
+
+### Objective
+
+Currently extensions can not easily get a list of supported languages without
+hardcoding them first or need to rely on fetch try and error.
+
+#### Use Cases
+
+Allow to list the languages for extension settings language pickers.
+
+Use the list of supported languages to match a specific website when applying
+a content script.
+
+Have this list independent of distribution (As some extension stores disallow
+certain language tags).
+
+Allow third parties to create polyfills for `i18n.withLanguage()` and
+`i18n.setLanguage()`.
+
+### Known Consumers
+
+Extensions who currently have to maintain a manually currated list of supported
+languages to offer content script language matching and/or allowing to change
+the extension language in the extension settings.
+
+An example of content-script matching is "Category Tabs for Google Keep", which
+needs to match it's language with the page in question to properly integrate.
+
+A simple naive implementation would be:
+```js
+const [pageLanguage, supportedLanguages] = await Promise.all([
+  tabs.detectLanguage(),
+  i18n.getAvailableLanguages(),
+]);
+const resourceLanguage = supportedLanguages.includes(pageLanguage);
+const resourceFile = chrome.runtime.getURL(`_locales/${resourceLanguage}/messages.json`);
+const messages = await fetch(resourceFile);
+```
+
+Future consumers of future related APIs like the `i18n.withLanguage()` and
+`i18n.setLanguage()` APIs.
+
+## Specification
+
+### Schema
+
+```typescript
+namespace i18n {
+  function getAvailableLanguages(): Promise<string[]>;
+}
+```
+
+This API should also be exposed in content scripts.
+
+### Behavior
+
+Given an extension with the following resource files:
+`_locales/en/messages.json`, `_locales/es/messages.json` and
+`_locales/es_419/messages.json`.
+
+The `i18n.getAvailableLanguages()` method would return a promise which resolves
+with the following array: `['en', 'es', 'es-419']`.
+
+If no resource files are specified, and/or `default_locale` in the manifest is
+missing, the promise will reject with an error stating extension localisation
+is required for this method to work properly.
+
+If a browser is unable to offer a specific language tag in `i18n.withLanguage()`
+or `i18n.setLanguage()`, for example, because it does not understand the tag
+(think about `_locales/n/messages.json`), or if a directory has no messages.json
+file, it should not include said language tag. This does not mean a browser
+should only return languages the browser UI itself supports. An extension could
+offer additional languages outside the set of languages used for its own browser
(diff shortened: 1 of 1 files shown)


proposals/i18n-get-available-languages.md
```diff
@@ -0,0 +1,132 @@
...
+certain language tags).
+
+### Known Consumers
+
+Extensions who currently have to maintain a manually currated list of supported
+languages to offer content script language matching and/or allowing to change
+the extension language in the extension settings.
+
+An example of content-script matching is "Category Tabs for Google Keep", which
+needs to match it's language with the page in question to properly integrate.
```
Rob--W, line 47: Could you show an example of how the API would be used in practice for this use case?

I'm wondering what a browser-defined API can provide that is not already possible by the extension. I'd expect that the extension can pre-generate a list of supported languages in relation to the files output to `_locales/`, which would be far more efficient than what a browser can offer.

You have already mentioned this possibility in the "Existing Workarounds", so it is clear that you are aware of the possibility.

proposals/i18n-get-available-languages.md
```diff
@@ -0,0 +1,132 @@
...
+`_locales/en/messages.json`, `_locales/es/messages.json` and
+`_locales/es_419/messages.json`.
+
+The `i18n.getAvailableLanguages()` method would return a promise which resolves
+with the following array: `['en', 'es', 'es-419']`.
+
+If no resource files are specified, and/or `default_locale` in the manifest is
+missing, the promise will reject with an error stating extension localisation
+is required for this method to work properly.
+
```
Rob--W, line 87: Based on the description from the meeting, it sounds like this is intended to be the intersection of languages that the browser recognizes, and languages that the extension includes. Is this the case? If so, let's describe that here for clarity.

proposals/i18n-get-available-languages.md
```diff
@@ -0,0 +1,132 @@
...
+Use the list of supported languages to match a specific website when applying
+a content script.
+
+Have this list independent of distribution (As some extension stores disallow
+certain language tags).
+
+### Known Consumers
+
+Extensions who currently have to maintain a manually currated list of supported
+languages to offer content script language matching and/or allowing to change
```
Rob--W, line 43: What is the relevance of content scripts? Is it generally to match the language of the web page? (maybe with `tabs.detectLanguage`?)

And since `i18n` as a namespace is available to content scripts, would be good to emphasize whether the proposed method is available to content scripts.

proposals/i18n-get-available-languages.md
carlosjeurissen: @Rob--W thanks for taking a look at the proposal!

That is correct, I have updated the proposal to more clearly define how browsers should handle invalid/unrecognised tags and missing messages.json files.

proposals/i18n-get-available-languages.md
carlosjeurissen: Updated the proposal to include the practical example.

The browser-defined API would allow to list the intersection of languages the browser recognises and the extension includes. As mentioned, the set of resource files can be different depending on the extension store.

For Naver Whale specifically, I have a build script which removes unsupported resource files from the extension package after being built. There are even custom scenarios in which too many resource files requires the removal of the majority of resource files. In these scenarios, updating the set of supported languages in some JS file would add another layer of already complex build setup.

proposals/i18n-get-available-languages.md
carlosjeurissen: @Rob--W added `tabs.detectLanguage` as example in the proposal. Also clarified the API should be available in content scripts. Thanks for your feedback!

proposals/i18n-get-available-languages.md
oliverdunk: Similar to Rob, I'm not convinced that there are use cases which justify the addition of this API.

Adding a build step which exposes the bundled languages to your extension at runtime seems reasonable to me. The API would not provide anything which you could not also achieve in this way.

I can see value if we add the ability to set your extension's locale programatically. At that point, it is useful to know which of your locale directories the browser has parsed. However, it would make sense to add both APIs at the same time (as we discussed in https://github.com/w3c/webextensions/pull/641), instead of just adding this API in isolation.

carlosjeurissen: @oliverdunk thanks for taking a look at the proposal!

First of all, https://github.com/w3c/webextensions/issues/641 has been stalled for more than a year and has not seen any progress since. Other APIs like the proposed `i18n.withLanguage()` would go well with `i18n.getAvailableLanguages()`. Having `i18n.getAvailableLanguages()` stuck in the `i18n.setLanguage()` proposal could block it's usage for `i18n.withLanguage()`.

There are many reasons this API is useful in isolation even before `i18n.withLanguage()` and `i18n.setLanguage()` are available. besides the build step complexity listed above and the extension store compatibility. There is also the use case of being able to polyfill `i18n.withLanguage()` and `i18n.setLanguage()`. With `i18n.getAvailableLanguages()` one would be able to have a third party dependency polyfilling these potential future APIs. Thus especially now when `i18n.withLanguage()` and `i18n.setLanguage()` are not available, `i18n.getAvailableLanguages()` would unlock these use cases.

Issue #960, opened 2026-03-10
Proposal: limit length of extension action badge text


bershanskiy:
Updated: 24.04.26.

## Proposal

I propose to implement a simple check for the length of input to [`setBadgeText()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/setBadgeText). I believe the API should limit the number of bytes in the input string, not the character or render width, since those checks are more complex and depend on font rendering and particular style of UI which would differ between browsers.

## Background

Extensions can set a custom badge (text) which is displayed over the extension action button in browser chrome. This UI element is very small and can display only a few characters or symbols or emojis, may be a few words written in Chinese or Korean. Despite clear visual limitations, [`setBadgeText()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/setBadgeText) method accepts inputs of unbounded length.

1. Extensions can (accidentally) create long strings and consume the browser resources while all performance profilers would miss-attribute this memory usage to browser process.
2. Unlikely, but theoretically possible that extensions would use badge to store data. This may be convenient, since the actual storage APIs impose per-item limit and do not natively support keying storage by Tab ID and Window ID.

## Considerations

1. Limit should be very permissive to enable extension developers' creativity.
2. Valid strings may contain multi-byte characters, including 4-byte emojis and emojis combined with 1-byte zero-width joiner (ZWJ). A single emoji rendered can be 19 bytes long if it combines four 4-byte emojis with ZWJ between them. Here is a concrete example of [19-byte string](https://emojipedia.org/family-man-medium-light-skin-tone-woman-medium-light-skin-tone-girl-medium-light-skin-tone-boy-medium-light-skin-tone) rendered by Chrome and Firefox as a single character. Only two such emojis fit in current Chrome UI for total length of 38 bytes.
3. Future or alternative browser UI revisions might be able to display more characters. Current Chrome UI can comfortably display only 4 digit numbers or 12 dots.

To ensure broad compatibility, we should introduce a histogram tracking the length of real-world badge strings.

## Next actions

1. Wait for histogram data - Oliver
2. Determine a suitable limit - Oliver
3. Implement experiment with a limit - Anton
4. Run experiment - TBD
5. Remove experiment and histogram code - Anton

## External links
[Chromium bug 491158086](https://issues.chromium.org/issues/491158086)
[...]

Issue #957, opened 2026-02-26
Agenda discussion for public meeting on 2026-03-12


Rob--W:
This issue is used to collect topic suggestions for our March 12, 2026 meeting. We will stop accepting suggestions ~48 hours before the meeting and the final agenda will be added to the [meeting notes doc](https://docs.google.com/document/d/1QkwhEMtMS67JBUkl_WVPZ4lRSKoWcQNlLJSf_GwSXg8/edit) ~24 hours before the meeting.

The content below this line will be copy-pasted to the agenda section of the meeting.

## Agenda

The meeting will start at 3 minutes after the hour.
See [issue 531](https://github.com/w3c/webextensions/issues/531) for an explanation of this agenda format.

 * **Announcements** (2 minutes)
   * [2026 WECG Face to Face](https://github.com/w3c/webextensions/issues/951)
 * **Triage** (15 minutes)
   * [Issue 960](https://github.com/w3c/webextensions/issues/960): Proposal: limit length of extension action badge text
* **Timely issues** (10 minutes)
 * **Check-in on existing issues** (20 minutes)
   * [Issue 935](https://github.com/w3c/webextensions/issues/935): Proposal: Limits on lengths of strings passed to WebExtension APIs
   * [Issue 955](https://github.com/w3c/webextensions/issues/955): Alert users when an extension is disabled programmatically
   * [Issue 934](https://github.com/w3c/webextensions/pull/934): Add runtime.getDocumentId() proposal

Rob--W:
Warning: The next meeting is scheduled in Pacific time, and the daylight saving time changes (PST -> PDT). In Europe, the next two meetings shift by one hour (e.g. in Amsterdam, Berlin, etc., the meeting happen at 4 PM instead of 5 PM).

bershanskiy:
It would be nice to align on limits for alarm names passed to `alarms.create({name})` (#935) either async or during the meeting.

Issue #955, opened 2026-02-26
Alert users when an extension is disabled programmatically


bbruneau:
Given the recent publicity of [polymorphic extension attacks](https://passpack.com/polymorphic-browser-extension-attacks/), it would seem that there is utility in providing some feedback to the user in the case that an extension is disabled by another extension via the `management` API.

Issues have been raised previously that would help ameliorate this kind of situation, such as #836 (Allow extensions to track their own disable/uninstall events) and #941 (Enhancement: management.ExtensionInfo.disabledReason - add value that indicates the extension has been disabled due to incompatibility or even just client-specific reason), but browser vendors have indicated during discussions that these issues introduce new APIs with their own complications and possible abuses, making them low priority at best, or undesirable at worst.

I think it would be sufficient, however, to have a browser-controlled alert appear when a given extension is disabled programmatically, something as simple as a popup alert stating `"Extension ABC has been disabled."` This would create a moment of friction that could serve to alert the user that something fishy may be occurring, or at least that something unexpected has occurred. An alert in the browser's own livery would both indicate that the message came from a trusted place and obviate situations where abusive extensions could attempt to re-enable themselves or send tracking data.

An even better alert might provide the names and IDs of both the disabled extension, and the extension that disabled it, providing further context about suspicious extension activity. In the case of a spoofing attack, a message stating `"The extension Legitimate Password Manager (id: 123) was disabled by the extension Totally Not A Malware Extension (id: 456)"` would provide plenty of actionable information to the user, without giving new powers to extensions.

xeenon:
Neutral for Safari since we don't support the management APIs for this reason.

Rob--W:
FYI in Firefox `management.setEnabled` for extensions is not supported yet, but might be with explicit user interaction. We recently got a patch from a contributor implementing that at https://bugzilla.mozilla.org/show_bug.cgi?id=1282982

Rob--W:
I put https://github.com/w3c/webextensions/labels/supportive%3A%20firefox on the issue, because on the Firefox side I want to require user input before `management.setEnabled` acts upon the request. That fits the description "Alert users when an extension is disabled programmatically"

Issue #951, opened 2026-02-05
2026 WECG Face to Face (London)


dotproto:
The WebExtensions Community Group is holding an hybrid (in-person and online) Face to Face meeting  London, United Kingdom from April 7–9, 2026, hosted by Google. This issue will serve as our main thread for planning and scheduling our activities during this event.

Please fill out [2026 WECG London Face to Face Signup](https://docs.google.com/forms/d/e/1FAIpQLSedMSYmPll5cLMewS-Fn-HI2yD27OxNVzdVClYWDGF5lYcMNQ/viewform) if you're considering attending, and we'll follow up with more logistics details.

More information: https://github.com/w3c/webextensions/wiki/2026-London-F2F-Coordination

dotproto:
Topic suggestion from David Johnson: community curated cross-browser extension mapping.

dotproto:
@jspivack posted to the [#wecg](https://matrix.to/#/#wecg:mozilla.org) room to raise the topic of WebMCP integration. Might be worth spending some time ideating how to support ML/automation.

dotproto:
Following up on my comment from Feb 26, the [Web Machine Learning Community Group](https://www.w3.org/groups/cg/webmachinelearning/) ([homepage](https://webmachinelearning.github.io/)) has a [draft WebMCP spec](https://webmachinelearning.github.io/webmcp/). In webmachinelearning/webmcp#74, the group has started discussions about collaborating with us (WECG) on exposing WebMCP capabilities to extensions.

Rob--W:
I put up a PR with meeting minutes at https://github.com/w3c/webextensions/pull/979. The PR body also includes an overview of all topics and links all discussed issues and PRs.

Rob--W:
The meeting minutes at https://github.com/w3c/webextensions/pull/979 got merged, with that we can close this. See you all at the next meetup!

PR #962, opened 2026-03-12
Publish minutes of 2026-03-12 meeting

Rob--W:

During this meeting we discussed or mentioned issues #957, #531, #951, #960, #935, #955 and PR #934.

Note: The Pacific time zone has recently shifted from Daylight Saving Time, but in Europe the Daylight Saving Time is still in effect. Double-check the time of the meeting if needed.

### DIFF
--- a/_minutes/2026-03-12-wecg.md
+++ b/_minutes/2026-03-12-wecg.md
@@ -1,0 +1,130 @@
+# WECG Meetings 2026, Public Notes, Mar 12
+
+ * Chair: Kiara Rose
+ * Scribes: Rob Wu
+
+Time: 8 AM PDT = https://everytimezone.com/?t=69b35380,384
+Call-in details: [WebExtensions CG, 12th March 2026](https://www.w3.org/events/meetings/6a0eda89-558c-408d-b83d-5f03b8853c30/20260312T080000/)
+Zoom issues? Ping @zombie (Tomislav Jovanovic) in [chat](https://github.com/w3c/webextensions/blob/main/CONTRIBUTING.md#joining-chat)
+
+
+## Agenda: [discussion in #957](https://github.com/w3c/webextensions/issues/957), [github issues](https://github.com/w3c/webextensions/issues)
+
+The meeting will start at 3 minutes after the hour.
+
+See [issue 531](https://github.com/w3c/webextensions/issues/531) for an explanation of this agenda format.
+
+ * **Announcements** (2 minutes)
+   * [2026 London F2F Sign-up](https://docs.google.com/forms/d/e/1FAIpQLSedMSYmPll5cLMewS-Fn-HI2yD27OxNVzdVClYWDGF5lYcMNQ/viewform) ([coordination](https://github.com/w3c/webextensions/wiki/2026-London-F2F-Coordination), [topic discussion](https://github.com/w3c/webextensions/issues/951))
+ * **Triage** (15 minutes)
+   * [Issue 960](https://github.com/w3c/webextensions/issues/960): Proposal: limit length of extension action badge text
+ * **Timely issues** (10 minutes)
+   * N/A
+ * **Check-in on existing issues** (20 minutes)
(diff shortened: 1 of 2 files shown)


_minutes/2026-03-12-wecg.md
```diff
@@ -0,0 +1,130 @@
...
+ * [oliver] Would be interesting to see what we imagine this limit being. If the limit were 50 characters and the max size is 30x that size for large unicode characters, that seems reasonable.
+ * [kiara] Seems reasonable. Having a standard size we can apply to multiple APIs makes sense.
+ * [simeon] Don't think I included it in my last reply, but was thinking about something like 100 characters and 10 grapheme clusters as a combined limit.
+ * [tomislav] Getting into the details of unicode is perilous. Strongly suggest we avoid going that route.
+   * Background reading: https://hsivonen.fi/string-length/
+ * [rob] Can we just truncate, without throwing/rejecting from the API?
+ * [anton] Difficult to know how to correctly extract a sub-string. In order to do it well, you'd have to render the string to determine the effective characters.
+ * [rob] Would you prefer to throw an error or silently truncate?
+ * [anton] Thinking of silently truncating for now. Throwing changes execution flow, truncating changes return value. Maybe change the behavior in a future manifest version?
```
bershanskiy: > silently truncating

I might have miss-spoken or may be I was mis-heard, I meant to say "silently ignoring the value" by just resolving the promise and hiding the badge.

Either way, we are very early in the decision process. We do not even have a histogram for measuring the impact of the change (Chrome is currently working on this histogram). We are a few months way from even writing a console warning, which is the earliest possible publicly visible change.

_minutes/2026-03-12-wecg.md
```diff
@@ -0,0 +1,130 @@
...
+ * [tomislav] Added a doc (above in the note) from a colleague that goes into detail around challenges related to string handling. Suggest we have something like a limit of 1000 characters for the string and a separate limit for the number of characters shown. So two limits: how much we store and how much we display.
+ * [rob] I think that `action.setBadgeText()` is limited in practice, and there are no objections to enforcing some limit somewhere. If we are thinking about reasonable limits for APIs that take strings, we may also want to consider `action.setTitle()`. I tried just now to render 8k characters in Firefox, and a tooltip does show up with all of the specified text.
+ * [anton] Only tested in Chrome, but observed a 1024 byte limit for setTitle.
```
bershanskiy: > observed a 1024 byte limit for setTitle

Observed Chrome accepting unlimited input to `setTitle` and then reproducing it verbatim via `getTitle`, but having a display limit of 1024 bytes.

_minutes/2026-03-12-wecg.md
```diff
@@ -0,0 +1,130 @@
...
+
+ * [kiara] Reminder: London f2f meeting in April, please sign up if you are interested in attending.
+ * [rob] Recommendations for hotels in the area?
+ * [oliver] I found a hotel list. I'll add it to the wiki. The St Pancreas station and Google office have many hotels in the neighborhood.
+ * [tomislav] Is public transport reliable?
+ * [oliver] The underground (tube) is very reliable.
+
+[Issue 960](https://github.com/w3c/webextensions/issues/960): Proposal: limit length of extension action badge text
+
+ * [anton] setBadgeText can receive an arbitrary amount of text. The rendered text is a fraction of the space, no room for much text. Suggest to choose a limit.
```
bershanskiy: > The rendered text is a fraction of the space, no room for much text.

I suggest changing to:
> The rendered text is displayed over extension icon in the browser chrome (as in the part of the browser UI with omnibox), so the rendered badge is smaller than even the extension icon. That works out to be always smaller than 100x100 px in size. There is no room for much text.

_minutes/2026-03-12-wecg.md
Rob--W: I applied your changes; PR notes are not a place for discussion, just fixing up mistakes. I'm glad to see that you added a comment to clarify your position in the specific issue at https://github.com/w3c/webextensions/issues/960#issuecomment-4047518854

Rob--W:
Meeting going to happen soon to day, so I've fixed the notes and will merge it.

PR #975, opened 2026-04-09
Update `runtime.onInvalidated` with unload reason and isolated world restriction

carlosjeurissen:
During the 2026 London F2F triage we talked about making some changes to the `runtime.onInvalidated` event.

Specifically:
- unload reason
- specifically mention this should only be exposed in isolated content script worlds

### DIFF
--- a/proposals/runtime_on_invalidated.md
+++ b/proposals/runtime_on_invalidated.md
@@ -78,13 +78,13 @@ OnInvalidatedReason:
   "type": "string",
   "enum": [
     {"name": "uninstall", "description": "Specifies the event reason as an uninstallation."},
-    {"name": "update", "description": "Specifies the event reason as an extension update."},
-    {"name": "reload", "description": "Specifies the event reason as an extension reloading."},
-    {"name": "disable", "description": "Specifies the event reason as an extension disabling."}
+    {"name": "unload", "description": "Specifies the event reason as an extension unload (like extension updates, reloads, and disabling)."}
   ],
   "description": "The reason that this event is being dispatched."
 }
 ```
+
+The API should only be exposed to isolated content script worlds.
 
 ### Behavior
 
@@ -150,6 +150,9 @@ There is no viable path to turn this into an Open Web API.
 
 The event should be fired in the situation in which extension APIs would throw
 with an context invalidated error.
+
+For this to work, browsers need to keep the content script context alive for
+the extension to handle the invalidation event.
 
 ## Future Work
 


proposals/runtime_on_invalidated.md
```diff
@@ -78,6 +78,7 @@ OnInvalidatedReason:
   "type": "string",
   "enum": [
     {"name": "uninstall", "description": "Specifies the event reason as an uninstallation."},
+    {"name": "unload", "description": "Specifies the event reason as an extension unload (could be update, reload, or disabling)."},
     {"name": "update", "description": "Specifies the event reason as an extension update."},
     {"name": "reload", "description": "Specifies the event reason as an extension reloading."},
     {"name": "disable", "description": "Specifies the event reason as an extension disabling."}
```
rdcronin: Per our chat at the London F2F, these three reasons should be removed (and effectively collapsed into unload)

rdcronin (changes requested):

proposals/runtime_on_invalidated.md
carlosjeurissen: Keeping these reasons allows other browsers who are in favour of further specifying the unload reason to do so. As raised by Oliver, as developer, one would want different behaviour for `update`/`reload` (keep UI until replacement), compared to `disable` (remove UI from the DOM).

Having the unload reason does not necessarily mean the other reasons could not sit alongside it, and could be used in scenarios in which the browser is not sure, does not know, or does not want to share the specific unload reason.

oliverdunk: Similar to Devlin, my understanding was that we concluded the discussion resolving to having two reasons. I'm not supportive of specifying additional reasons that are only implemented by some browsers as that seems like quite a significant inconsistency we would be introducing by design.

oliverdunk: cc/ @Rob--W @xeenon

xeenon: I'm in favor of just `uninstall` and `unload`.

kiaraarose: Agree that only two reasons would be needed here

carlosjeurissen: @rdcronin removed the other reasons in the PR. Thanks for the note.

carlosjeurissen:
@Rob--W updated the pr to mention the need for browsers to keep the content script context alive.

proposals/runtime_on_invalidated.md
```diff
@@ -78,14 +78,14 @@ OnInvalidatedReason:
   "type": "string",
   "enum": [
     {"name": "uninstall", "description": "Specifies the event reason as an uninstallation."},
-    {"name": "update", "description": "Specifies the event reason as an extension update."},
-    {"name": "reload", "description": "Specifies the event reason as an extension reloading."},
-    {"name": "disable", "description": "Specifies the event reason as an extension disabling."}
+    {"name": "unload", "description": "Specifies the event reason as an extension unload (like extension updates, reloads, and disabling)."},
```
Rob--W: Trailing comma is technically invalid JSON.

```suggestion
    {"name": "unload", "description": "Specifies the event reason as an extension unload (like extension updates, reloads, and disabling)."}
```

Rob--W (approved):
Thanks!

proposals/runtime_on_invalidated.md
carlosjeurissen: Good catch! Resolved

oliverdunk (approved):
LGTM - thanks!

rdcronin (approved):

kiaraarose (approved):

Issue #456, opened 2023-09-25
Proposal: Provide reporting event for browser's slow extension warnings


kzar:
## Background
Browsers sometimes display a warning to the user that one of the browser extensions that they have enabled is slowing down their experience. Browsers seem to trigger such warnings under different circumstances, and there doesn't appear to be much documentation around this, or ways for extension developers to diagnose the performance issues. It would be nice to support extension developers in proactively improving their extension's performance.

## Suggestion
- Provide some way of reporting performance issues to extension developers in an actionable way. For example, perhaps an event could be fired that included at least some details (e.g. a stack trace) of the slow extension code? I realise this would need to be balanced with user privacy however.
- I believe that so far, Firefox will warn the users if a content script is extremely slow (and the user tries to interact with the page), whereas Chrome will warn the user if a background script (in particular inside a blocking `onBeforeRequest` event handler) is slow. Perhaps the event could also include a description (e.g. "Slow content script", or "Slow onBeforeRequest handler") to point developers in the right direction?

## References
- https://searchfox.org/mozilla-central/source/dom/ipc/ProcessHangMonitor.cpp#63-86
  - `"EXTENSION NAME" is slowing down Firefox. To speed up your browser, stop that extension. Learn more <Stop>`
- https://source.chromium.org/chromium/chromium/src/+/main:extensions/browser/warning_set.cc;l=71-80;bpv=1;bpt=1
  - `This extension is slowing down Google Chrome. You should disable it to restore Google Chrome's performance.`

patrickkettner:
May I suggest you change the issue to be just about the reporting portion? While I agree with everything, I do not believe this is the right venue you for the conversation for topics other than the reporting piece. The WECG does not provide documentation, and what actually triggers such a warning is an implementation detail, rather than something that would normally be covered by standards.

kzar:
Thanks for the feedback @patrickkettner :+1: , I've updated the issue.

dotproto:
While this group has not produced documentation and we don't have immediate plans to do so, I should note that that kind of material is within the scope of community group deliverables:

> Community Group deliverables may be anything, including documents, test suites, tutorials, demos, code, discussion, etc.
>
> \- https://www.w3.org/community/about/process/#deliverables

> For instance, a Community Group might … convene to have discussions about a tutorial for an existing specification.
>
> \- https://www.w3.org/community/about/faq/#what-is-a-w3c-community-group

[...]

PR #989, opened 2026-04-27
Proposal: Asynchronous Listener Registration

AndreaOrru:
Currently, the WebExtensions platform requires event listeners to be registered synchronously during a service worker's initial script evaluation. This makes e.g. conditional listener registration based on asynchronous state (like `browser.storage`) brittle, forcing developers to rely on complex workarounds.

This proposal introduces an `"async_initialization": true` manifest opt-in and a new `browser.runtime.markInitializationComplete()` API to address this issue.

### DIFF
--- a/proposals/async-listener-registration.md
+++ b/proposals/async-listener-registration.md
@@ -1,0 +1,237 @@
+# Proposal: Asynchronous Listener Registration
+
+**Summary**
+
+An API proposal to allow extensions to explicitly manage their service worker initialization phase via a manifest opt-in field and a `markInitializationComplete()` API, enabling the asynchronous registration of event listeners. 
+
+**Document Metadata**
+
+**Author:** andreaorru@chromium.org
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:** rdevlin.cronin@chromium.org
+
+**Created:** 2026-04-22
+
+**Related Issues:** N/A
+
+## Motivation
+
+### Objective
+
+Currently, the extension platform requires event listeners to be registered synchronously during the execution of a service worker script. The browser process manages event routing and relies on the renderer process to report which events an extension is listening to immediately after the initial script evaluation.
+
+If an extension registers a listener asynchronously (e.g., inside a `setTimeout`, after a `fetch()`, or after reading from `browser.storage`), the initial script evaluation finishes before the listener is attached. Consequently, when the relevant event fires later, the service worker wakes up, executes its top-level script synchronously, and the browser dispatches the event to all registered listeners (of which there may be none). Thus, the service worker wakes up but receives no event.
+
+This proposal introduces a mechanism to explicitly delay the finalization of the "initialization phase" of the worker and formalizes how listener state is managed and persisted across service worker restarts.
+
+#### Use Cases
+
+* **Conditional event registration:** An extension may need to read user preferences from local storage to determine whether it should register listeners for high-traffic events (e.g., `tabs.onUpdated` or `webRequest`).
+* **Dynamic configuration:** An extension might need to fetch a configuration file to decide which listener options to set up (e.g. `webRequest` URL filters).
+
+### Known Consumers
+
+Extensions which require dynamic or conditionally registered event listeners based on asynchronous state, without resorting to unconditionally waking up the service worker for events they may not care about.
+
+## Specification
+
+### Schema
+
+#### Manifest
+
+```json
+{
+  "background": {
+    "service_worker": "background.js",
+    "async_initialization": true
+  }
+}
+```
+
+#### Runtime API
+
+```typescript
+namespace runtime {
+  /**
+   * Signals to the browser that asynchronous initialization is complete.
+   * Transitions the service worker out of the initialization phase, replaces 
+   * the previous run's persisted listeners with the newly registered ones, 
+   * and flushes the event queue.
+   */  
+  export function markInitializationComplete(): void;
+}
+```
+
+### Behavior
+
+#### Initialization Lifecycle
+
+* **Manifest opt-in:** Extensions declare `"async_initialization": true` nested inside the `"background"` key in their `manifest.json`.
+* **Renderer-side queueing:** During the initialization phase, the browser routes incoming events that match the previously registered listeners to the service worker's renderer process, which queues them instead of immediately attempting to dispatch them to JavaScript callbacks (which might not be attached yet).
+* **Event routing (union of old and new):** While in the initialization phase, the browser will route an event to the renderer queue if it matches *either* the old persisted routing state (from the previous run) or any new listeners currently being registered by the executing script.
+* **Listener registration:** Developers perform their asynchronous setup (e.g., reading from storage) and call `addListener()` to attach their desired JavaScript callbacks *before* calling the completion API.
+* **State commit & completion API:** The extension must call `browser.runtime.markInitializationComplete()` once its async setup is complete. At that moment:
+	1. The old persisted routing state is **thrown away**.
+	2. The new listeners currently registered are committed to the browser as the new persisted routing state for future wake-ups.
+* **Event flush:** Once `markInitializationComplete()` ends the initialization phase, the renderer flushes its queue, dispatching all held events to the newly registered JavaScript callbacks in FIFO order.
+
+#### Example Flow
+
+To illustrate how these lifecycle changes resolve the timing issues of asynchronous setup, consider an extension that conditionally tracks tab updates based on an asynchronous storage read.
+
+```javascript
+browser.storage.local.get("shouldListen").then((settings) => {
+  // Asynchronously evaluate state.
+  if (settings?.shouldListen) {
+    // Conditionally register the listener based on the async result.
+    browser.tabs.onUpdated.addListener(handleTabUpdate);
+  }
+  // Signal the end of the initialization phase.
+  browser.runtime.markInitializationComplete();
+});
+```
+
+Assume that during a previous run, `settings.shouldListen` was `true`, the listener was registered, and the browser persisted this routing state. The service worker has since been terminated. 
+
+When a new `tabs.onUpdated` event fires, the browser checks the persisted state and wakes the service worker. Because the manifest includes `"async_initialization": true`, the renderer process queues the waking event rather than attempting to dispatch it immediately. 
+
+From here, execution resolves into one of two cases.
+
+##### Case 1: The listener is registered
+
+If the storage promise resolves and `settings.shouldListen` evaluates to `true`:
+1. The extension executes the `addListener()` call.
+2. The extension calls `browser.runtime.markInitializationComplete()`.
+3. The browser commits the routing state (maintaining the `tabs.onUpdated` registration for future wake-ups). The renderer flushes its holding queue, and the queued `tabs.onUpdated` event is safely dispatched to the newly attached `handleTabUpdate` callback. No events were missed.
+
+##### Case 2: The listener is not registered
+
+If the storage promise resolves but `settings.shouldListen` evaluates to `false`:
+1. The extension skips the `addListener()` call.
+2. The extension calls `browser.runtime.markInitializationComplete()`.
+3. The browser commits the routing state. Because the listener was omitted during this run, the browser overwrites and **clears** the old persisted routing state. The renderer flushes the queued event, which is harmlessly dropped since no JavaScript callback is attached.
+4. Because the routing state is now cleared, future `tabs.onUpdated` events will no longer wake up the service worker. This implicitly cleans up the routing state without the developer needing to explicitly call `removeListener()` (see Workaround B) or accept unnecessary wake-ups on future events (see Workaround A).
+
+#### Edge Cases
+
+* **Initialization failure fallback:** If the initialization phase fails (e.g., the worker crashes or hits a timeout before the completion API is called), the browser will abort the state commit. It will keep the previous persisted routing state to prevent leaving the extension in a broken state.
+* **Late registration:** It is permissible to register new listeners after the `markInitializationComplete()` API call. If an extension does so, developers must understand the following lifecycle:
+	* The browser will persist these "late" listeners across service worker terminations to successfully trigger a future wake-up.
+	* On the subsequent wake-up, all relevant listeners must be re-registered. Calling `markInitializationComplete()` entirely replaces the routing state from the previous run.
+	* Therefore, any late-registered listeners from a previous run will be removed from the registered listeners during the next run's state commit unless the extension explicitly re-registers them before calling the completion API.
+
+#### Performance Considerations
+
+Because MV3 service workers are ephemeral and wake up frequently, developers must be aware that delaying initialization adds latency to event dispatch. If an extension relies on slow I/O before calling `markInitializationComplete()`, that latency penalty applies to every single event that triggers a cold start. Developers are heavily encouraged to use fast, local I/O for conditional listener registration during routine wake-ups.
+
+### New Permissions
+
+| Permission Added | Suggested Warning |
+| ---------------- | ----------------- |
+| N/A              | N/A               |
+
+This API modifies the initialization lifecycle of the extension's service worker but does not grant any new access to user data or privileged browser features. Therefore, no new permissions are required.
+
+### Manifest File Changes
+
+Add an optional (default false) `async_initialization` field inside the `background` key.
+
+```json
+{
+  "background": {
+    "service_worker": "background.js",
+    "async_initialization": true
+  }
+}
+```
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+This API does not expose any sensitive data or personally identifiable information.
+
+### Abuse Mitigations
+
+* **Indefinite hangs:** A malicious or poorly coded extension could fail to call `markInitializationComplete()` in an attempt to consume memory by forcing the browser to queue a massive number of events. This is mitigated by queuing the events in the renderer process, scoping the impact to the renderer only. The browser can also enforce an initialization timeout, falling back to the previous run's routing state if it expires.
+* **Queue limits:** Implementations could enforce an upper bound on the event queue size, dropping the oldest events if the limit is exceeded during the wait phase to prevent unbounded memory growth. Again, because the queuing happens renderer-side, in the worst case scenario a misbehaving renderer can be killed.
+
+## Alternatives
+
+### Existing Workarounds
+
+Currently, developers must work around the lack of async initialization using patterns that are suboptimal.
+
+#### Workaround A: Discarding events inside the listener callback
+
+Developers register the listener synchronously, load their configuration asynchronously, and check the state *inside* the callback.
+
+```javascript
+let config = null;
+browser.storage.local.get("config").then(res => config = res);
+
+browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+  if (!config || !config.enableTabTracking) return; // Discard event.
+  // ... actual logic ...
+});
+```
+
+This incurs significant performance overhead. It forces the browser to spin up the service worker environment and dispatch the event every single time it fires, only for the extension to immediately discard it.
+
+#### Workaround B: The `removeListener` pattern
+
+Developers register the listener unconditionally on startup. Once the asynchronous state resolves, if the feature is disabled, they explicitly unregister the listener.
+
+```javascript
+const handleTabUpdate = (tabId, changeInfo, tab) => { /* logic */ };
+browser.tabs.onUpdated.addListener(handleTabUpdate);
+
+browser.storage.local.get("config").then(config => {
+  if (!config.enableTabTracking) {
+    browser.tabs.onUpdated.removeListener(handleTabUpdate);
+  }
+});
+```
+
+While this prevents *future* unnecessary wake-ups (as the browser persists the updated routing table), it still incurs an initial wake-up penalty. It also requires excessive boilerplate to manually map, re-add, and remove listener references as state changes, complicating extension logic. The proposed API avoids both the initial wake-up penalty and the complex teardown boilerplate, as old listeners are automatically replaced upon completion of initialization.
+
+#### Workaround C: Listener substitution hack for dynamic filters
+
+When an extension relies on dynamically configured event filters (e.g., user-defined URLs for `webNavigation`), it cannot register the filtered listener synchronously on a cold start because the filters haven't been loaded from storage yet. To avoid missing the event that triggered the service worker's wake-up, the extension must register a broad, unfiltered listener to act as a temporary catch-all, and later substitute it.
+
+```javascript
+// 1. Synchronously register an unfiltered catch-all to receive the waking event.
+const catchAllListener = (details) => { /* temporarily handle or queue event */ };
+browser.webNavigation.onBeforeNavigate.addListener(catchAllListener);
+
+// 2. Asynchronously load the dynamic filters.
+browser.storage.local.get("userFilters").then(filters => {
+  // 3. Register the actual filtered listener so the browser's persistent 
+  // event router knows to wake the SW for these URLs in the future.
+  browser.webNavigation.onBeforeNavigate.addListener(
+    (details) => { /* handle actual event */ }, 
+    { url: filters }
+  );
+  
+  // 4. Remove the temporary catch-all listener.
+  browser.webNavigation.onBeforeNavigate.removeListener(catchAllListener);
+});
+```
+
+This pattern (really just a hack) is very convoluted and brittle: it forces the extension to temporarily over-listen, catching potentially irrelevant events while the async task is pending. Notably, the Chromium codebase contains tests specifically designed to ensure this workaround functions (see https://crrev.com/c/7604661), with comments explicitly noting that this is required *"until we have a better way for extensions to register listeners asynchronously at startup."*
+
+### Open Web API
+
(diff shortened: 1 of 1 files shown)


proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
+# Proposal: Asynchronous Listener Registration
+
+**Summary**
+
+An API proposal to allow extensions to explicitly manage their service worker initialization phase via a manifest opt-in field and a `markInitializationComplete()` API, enabling the asynchronous registration of event listeners.
```
Rob--W: "manage their service worker initialization" sounds broad (potentially raising concerns about impact on service workers), but my impression after reading the whole document is that this is really about delaying all extension API event dispatch at the start of the extension service worker wakeup, until the extension signals that it ready to receive (or drop) them.

Notably out of scope are: standard SW lifecycle events, web platform APIs events, non-event.

Is my interpretation accurate? If so, can we make this more explicit in the description and choice of the API names (I'm not suggesting names yet, bikeshedding can be done once we have a common understanding of its semantics).


E.g.

```suggestion
An API proposal to allow extensions to explicitly delay the dispatch of extension API events to their service worker during its initialization phase via a manifest opt-in field and a `markInitializationComplete()` API, enabling the asynchronous registration of event listeners. 
```

P.S. "extension service worker" is not supported by Firefox yet, the predecessor (event pages) are still supported by Firefox and Safari ( https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background#browser_support ).
Where I read "extension service worker" across the proposal, I have the understanding that it refers to this broader concept (perhaps it makes sense to write that in a small note somewhere, for those not in the loop).

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+
+If an extension registers a listener asynchronously (e.g., inside a `setTimeout`, after a `fetch()`, or after reading from `browser.storage`), the initial script evaluation finishes before the listener is attached. Consequently, when the relevant event fires later, the service worker wakes up, executes its top-level script synchronously, and the browser dispatches the event to all registered listeners (of which there may be none). Thus, the service worker wakes up but receives no event.
+
+This proposal introduces a mechanism to explicitly delay the finalization of the "initialization phase" of the worker and formalizes how listener state is managed and persisted across service worker restarts.
+
+#### Use Cases
+
+* **Conditional event registration:** An extension may need to read user preferences from local storage to determine whether it should register listeners for high-traffic events (e.g., `tabs.onUpdated` or `webRequest`).
+* **Dynamic configuration:** An extension might need to fetch a configuration file to decide which listener options to set up (e.g. `webRequest` URL filters).
+
```
Rob--W, line 34: Here is another use case, based on https://bugzilla.mozilla.org/show_bug.cgi?id=1822735 (starting at the fourth paragraph, starting with "The more serious issue"):

```suggestion
* **Atomic startup completion signal**: An extension that partially failed to register its listeners at its first run can have another chance at the next startup, without being in an inconsistent state.

```

Feel free to include a link to the bug in the text. Having concrete examples with more text might sometimes make it easier to understand what a short note is about.

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+#### Runtime API
+
+```typescript
+namespace runtime {
+  /**
+   * Signals to the browser that asynchronous initialization is complete.
+   * Transitions the service worker out of the initialization phase, replaces 
+   * the previous run's persisted listeners with the newly registered ones, 
+   * and flushes the event queue.
```
Rob--W: Since the functionality is specifically for the background context, shall we specify that it is only available there? So that e.g. an extension popup or options page cannot call it. Or do you want to leave that possibility open?

```suggestion
   * and flushes the event queue.
   *
   * This method is only available to the background script context.
```

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Behavior
+
+#### Initialization Lifecycle
+
+* **Manifest opt-in:** Extensions declare `"async_initialization": true` nested inside the `"background"` key in their `manifest.json`.
+* **Renderer-side queueing:** During the initialization phase, the browser routes incoming events that match the previously registered listeners to the service worker's renderer process, which queues them instead of immediately attempting to dispatch them to JavaScript callbacks (which might not be attached yet).
+* **Event routing (union of old and new):** While in the initialization phase, the browser will route an event to the renderer queue if it matches *either* the old persisted routing state (from the previous run) or any new listeners currently being registered by the executing script.
+* **Listener registration:** Developers perform their asynchronous setup (e.g., reading from storage) and call `addListener()` to attach their desired JavaScript callbacks *before* calling the completion API.
+* **State commit & completion API:** The extension must call `browser.runtime.markInitializationComplete()` once its async setup is complete. At that moment:
```
Rob--W: Added "potentially" before "async" again.

```suggestion
* **State commit & completion API:** The extension must call `browser.runtime.markInitializationComplete()` once its potentially async setup is complete. At that moment:
```

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+```
+
+### Behavior
+
+#### Initialization Lifecycle
+
+* **Manifest opt-in:** Extensions declare `"async_initialization": true` nested inside the `"background"` key in their `manifest.json`.
+* **Renderer-side queueing:** During the initialization phase, the browser routes incoming events that match the previously registered listeners to the service worker's renderer process, which queues them instead of immediately attempting to dispatch them to JavaScript callbacks (which might not be attached yet).
+* **Event routing (union of old and new):** While in the initialization phase, the browser will route an event to the renderer queue if it matches *either* the old persisted routing state (from the previous run) or any new listeners currently being registered by the executing script.
+* **Listener registration:** Developers perform their asynchronous setup (e.g., reading from storage) and call `addListener()` to attach their desired JavaScript callbacks *before* calling the completion API.
```
Rob--W: Presumably the setup may still be synchronous, adding "potentially" to signal that.

```suggestion
* **Listener registration:** Developers perform their potentially asynchronous setup (e.g., reading from storage) and call `addListener()` to attach their desired JavaScript callbacks *before* calling the completion API.
```

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+* **State commit & completion API:** The extension must call `browser.runtime.markInitializationComplete()` once its async setup is complete. At that moment:
+	1. The old persisted routing state is **thrown away**.
+	2. The new listeners currently registered are committed to the browser as the new persisted routing state for future wake-ups.
+* **Event flush:** Once `markInitializationComplete()` ends the initialization phase, the renderer flushes its queue, dispatching all held events to the newly registered JavaScript callbacks in FIFO order.
```
Rob--W: What is the expectation when this method is called while the script is still executing synchronously at startup?

My expectation is:
- that there are no queued events at that time (because scripts execute synchronously and is not interrupted/paused to receive events)
- that any events delivered immediately after the script finishes execution are dispatched immediately, without a queue.

This is bordering on the implementation detail, but to encourage extensions to call the method ASAP (e.g. by relying on synchronous data at startup (#703) to cache results where possible), I'd like to include a note along the lines of:

> When the initialization completion is signaled while the background script is still starting/executing, events are not queued and the usual event dispatch happens as soon as possible.

This note does not have hard guarantees but expresses the intended semantics.

What are your thoughts on that?

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Implementation Notes
+
+* The `async_initialization` manifest key allows the browser to know before spinning up the service worker that it should start queuing events *in the renderer*.
+* The browser's event router must compute the union of the old persisted listener configurations and any newly evaluated configurations during startup to determine which events to forward to the initializing renderer.
+* Implementations must ensure that enqueued tasks are flushed and dispatched normally upon calling `markInitializationComplete()`. If the initialization fails before the method is called, the state commit is aborted, and the previous run's listener state is preserved in the browser process.
+
+## Future Work
+
+N/A
```
Rob--W: Potential future work: dispatch `runtime.onPerformanceWarning` event when events have been dropped. This would enable extensions to detect bugs. This event was already described in #456 and implemented in Firefox, documented at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onPerformanceWarning

Rob--W (commented):
This proposal looks useful to extensions, and feasibly implementable by browsers.

Right now it is focused on adjusting event registration behavior at initialization.
Would you be willing to consider expanding to the post-initialization phase? With a method to control listener registration behavior, we can reduce the number of unnecessary wake-ups (#719) and writes to disk (frequency and size). E.g.:

```javascript
browser.runtime.markBackgroundEventStage({
  // Commit listeners and flush (as proposed)
  listenersReady: true, // can only be true once

  // Whether to persist across browser/extension restarts
  persistAcrossSessions: true, // default true

  // Whether to persist past worker termination.
  persistAcrossUnload: true, // default true
});

try {
  browser.runtime.markBackgroundEventStage({
    persistAcrossUnload: false,
  });
  browser.webNavigation.onBeforeNavigate.addListener(...);
} finally {
  browser.runtime.markBackgroundEventStage({
    persistAcrossUnload: false,
  });
}
```

This is just an example, an enum could also apply here. The method would update a renderer-side flag specific to the extension worker, and whenever addListener is called, the flag would be passed along to help the browser with making the decision on whether to persist, and to what extent.

proposals/async-listener-registration.md
AndreaOrru: Yes, your interpretation is accurate. I'll reword it to clarify that this only delays dispatch of extension API events to the background context during startup.

I’ll use "background context" to cover extension service workers and event pages.

proposals/async-listener-registration.md
AndreaOrru: Agreed that calling the method during the initial run should ideally not add extra queueing latency compared with the usual registration path.

I'll add a note along those lines -- but it seems like it belongs in "Implementation Notes"? I put it there, but let me know if you feel strongly about it.

proposals/async-listener-registration.md
AndreaOrru: What do you mean by "events have been dropped"? Because of queue overflow / initialization timeout?

Completing initialization without re-registering a listener can lead to "events being dropped" in some sense, but that would be working as intended.

AndreaOrru:
> [...] Would you be willing to consider expanding to the post-initialization phase? With a method to control listener registration behavior, we can reduce the number of unnecessary wake-ups (#719) and writes to disk (frequency and size). E.g.:
> 
> ```js
> browser.runtime.markBackgroundEventStage({
>   // Commit listeners and flush (as proposed)
>   listenersReady: true, // can only be true once
> 
>   // Whether to persist across browser/extension restarts
>   persistAcrossSessions: true, // default true
> 
>   // Whether to persist past worker termination.
>   persistAcrossUnload: true, // default true
> });
> 
> ...
> ```

I agree this is a real problem, and this API shouldn't preclude a future solution for "non-persistent" listeners. But I prefer not to fold that into this proposal. 

I'd rather add this as "Future Work", and I think a per-listener API would be better, e.g. an explicit `persistent` listener option for `addListener`, which is similar to what you've already proposed here: https://github.com/w3c/webextensions/issues/719#issuecomment-2761477404

AndreaOrru:
Additionally, I realized `markInitializationComplete` is not very explicit ("initialization" can mean different things), and it would be better to call it `markListenerRegistrationComplete`.

AndreaOrru:
Sorry for the delayed response -- please take another look.

carlosjeurissen:
@AndreaOrru checked on the potential backwards compatibility issue. Tested the recent versions of Chrome, Firefox, Safari and Orion on desktop and all of them load fine with additional properties in the background manifest key. See https://jeurissen.co/webext-demos/background-unknown-key

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Edge Cases
+
+* **Initialization failure fallback:** If the initialization phase fails (e.g., the worker crashes or hits a timeout before the completion API is called), the browser will abort the state commit. It will keep the previous persisted routing state to prevent leaving the extension in a broken state.
+* **Late registration:** It is permissible to register new listeners after the `markListenerRegistrationComplete()` API call. If an extension does so, developers must understand the following lifecycle:
+	* The browser will persist these "late" listeners across background context terminations to successfully trigger a future wake-up.
+	* On the subsequent wake-up, all relevant listeners must be re-registered. Calling `markListenerRegistrationComplete()` entirely replaces the routing state from the previous run.
+	* Therefore, any late-registered listeners from a previous run will be removed from the registered listeners during the next run's state commit unless the extension explicitly re-registers them before calling the completion API.
```
dotproto: IMO "late registration" is a notable aspect of this problem space. While this isn't the main way we expect extensions to (intentionally) register event listeners, I think it's more than an edge case. I'd suggest moving this content into a "Runtime Lifecycle" (or similarly named) section directly under "Behavior" and describing the related systems in more detail. For example, the late listener registration list should also be mentioned in the Initialization Lifecycle section.

I'm also concerned about scenarios where session event listeners are misinterpreted as being late registrations for init listeners. While I don't have a concrete use case in mind, I'm imagining scenarios where an extension registers temporary event listeners on events like `webNavigation.onCompleted` or `tabs.onUpdated` to get updates on events related to a specific task being performed. If these registrations occur in the course of an extension's typical execution, could the behavior described here result in a significant increase in the number of unnecessary background wakes? It seems like it may, even in situations where developers attempt to limit the scope of the listener using filters. 

Admittedly some extensions may use `unregister()` to clean up these temporary event listeners, but as I see it the history of async extension registration is working against us here. Since post-init event listeners haven't impacted an extension outside of the current background context's lifetime, I expect that many extensions won't have listener clean up logic because the platform "just took care of it" for them.

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,241 @@
...
+* **Event routing (union of old and new):** While in the initialization phase, the browser will route an event to the renderer queue if it matches *either* the old persisted routing state (from the previous run) or any new listeners currently being registered by the executing script.
+* **Listener registration:** Developers perform their potentially asynchronous setup (e.g., reading from storage) and call `addListener()` to attach their desired JavaScript callbacks *before* calling the completion API.
+* **State commit & completion API:** The extension must call `browser.runtime.markListenerRegistrationComplete()` once its potentially async setup is complete. At that moment:
+	1. The old persisted routing state is **thrown away**.
+	2. The new listeners currently registered are committed to the browser as the new persisted routing state for future wake-ups.
+* **Event flush:** Once `markListenerRegistrationComplete()` ends the initialization phase, the renderer flushes its queue, dispatching all held events to the newly registered JavaScript callbacks in FIFO order.
+
+#### Example Flow
```
dotproto: Building on my comment about [late registration](https://github.com/w3c/webextensions/pull/989/changes#r3211706372), this section should be updated to include mention of what how the extension and browser handle the value of `shouldListen` changing in storage. This mentioned in the paragraph immediately after the code sample, but so quickly that I didn't absorb the impact of this passage. 

I'd suggest adding another extension code sample showing the extension updating the value of `shouldListen` and registering an event listener, then describing how this feeds back into the browser's extension initialization process.

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,241 @@
...
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:** @rdcronin
+
+**Created:** 2026-04-27
+
+**Related Issues:** N/A
+
+## Motivation
```
dotproto, line 19: I think it may be worth introducing a "non-goals" or "out of scope" section to provide more information about what this proposal intentionally does not cover. As @Rob--W mentioned [in this comment](https://github.com/w3c/webextensions/pull/989#discussion_r3170648798), "standard SW lifecycle events, web platform APIs events, [and] non-event" are out of scope. It's not directly called out, but non-background contexts are also out of scope. 

A concrete use case for the latter is an extension popups fetching data from storage before calling `downloads.open()`. The user activation is lost because of the async storage operation, which will cause the `downloads.open()` call fail even though the call would succeed if it was made synchronously.

proposals/async-listener-registration.md
```diff
@@ -0,0 +1,241 @@
...
+   *
+   * This method is only available to the background script context.
+   */  
+  export function markListenerRegistrationComplete(): void;
+}
+```
+
+### Behavior
+
+#### Initialization Lifecycle
```
dotproto: We should call out in this section that handling of and limits on the event queues up to individual user agents.

While the limits themselves may be undefined, we may want to define some reporting mechanisms for these cases similar to what @Rob--W suggested [in this comment](https://github.com/w3c/webextensions/pull/989/changes#r3171127476).

tophf:
This solutions has an inherent problem of delaying the first dispatch, it's conceptually like the `if (initStorage) await initStorage;` in each listener that we currently use, so hopefully its implementation won't become a reason to delay implementation of an arguably cleaner solution of providing a synchronous small config storage available in the service worker, which was discussed in WECG some time ago.

proposals/async-listener-registration.md
Rob--W: > What do you mean by "events have been dropped"? Because of queue overflow / initialization timeout?

Yes to both. If the extension is somehow in a deadlock (e.g. initialization depends on an event, but events are all suppressed), then knowing that (critical) events have been dropped would be useful.

If the browser is permitted to flush the queue and force the initialization when a timeout is reached, notifying extensions would enable them to detect bugs. Perhaps this should be one of the events that are not delayed once registered.

In theory a full queue could eventually trigger a crash and theoretically trigger crash reports that could integrate in the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API). Only Chrome  recognizes the "crash-reporting" [endpoint](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints) though.

> Completing initialization without re-registering a listener can lead to "events being dropped" in some sense, but that would be working as intended.

Indeed.

Rob--W:
> This solutions has an inherent problem of delaying the first dispatch, it's conceptually like the `if (initStorage) await initStorage;` in each listener that we currently use, so hopefully its implementation won't become a reason to delay implementation of an arguably cleaner solution of providing a synchronous small config storage available in the service worker, which was discussed in WECG some time ago.

These features are developed independently, the PR for the synchronous storage is at https://github.com/w3c/webextensions/pull/793 and actively going through review and revision cycles right now.

It is a subject of discussion whether synchronous storage should persist across browser restarts. If that is not the case, an implementation could try to synchronously access data where available, and otherwise fall back to the delayed registration proposed here.

Rob--W:
@AndreaOrru The current proposal postpones all events, by design. While this penalty hopefully encourages extension developers to minimize the delay, the reality may be different.

I'm sketching a few scenarios below where the specified behavior may be too strict. They could be mitigated but that may result in additional complexity. I'd like to see your thoughts on these. To be clear, this not intended as a long lasting blocker, but to check if the API is in the desired shape before we proceed to the implementation.

In [a comment above](https://github.com/w3c/webextensions/pull/989/files#r3171127476) I mentioned `runtime.onPerformanceWarning` as an example of an event that could enable extension developers to discover issues. If events were postponed indefinitely, extensions could not be notified of issues and handle them accordingly.

Time-sensitive events: Network requests may be blocked on webRequest event handlers (unconditionally in Firefox, in Chrome conditional for policy-installed extensions except `webRequest.onAuthRequired`). A potential mitigation is to dispatch events as soon as a listener is registered, instead of queueing them all up until the end. There are pros and cons to this.

Documenting risks of API usage: When events are delayed, [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues become more likely. This has always been an issue with asynchronous APIs and is not new, but I think that it is worth noting that the use of this API increases the risk of that. There may be some API-specific solutions in some cases (e.g. `documentId`).

Some events may have different requirements for handling them. E.g. maybe `runtime.onMessage` and `webRequest` listeners should be processed ASAP, but `webNavigation` events might be less urgent.

---

I see the appeal in the proposal here from the ease of implementation and their obvious semantics. Perhaps it is sufficient - I am not opposed to the approach. Alternative API design knobs I can think of:

- decoupling async registration from blocking event dispatch: fire at the next microtask (or later) after the first registration. Potential issues: This can result in out-of-order event dispatches if an extension registers events at different times. This can also result in missed events if an extension registers two listeners for the same event at different times.
- letting extensions declare (in the manifest? At runtime, per synchronous API call at the top level?) whether they would like to postpone the registration of specific events (independently of event filters).

proposals/async-listener-registration.md
AndreaOrru: I agree that "late registration" shouldn't be described only as an edge case. I would move this into its own expanded section as you recommend, but let's discuss the semantics first. :)

The concern about temporary/session-scoped listeners is valid, although it's not clear to me whether you're advocating for changing the semantics or clarifying the documentation. At a minimum, we should make it very clear that a listener registered after `markListenerRegistrationComplete()` can affect future wake-up routing, and that listeners intended to be session-scoped must be removed explicitly.

That said, it's worth noting that under the current semantics, the impact of an accidentally persisted late listener is bounded by the next successful initialization commit: if the listener is not re-registered before the next `markListenerRegistrationComplete()` call, it is removed from the persisted routing state. This can still cause an unnecessary wake-up, so it's worth documenting prominently, but it doesn't persist indefinitely unless the extension keeps re-registering it.

The alternative would be to make late registrations ephemeral/session-scoped by default. The main downside is that intentional runtime configuration changes become harder. If an extension enables a feature after initialization and registers a new listener, that listener would not wake up the extension after the background context is terminated. The extension would then need some other way to commit updated routing state.

An explicit `persistent` event listener option, like the one I mentioned in https://github.com/w3c/webextensions/pull/989#issuecomment-4392317207, seems desirable because it would give developers direct control over whether a listener should survive background termination. Still, I'd prefer framing that as future work rather than expanding the scope of this proposal.

For this proposal, I think the better default is to keep late registrations persistent, document the lifecycle and risks clearly, and call out explicit session-scoped/non-persistent listeners as future work. But I am open to the alternative if others feel strongly about it.

AndreaOrru:
Thanks @Rob--W, this is a good articulation of the main trade-offs in the proposal. Allow me to group a few paragraphs and answer inline. =)

> I see the appeal in the proposal here from the ease of implementation and their obvious semantics [...]

Yes, the current design postpones extension API events until `markListenerRegistrationComplete()` because that gives the extension a single, atomic initialization boundary. Async state can be loaded, all intended listeners can be registered, routing state can be committed, and only then are queued events flushed.

> Time-sensitive events: Network requests may be blocked on webRequest event handlers (unconditionally in Firefox, in Chrome conditional for policy-installed extensions except `webRequest.onAuthRequired`). A potential mitigation is to dispatch events as soon as a listener is registered, instead of queueing them all up until the end. There are pros and cons to this.
> [...]
> decoupling async registration from blocking event dispatch: fire at the next microtask (or later) after the first registration. Potential issues: This can result in out-of-order event dispatches if an extension registers events at different times. This can also result in missed events if an extension registers two listeners for the same event at different times.

I don't think we should switch the default behavior to "dispatch as soon as a listener for this event is registered". Although it improves latency in some cases, it introduces order-dependent behavior during initialization, as you correctly point out. That undermines the atomicity that this proposal is trying to provide, which seems like a greater loss to me.

> Some events may have different requirements for handling them. E.g. maybe runtime.onMessage and webRequest listeners should be processed ASAP, but webNavigation events might be less urgent.
> [...]
> letting extensions declare (in the manifest? At runtime, per synchronous API call at the top level?) whether they would like to postpone the registration of specific events (independently of event filters).

Agreed that not all events have the same requirements. More granular controls, such as per-event manifest declarations or an API to opt individual events in/out of postponed dispatch, may be useful. My inclination is to treat those as future work unless the group feels the default semantics are too broad for the initial implementation.

> In [a comment above](https://github.com/w3c/webextensions/pull/989/files#r3171127476) I mentioned runtime.onPerformanceWarning as an example of an event that could enable extension developers to discover issues. If events were postponed indefinitely, extensions could not be notified of issues and handle them accordingly.

This seems qualitatively different from dispatching e.g. `webRequest` early. Those events are part of the extension's normal application logic, and dispatching them during partial initialization can create order-dependent behavior as explained above. `runtime.onPerformanceWarning`, by contrast, is a diagnostic signal, IIUC. Delivering it early does not compromise the atomicity of normal listener initialization in the same way. So it seems worth treating as an exception.

> Documenting risks of API usage: When events are delayed, [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues become more likely. This has always been an issue with asynchronous APIs and is not new, but I think that it is worth noting that the use of this API increases the risk of that. There may be some API-specific solutions in some cases (e.g. documentId).

Acknowledged, but I don't see a way around it; it's the nature of async APIs, and this proposal can increase the frequency of such scenarios. The documentation should warn that delayed event delivery can increase stale-state / TOCTOU risks. Extensions should revalidate when handling delayed events and prefer stable identifiers where available.

<hr/>

@dotproto, once we reach consensus on the semantics of "late registration", I will update the proposal with your suggestions. @Rob--W, let me know if you're comfortable with the trade-offs, and I'll update the proposal to call out some of these points explicitly as well.

xeenon (approved):

AndreaOrru:
Addressed several of @dotproto and @Rob--W comments, also renamed the "Initialization Lifecycle" section to "Listener Registration Lifecycle" and rewritten for extra clarity.

Rob--W (approved):
LGTM!

Issue #981, opened 2026-04-14
Persist `runtime.setUninstallURL()` until update


carlosjeurissen:
### Intro
Currently the persistence of `runtime.setUninstallURL()` differs per browser:
- Firefox does not persist the `uninstallURL` and requires it on every run of the background script
- Chrome/Edge persist the `uninstallURL` even after update

This leads to the following issues:
- Firefox does not open an `uninstallURL` when an extension is disabled beforehand (see https://github.com/w3c/webextensions/issues/970)
- Chrome/Edge in it's current shape can not undo setting an `uninstallURL`. If a developer wants to stop opening an `uninstallURL`, pushing an update without an `uninstallURL` will not impact existing installations where `setUninstallURL()` is called.

Orion seems to at least persist between browser sessions. However did not test if it is persisted on update.

### Solution
All browsers persist the `uninstallURL` UNTIL and excluding in the case of an extension update, resolving both issues.

The only workaround right now in Chrome would be to use `runtime.setUninstallURL('')` indefinitely if historically it used an `uninstallURL` yet nolonger wants this in newer versions.

Rob--W:
@oliverdunk will check on the acceptable behaviors in Chrome, and then we can consider aligning in Firefox. Likely considering to persist the uninstall URL, but not across extension updates.

bershanskiy:
> Alternatively Chrome could introduce a way to clear the `uninstallURL` with a new `runtime.clearUninstallURL()` or by supporting a currently undocumented Firefox and Orion approach of passing an empty string `setUninstallURL("")` to 'unset' the `uninstallURL`.

I believe Chrome supports `setUninstallURL("")` and calling it clears uninstall URL. It is even [documented in the official Chrome Docs](https://developer.chrome.com/docs/extensions/reference/api/runtime#method-setUninstallURL):

> Set an empty string to not open a new tab upon uninstallation.

Or did I mis-read something in the original comment?

carlosjeurissen:
@bershanskiy just tried the setUninstallURL("") in Chrome. Seems this works totally fine. Good catch. Somehow in London we all assumed Chrome currently does not allow this pattern at this moment given it does not support similar empty strings in other APIs like the `action.setPopup()` method. Updated the issue description to reflect this.

bershanskiy:
> If a developer wants to stop opening an `uninstallURL`, pushing an update without an `uninstallURL` will not impact existing installations where `setUninstallURL()` is called.

[...]

PR #990, opened 2026-05-01
Add setUninstallURL to persistence-of-states table

carlosjeurissen:
As discussed during the London F2F.

See also active discussion on setUninstallURL persistence:
https://github.com/w3c/webextensions/issues/981

Sidenote: we may want to consider a more semantic replacement for the A + B + C notation.


oliverdunk (approved):
Thanks Carlos!

PR #676, opened 2024-08-19
Proposal: Public Suffix API

mckenfra:
This formalizes #231 into a concrete proposal.

Issue #231, opened 2022-06-17
Extension API to find the public suffix (eTLD) of a given URL/domain


Rob--W:
The [public suffix list](https://publicsuffix.org/) is a database of effective top-level domains (eTLD), which are the public suffix of URLs. This database is included in browsers (at least by Firefox, Chrome and Safari - sources below) and may be updated remotely. There have been feature requests for an API that allows extensions to identify the public suffix (eTLD) of a given URL:

- Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1315558
- Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1163295
- (rejected) request for a web platform API: https://github.com/whatwg/url/issues/528

There are solve known problems with the application of public suffix list (https://github.com/sleevi/psl-problems), but that does not necessarily rule out an extension API with such access. Extensions that need to rely on the public suffix list currently need to rely on alternatives, such as bundling the database with the extension, at the risk of having incompatible interpretations of the "public suffix of a URL" between the browser and the extension. With proper documentation of the problems associated with the public suffix list, extension authors can make a conscious decision to use the API when they need to.

This issue is to track use cases and the desired shape of the API. For example, the following would be the minimum:

```js
let suffix = await browser.publicsuffix.getPublicSuffix("www.example.co.uk");
// Result: co.uk
```

Here are other examples of APIs to query the public suffix:
- https://searchfox.org/mozilla-central/rev/49566d906ad040cf233b0a5acd597e1a63b92f72/netwerk/dns/nsIEffectiveTLDService.idl
- https://source.chromium.org/chromium/chromium/src/+/main:net/base/registry_controlled_domains/registry_controlled_domain.h;drc=4f516be19f2c1d35dc3240d050d84d10f0d6f726
- https://github.com/WebKit/WebKit/blob/0a51cee7874970c895c648c0b664b511ecf72ed5/Source/WebCore/platform/PublicSuffix.h

oliverdunk:
Similar proposal: https://github.com/w3c/webextensions/issues/58

xeenon:
I would be in favor of this for Safari.

gijsk:
In favour of this proposal, in addition to the consistency issue that was pointed out in the meeting (ie if the extension and browser have a different version of the PSL, there's a potential for security issues), some other arguments for having this exposed as an API:

[...]

PR #1008, opened 2026-05-25
Drop publicSuffix.getVersion()

Rob--W:
This is an update to the `publicSuffix` proposal to drop the specified `publicSuffix.getVersion()` method.

It was originally added in #676 following a request from @oliverdunk , referencing a PSL maintainer's request (quoted again at https://phabricator.services.mozilla.com/D295513#10264666 ):

> "Would you be open to adding a method to retrieve the PSL version in this proposal (I know it's mentioned in future work)? My understanding is that this currently doesn't exist, but the PSL maintainers have had requests for that in the past and were open to adding something. I'll let them know about this PR and perhaps they will have additional thoughts."

Another early position at https://github.com/w3c/webextensions/issues/231#issuecomment-2127529231 :

> Related, we should expose a list version to developers. There isn't currently an official version but Simone took an action item to investigate adding this as it has come up in the past.

Mozilla's PSL maintainer's recent position is as follows: https://phabricator.services.mozilla.com/D295513#10434003

> I am not a fan of providing a version. Mostly because it is not clear to me why this would be useful. Since fingerprinting is not a big issue for extensions I am not strongly opposed. Only from a "why build the API if it doesn't make sense" point.

@kzar is looking into a Chromium implementation, and mentioned the following:

> FWIW I was looking at the Chromium implementation recently, and so far my draft stubbed out `getVersion()` to always return null, as the version didn't seem to be available so far. Anything's possible, but if there's also difficulty adding this from the Firefox side as well, and as it's IMO not something most users of the API will need, dropping `getVersion()` for now seems pragmatic to me.

### DIFF
--- a/proposals/public-suffix.md
+++ b/proposals/public-suffix.md
@@ -400,9 +400,6 @@ namespace publicSuffix {
   )
   : string | null;
 
-  // Gets the value of the VERSION metadata field in the PSL dataset if available
-  export function getVersion(): string | null;
-
 
   // INTERFACES
 
@@ -761,22 +758,11 @@ It is noted that:
 In the future, if PSL lookups are resolved by way of DNS requests rather than by querying a
 browser-bundled PSL dataset (as is the case currently), then an async API may be a better fit.
 
-#### 11. PSL Version
-
-Versioning metadata was introduced into the PSL with [this commit](https://github.com/publicsuffix/list/issues/1808#issuecomment-2455793503).
-
-Method `getVersion()` of this API should return the value of the
-VERSION metadata field contained in the specific PSL dataset used by the browser.
-
 ### New Permissions
 
 | Permission Added | Suggested Warning |
 | ---------------- | ----------------- |
 | publicSuffix     | N/A               |
-
-Method `publicSuffix.getVersion()` may theoretically contribute to making the browser
-fingerprintable. However, it is already possible for malicious sites to get the browser's
-version, therefore this API does not introduce significant additional fingerprintability.
 
 ### Manifest File Changes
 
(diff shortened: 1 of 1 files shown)


Rob--W:
In the WECG meeting a few days ago I asked @oliverdunk whether `publicSuffix.getVersion()` is still desired (last topic, meeting notes pending review in #1002). I'll therefore let @oliverdunk review this change.

Rob--W:
Since I see a few thumbs up on this PR, I'd like to emphasize that I'm still open to considering the possibility of keeping `publicSuffix.getVersion()`, especially if Chromium commits to supporting it. The PR here just serves to show the proposal delta between keeping and dropping it, with two buttons acting on either outcome - closing PR or merging it.

On the Firefox side we already have a patch pending review that adds `getVersion()`, at https://phabricator.services.mozilla.com/D295513. The implementation is feasible, it is a question of whether the approach has cross-browser approval.

oliverdunk:
Thanks for opening this @Rob--W! I've reached out to the PSL folks on our side and will try to follow-up soon.

oliverdunk:
After some digging it seems like this is something @weppos wanted (or at the very least, something that came up in discussion with them). I've reached out to ask for their thoughts - alternatively, feel free to reply here Simone.

oliverdunk (approved):
I don't have strong feelings on this - I'm happy to land if we don't get feedback to the contrary.

weppos:
> After some digging it seems like this is something @weppos wanted (or at the very least, something that came up in discussion with them). I've reached out to ask for their thoughts - alternatively, feel free to reply here Simone.

I was proposing the "VERSION" as a way to surface to developers what snapshot they are building upon. Not critical to keep. If you have alternative ways to determine or expose what is the specific copy of the PSL data, that would also work.

PR #1012, opened 2026-05-26
Drop publicSuffix.DomainOptions.encoding

kzar:
Since safely displaying Unicode domains to the user seems out of scope for the
publicSuffix API, let's always encode returned registrable domains using
Punycode.

See the discussion here https://github.com/w3c/webextensions/issues/231#issuecomment-4542627098

### DIFF
--- a/proposals/public-suffix.md
+++ b/proposals/public-suffix.md
@@ -216,27 +216,24 @@ The PSL uses Unicode, and contains International Domain Name (IDN) eTLDs. Depend
 on the use case, when returning the registrable domain for a domain name,
 either Unicode or Punycode may be preferred.
 
-According to a possible [PSL algorithm](https://github.com/publicsuffix/list/wiki/Format#formal-algorithm)
+Points to consider regarding the encoding used for returned domains are:
+
+* According to a possible [PSL algorithm](https://github.com/publicsuffix/list/wiki/Format#formal-algorithm)
 for interpreting the PSL dataset, candidate domains should be converted to Punycode
-before matching against the PSL. Therefore, a requirement to convert Punycode
-back to Unicode would involve extra work, and it may be preferable to avoid this in
-performance-sensitive use cases. However, it has been suggested that this PSL algorithm
-is not authoritative. Indeed, any algorithm used to interpret the PSL dataset is likely
-to be fairly trivial, since it essentially involves comparing labels for equality.
-
-Further points to consider in this context are:
-
+before matching against the PSL. Therefore, returning Punycode avoids requiring the
+browser to convert matching results back to Unicode.
 * Punycode may be the most appropriate encoding to use, since a Punycode hostname is a
-valid URL whereas a Unicode hostname is not. Therefore Punycode may be the most sane default.
-
-* When this API is used to obtain registrable domains intended for display to the user,
-it is likely that the end result will at some point need to be converted to Unicode,
-since users may be less familiar with Punycode.
+valid URL whereas a Unicode hostname is not.
+* When intended for display to the user, converting a Punycode domain to Unicode is not always safe.
+For example, sometimes Unicode domains are used to spoof popular domains, in an attempt to trick users.
+Therefore, browsers have their own internal heuristics for deciding how URLs are displayed, sometimes
+choosing to stick to Punycode where it is safer - those heuristics should likely be exposed as a more
+general web or extensions API, rather than as a part of this one.
 
 ##### 4.1 Recommendation
 
-It is recommended that registrable domains should be returned as Punycode by default,
-but the API should also provide an option to convert these to Unicode.
+It is recommended that registrable domains and eTLDs should always be returned
+using Punycode.
 
 #### Use Cases
 
@@ -409,10 +406,6 @@ namespace publicSuffix {
   // Options that may be passed to the API method to control its behaviour.
   interface DomainOptions {
 
-    // Determines how the returned domain should be encoded.
-    // Default = punycode
-    encoding?: DomainEncoding,
-
     // If true, and the input hostname is an IP address, then this is returned as-is.
     // Default = false
     allowIPAddress?: boolean,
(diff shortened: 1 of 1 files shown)


mckenfra:
The Public Suffix API proposal was reviewed and approved by W3C reviewers from Mozilla, Apple and Google. During the formulation of the proposal, reviewers discussed and were supportive of providing an option to return registrable domains encoded as unicode, e.g.:

- [2024-09-12 Timothy (Apple)](https://github.com/w3c/webextensions/blob/fdda10840f7ff9429c67d49fe5c55c394cddf69d/_minutes/2024-09-12-wecg.md?plain=1#L121) In favour of a unicode opt-in to facilitate displaying registrable domains to the user.
- 2025-05-17 rdcronin (Google) https://github.com/w3c/webextensions/pull/676#pullrequestreview-2847544356 (click to expand 2nd review comment) As well as optionally encoding returned registrable domains as unicode, suggested the possibility of providing further encoding options in future.

Mozilla plans to make use of this unicode-encoding option in an update to its [Multi-Account-Containers extension](https://github.com/mozilla/multi-account-containers) extension: on the screen showing the domains in the user's containers, the domains will be shown grouped by their registrable domains, which will need to be encoded as unicode since users may be unfamiliar with punycode.

Given the reviewers' support and above example use case for returning registrable domains optionally encoded as unicode, it may not be appropriate to drop this option entirely. However, a solution might be to rename the `unicode` member of the `DomainEncoding` enum. Possible alternative names:

- `display`
- `safe-unicode`

kzar:
Thanks @mckenfra , given there's a concrete use case I'm more inclined to keep it like you say, and I agree that if we do keep it, a rename might make sense.

However, I still wonder if the "safe Unicode" encoding of domains would make more sense as a separate extensions (or web) API. I could imagine folks misusing `publicSuffix.getDomain` just as a way to format a domain for display, without needing to look it up in the PSL - wouldn't it make more sense to expose that functionality for them separately?

mckenfra:
@kzar It definitely seems like it could be useful to expose a generalised API to convert from punycode to other encodings. If this were to be a webextensions API, then a first step would be to raise an issue containing a high-level proposal of the API: https://github.com/w3c/webextensions/issues

Regarding the publicSuffix API, I am happy to raise a PR to change the name of the option value from `unicode` to `display`.

kzar:
_(Marking as a draft for now, since @mckenfra has [proposed an alternative approach](https://github.com/w3c/webextensions/pull/1013).)_

Rob--W:
> @kzar It definitely seems like it could be useful to expose a generalised API to convert from punycode to other encodings. If this were to be a webextensions API,

The conversion from/to punycode can be implemented entirely in JS libraries.

What is unique to PSL (and "display" encoding) is that browsers may have their own view, which makes it infeasible for a JS library to have a consistent and up-to-date view. That's why we have the `publicSuffix` API, and why I'm advocating for continuing support for partial unicode encoding (under the "display" name to avoid issues with confusables).

kzar:
_Fair enough, seems like I'm out voted here 😅. Will close this out._

kzar:
@Rob--W I realised I should write down my thinking more clearly here for posterity.

I figured that while you can convert a URL to punycode with existing APIs (e.g. `new URL("https://foo.bar.example.مليسيا").href`), I don't think there's a web API way to do the reverse yet. The closest I could find to that was Node.js' [url.domainToUnicode(domain)](https://nodejs.org/api/url.html#urldomaintounicodedomain).

I figured that adding such an API, to allow something like `new URL("https://foo.bar.example.xn--mgbx4cd0ab/").toDisplay()` would be generally useful. In particular, since as we discussed it's not only a question of converting punycode domains to Unicode, but also of hooking into the browser's logic to perform spoof checks (e.g. [IDNToUnicode](https://source.chromium.org/chromium/chromium/src/+/main:components/url_formatter/url_formatter.h;l=185?q=UnsafeIDNToUnicodeWithDetail&ss=chromium%2Fchromium%2Fsrc) in Chromium, and [domainToDisplay](https://searchfox.org/firefox-main/rev/9d2be0cb8dbf1c3a7ef6ba79be35d637d0327eaa/netwerk/dns/nsIIDNService.idl#55) in Firefox).

I wondered if adding the `{ encoding: "safe-unicode" }` (or similar) argument to `publicSuffix.getDomain()` instead could result in folks misusing that API as a way to display URLs, e.g. something like:

```javascript
URL.prototype.toDisplay = function toDisplay() {
  return this.href.replace(
    this.hostname,
    this.hostname.replace(/xn--[a-z0-9-]+/gi, label =>
      chrome.publicSuffix.getDomain(label + ".com", {
        encoding: "safe-unicode",
      }).slice(0, -4)
    )
  );
};
```

I appreciate adding a web API like `URL.prototype.toDisplay()` is likely a heavy lift however.

Rob--W:
> I figured that while you can convert a URL to punycode with existing APIs (e.g. `new URL("https://foo.bar.example.مليسيا").href`), I don't think there's a web API way to do the reverse yet. The closest I could find to that was Node.js' [url.domainToUnicode(domain)](https://nodejs.org/api/url.html#urldomaintounicodedomain).

> I wondered if adding the `{ encoding: "safe-unicode" }` (or similar) argument to `publicSuffix.getDomain()` instead could result in folks misusing that API as a way to display URLs,

That thought crossed my mind before; I view that as creative bending of existing APIs to fulfill functionality that is not available otherwise. As long as it does not cause undesired side effects (security, privacy, etc), I'm fine with it.

> I appreciate adding a web API like `URL.prototype.toDisplay()` is likely a heavy lift however.

FYI there is relevant (yet stalled) discussion for that kind of functionality at https://github.com/whatwg/url/issues/274 . This discussion references `URL.domainToUnicode` which was removed from the spec, but which Node.js had implemented (as you noticed). In that thread, I see the following notable remark, which also applies to the proposed publicSuffix API (resulting in this PR and #1013):

> Well, which "version" of ToUnicode do we want, the standard ICU implementation or what the browser URL bar displays?

PR #1013, opened 2026-05-27
Rename publicSuffix option `encoding = unicode`

mckenfra:
This is an update to the publicSuffix proposal to:

- Rename `getDomain()` option `{ encoding: "unicode" }` to `{ encoding: "display" }`.
- Note that unicode-encoded domains may be suspectible to [Unicode Confusables](https://www.unicode.org/reports/tr39/#Confusable_Detection) risk.
- Clarify that conversion to unicode should be done only if safe to do so for a given domain.

See https://github.com/w3c/webextensions/issues/231#issuecomment-4542627098 and PR https://github.com/w3c/webextensions/pull/1012

### DIFF
--- a/proposals/public-suffix.md
+++ b/proposals/public-suffix.md
@@ -233,10 +233,17 @@ valid URL whereas a Unicode hostname is not. Therefore Punycode may be the most 
 it is likely that the end result will at some point need to be converted to Unicode,
 since users may be less familiar with Punycode.
 
+* When converting domains to Unicode for display to the user, security considerations
+relating to ["Unicode confusables"](https://www.unicode.org/reports/tr39/#Confusable_Detection)
+should be taken into account. "Unicode confusables" are cases where two strings
+are visually confusable. It is understood that the major browsers already have
+functionality for safely converting Punycode domains to Unicode.
+
 ##### 4.1 Recommendation
 
 It is recommended that registrable domains should be returned as Punycode by default,
-but the API should also provide an option to convert these to Unicode.
+but the API should also provide an option to convert these to an encoding suitable
+for display to the user ("safe Unicode").
 
 #### Use Cases
 
@@ -432,7 +439,7 @@ namespace publicSuffix {
   // The available encoding types for the returned domain.
   enum DomainEncoding {
     punycode,
-    unicode,
+    display,
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


kzar:
FWIW I think this is a big improvement, and I'm not against going this way necessarily. I also think having a way to safely display a Unicode domain to the user is definitely a useful thing for the APIs to provide.

That said, I still wonder if it makes sense to include the display encoding option as a part of the `publicSuffix` API. To me, it would make more sense to split it out into its own API. I figure that displaying a domain to the user is quite a different thing to looking up a domain in the PSL. I figure that splitting that out would make the PSL API simpler, and it would make it easier for extensions to display domains in cases where a PSL lookup wasn't required.

In any case, it's not a hill I want to die on or anything like that, and I'm not too keen to slow things down if combining the display encoding option in the  `publicSuffix` API works for everyone.

Rob--W:
> That said, I still wonder if it makes sense to include the display encoding option as a part of the `publicSuffix` API. To me, it would make more sense to split it out into its own API.

The use case of "get the public suffix, for display purposes" sounds well within the scope of the `publicSuffix` API to me. I can imagine use cases (some of which are already listed in the publicSuffix proposal, and @mckenfra also mentioned one at https://github.com/w3c/webextensions/pull/1012#issuecomment-4546597783).

On the question of whether it should be a separate API -  "Decode punycode except when there are confusables" sounds very niche, and I'm not sure if it's worth having a separate dedicated API for that specific use case.

Rob--W (approved):
Looks good to me, thanks for putting this fix up. This matches with what I was thinking of in https://github.com/w3c/webextensions/issues/231#issuecomment-4543739132

I wonder whether we should try to emphasize more that the returned domain is intended to match what the browser displays in the URL bar. But even without that, the updated text looks good.

I have added @oliverdunk as a reviewer to this proposal update, to make sure that we have sufficient cross-browser coverage (and from more PSL maintainers; I already checked in with @simon-friedberger on Mozilla's side, but the more, the merrier!).

oliverdunk (approved):
I discussed this briefly with our PSL folk internally. One suggestion was to break this down into three options:

- Strict unicode
- Display mode (unicode or punycode based on browser logic)
- Strict punycode

This way you get full flexibility.

That said, there were no strong opinions so I'll approve this anyway and feel free to land with whatever we prefer.

As always, we haven't gone through internal security review for this API so we may have some follow up feedback once we finish that (I've just started the process). I don't expect significant changes though as this seems like a reasonable API surface.

proposals/public-suffix.md
```diff
@@ -672,9 +680,9 @@ classes of input `hostname` parameter:
 | مليسيا             | this is an IDN that is also an eTLD              | null                   |
 | xn--mgbx4cd0ab     | as above, but Punycode                           | null                   |
 | foo.مليسيا         | this is an IDN                                   | foo.xn--mgbx4cd0ab     |
-| foo.مليسيا         | as above, with `unicode = true`                  | foo.مليسيا             |
+| foo.مليسيا         | as above, with `encoding = display`              | foo.مليسيا             |
 | foo.xn--mgbx4cd0ab | this is an IDN, but Punycode                     | foo.xn--mgbx4cd0ab     |
-| foo.xn--mgbx4cd0ab | as above, with `unicode = true`                  | foo.مليسيا             |
+| foo.xn--mgbx4cd0ab | as above, with `encoding = display`              | foo.مليسيا             |
```
Rob--W, line 685: Let's include an example of a confusable for reference.

```suggestion
| foo.xn--mgbx4cd0ab | as above, with `encoding = display`              | foo.مليسيا             |
| xn--bs-red.com | `bսs.com` in Unicode, kept as punycode even with `encoding = display` due to presence of confusable | xn--bs-red.com | 
```

Rob--W:
Thanks for your quick input Oliver.

I added one suggested edit to include an example of a confusable for future reference.

Issue #967, opened 2026-03-26
Extension API to explicitly create split views or unsplit them


Rob--W:
Chrome and Firefox now have implementations of the split view feature:
- Firefox: https://support.mozilla.org/en-US/kb/split-view-firefox (announcement: https://blog.mozilla.org/en/firefox/split-view/)
- Chrome: https://support.google.com/chrome/answer/16971124

Early support for the `splitViewId` concept was added in #842 with https://github.com/w3c/webextensions/blob/main/proposals/split_tabs_proposal.md

Although it is possible to detect the presence of split views, it is not possible to explicitly create or remove split views. This issue is a placeholder to have discussions.

Relevant scenarios:
- Given a tab, create a split view (defaulting to whatever tab the browser shows by default).
  - Question: should there be a way to open the split view to the left or right? Or should extensions wishing to control that behavior make an additional API call, e.g. `tabs.move`?
- Given two tabs, put them in a new split view.
  - Edge case: what if either is already part of a split view? What if the tabs are not already adjacent to each other?
- Given two tabs in a split view, remove the split view (= transition `splitViewId` to `-1`).

Firefox ticket tracking the implementation of this: https://bugzilla.mozilla.org/show_bug.cgi?id=2016928

MentalGear:
Hi thanks for your work and proposal for this! 
Having a splitView add command is essential indeed.

> Question: should there be a way to open the split view to the left or right? Or should extensions wishing to control that behavior make an additional API call, e.g. tabs.move?

As an optional parameter, definitely good for DX convenience. E.g. splitViewAdd ( [...tabIds], side) but honestly I think as a first version the browser default side would be enough.

> Edge case: what if either is already part of a split view? What if the tabs are not already adjacent to each other?

Throw error for application code to handle. Error codes specific to "not_adjacent" and "already_split". Later on we can still provide convenience functions for this.

Rob--W:
FYI I put up an API proposal at https://github.com/w3c/webextensions/pull/1019, feedback is welcome.

Tai-ch0802:
**+1 — adding a real-world Chromium consumer for this proposal.**

[...]

PR #1019, opened 2026-06-04
Add proposal for modifying split views

Rob--W:
Proposal for:
- #967

### DIFF
--- a/proposals/split_view_modification.md
+++ b/proposals/split_view_modification.md
@@ -1,0 +1,213 @@
+# Proposal: Modifying split views
+
+**Summary**
+
+This proposal describes additions to the `tabs` namespace to support
+modification of split views.
+
+**Document Metadata**
+
+**Author:** @Rob--W
+
+**Sponsoring Browser:** Mozilla
+
+**Contributors:** N/A
+
+**Created:** 2026-06-04
+
+**Related Issues:** https://github.com/w3c/webextensions/issues/967
+
+## Motivation
+
+### Objective
+
+The Split View feature was introduced before in
+[split_tabs_proposal.md](https://github.com/w3c/webextensions/blob/main/proposals/split_tabs_proposal.md),
+and specified `splitViewId` in a read-only fashion in some `tabs` APIs.
+
+This proposal specifies the missing operations that extensions need to properly
+interface with split views:
+
+- Creating split views
+- Joining tabs in a split view
+- Separating tabs in a split view
+
+#### Use Cases
+
+Enabling extensions to offer the built-in functionality as documented at:
+
+- https://support.mozilla.org/en-US/kb/split-view-firefox
+- https://support.google.com/chrome/answer/16971124
+
+Specifically:
+
+- Creating a split view from an existing tab, which creates a new split view
+  consisting of the specified tab, plus a browser-native UI page where the user
+  can choose the tab to adopt in that split.
+
+- Creating a split view from two existing tabs, putting two splits together.
+
+- Separating split views.
+
+- Reversing tabs in a split view.
+
+### Known Consumers
+
+Tab managers, offering tab management functionality through a custom UI, e.g.:
+
+- [Tree Style Tabs](https://addons.mozilla.org/firefox/addon/tree-style-tab/),
+  tracked at https://github.com/piroor/treestyletab/issues/3806
+- [Sidebery](https://addons.mozilla.org/firefox/addon/sidebery/), tracked
+  at thttps://github.com/mbnuqw/sidebery/issues/2327
+
+Any extension that wants to open to a split view, e.g. requested at
+ https://issues.chromium.org/issues/510252106
+
+## Specification
+
+### Schema
+
+```javascript
+browser.tabs.create({
+  // ... existing properties, including: url, index, active
+  splitWithTabId: number,   // Existing tab to pair with
+}) : Promise<Tab>           // Returns new tab in split with given tab
+
+browser.tabs.createSplitView(
+  tabIds: Array<number>     // Two tabIds.
+) : Promise<number>         // Returns splitViewId
+
+browser.tabs.separateSplitView(splitViewId) : Promise<void>
+```
+
+
+### Behavior
+
+#### tabs.create() with splitWithTabId option
+
+First, validate the options and ensure that it would create a tab next to the
+tab specified by `splitWithTabId`. The method SHOULD reject if it cannot create
+a split view (see createSplitView). If a tab was created, the created tab must
+be returned as usual, even if the split view cannot be created.
+
+The returned `Tab.splitViewId` MUST reflect whether the tab successfully joined
+the split view.
+
+Tabs in a split view are neighbors of each other, and as such the new tab MUST
+be adjacent to the specified `splitWithTabId`. The `index` property affects the
+relative positioning:
+- `index` unspecified - default behavior, create tab at the right.
+- `index` set to index of `splitWithTabId` - create tab at the left.
+- `index` set to index of `splitWithTabId` plus one - create tab at the right.
+- Generally, if index is set to a position within the same split view, the new
+  tab is created at that position. (this clause covers a hypothetical future
+  where browsers support more than two tabs in a split view).
+- Any other `index` is rejected, because they fail to meet the requirement of
+  the tabs being neighbors.
+
+If `url` is not specified, the browser MAY open a custom split view-specific
+page instead of the default new tab page.
+
+
+#### tabs.createSplitView()
+
+Requires a pair of tabIds referring to distinct tabs, rejects otherwise.
+
+If the two tabs are already part of the same split view, the implementation MAY
+return the existing split view ID.
+
+If either tab is ineligible for joining the split, reject the call. Tabs can be
+ineligible for joining a split view for various reasons, including:
+- mismatching "pinned" states
+- mismatching "groupId" states
+- mismatching "splitViewId" states
+- mismatching "windowId" states
+
+At the bare minimum, two eligible tabs, with tabIds specified in order, should
+be able to join a split view, reflected in `Tab.splitViewId` and two
+`tabs.onUpdated` events.
+
+When tabs are not adjacent, or even in order, the implementation MAY try to
+move the tabs to the given order and join them.
+
+Upon returning, either the given tabs are in the same split view, or the method
+rejects without having changed the tab state.
+
+
+#### tabs.separateSplitView()
+
+Separates the tabs in a split view, "unsplitting" the split view.
+
+Rejects if `splitViewId` does not refer to a valid split view, or if the split
+view cannot be removed.
+
+Upon separation, the tabs' `splitViewId` changes to `SPLIT_VIEW_ID_NONE`, and
+returns successfully.
+
+
+#### tabs.move
+
+`tabs.move` does not receive new options, but is worth mentioning due to their
+potential impact on split views. The defining characteristic of a Split view is
+the adjacency and coupling of (two) tabs.
+
+The `tabs.move` method is specified to move one or more tabs, to end at the
+position given by `index`. This expectation is not always compatible with the
+request of moving a split view.
+
+If the `index` behavior of `tabs.move` weighs more strongly, then the move of
+tabs MAY separate tabs automatically.
+
+Browsers that treat split views as one unit MAY move multiple tabs, exposed via
+`tabs.onMoved`, `tabs.onDetached`, `tabs.onAttached` as needed. In that case,
+the `index` specifies the desired position of the split view after moving.
+
+
+### New Permissions
+
+No new permissions. This expands the existing `tabs` namespace, which itself
+does not have any permission requirements for tab management operations.
+
+### Manifest File Changes
+
+No manifest.json changes.
+
+## Security and Privacy
+
+### Exposed Sensitive Data
+
+Does not expose sensitive data.
+
+### Abuse Mitigations
+
+N/A
+
+### Additional Security Considerations
+
+N/A
+
+## Alternatives
+
+### Existing Workarounds
+
+`tabs.move` with a list of tabIds can be interpreted as a way to intentionally
+re-arrange tabs, and separating tabs as a consequence.
+
+There are no existing mechanisms for extensions to create new split views.
+
+### Open Web API
+
+Tab management is outside the scope of the web platform.
+
+## Implementation Notes
+
+If there are any significant notes on implementation that are relevant across
+browsers, please indicate them here.  (Feel free to add more sections.)
+
+## Future Work
+
+The `tabs.createSplitView` method could take an additional `options` object to
+enable additional behaviors, such as the relative size.
+
+The `tabs.createSplitView` method is forwards-compatible with the possibility
+of a split view having more than two tabs.
--- a/proposals/split_view_modification.md
+++ b/proposals/split_view_modification.md
@@ -58,7 +58,7 @@ Tab managers, offering tab management functionality through a custom UI, e.g.:
 - [Tree Style Tabs](https://addons.mozilla.org/firefox/addon/tree-style-tab/),
   tracked at https://github.com/piroor/treestyletab/issues/3806
 - [Sidebery](https://addons.mozilla.org/firefox/addon/sidebery/), tracked
-  at thttps://github.com/mbnuqw/sidebery/issues/2327
+  at https://github.com/mbnuqw/sidebery/issues/2327
 
 Any extension that wants to open to a split view, e.g. requested at
  https://issues.chromium.org/issues/510252106
--- a/proposals/split_view_modification.md
+++ b/proposals/split_view_modification.md
@@ -41,15 +41,15 @@ Enabling extensions to offer the built-in functionality as documented at:
 
 Specifically:
 
-- Creating a split view from an existing tab, which creates a new split view
-  consisting of the specified tab, plus a browser-native UI page where the user
-  can choose the tab to adopt in that split.
+- Creating a split view from a single existing tab, which creates a new split
+  view consisting of the specified tab, plus a browser-native UI page where the
+  user can choose the tab to adopt in that split.
 
 - Creating a split view from two existing tabs, putting two splits together.
 
 - Separating split views.
 
-- Reversing tabs in a split view.
+- Reversing the order of tabs in a split view.
 
 ### Known Consumers
 
@@ -73,11 +73,11 @@ browser.tabs.create({
   splitWithTabId: number,   // Existing tab to pair with
 }) : Promise<Tab>           // Returns new tab in split with given tab
 
-browser.tabs.createSplitView(
+browser.tabs.createSplit(
   tabIds: Array<number>     // Two tabIds.
 ) : Promise<number>         // Returns splitViewId
 
-browser.tabs.separateSplitView(splitViewId) : Promise<void>
+browser.tabs.unsplit(splitViewId) : Promise<void>
 ```
 
 
@@ -87,15 +87,14 @@ browser.tabs.separateSplitView(splitViewId) : Promise<void>
 
 First, validate the options and ensure that it would create a tab next to the
 tab specified by `splitWithTabId`. The method SHOULD reject if it cannot create
-a split view (see createSplitView). If a tab was created, the created tab must
-be returned as usual, even if the split view cannot be created.
+a split view (see createSplit). If a tab was created, the created tab MAY be
+returned as usual, even if the split view cannot be created.
 
-The returned `Tab.splitViewId` MUST reflect whether the tab successfully joined
-the split view.
+The returned `Tab.splitViewId` MUST reflect the split view that the tab joined.
 
 Tabs in a split view are neighbors of each other, and as such the new tab MUST
 be adjacent to the specified `splitWithTabId`. The `index` property affects the
-relative positioning:
+relative positioning (the following assumes a left-to-right browser UI):
 - `index` unspecified - default behavior, create tab at the right.
 - `index` set to index of `splitWithTabId` - create tab at the left.
 - `index` set to index of `splitWithTabId` plus one - create tab at the right.
@@ -109,7 +108,7 @@ If `url` is not specified, the browser MAY open a custom split view-specific
 page instead of the default new tab page.
 
 
-#### tabs.createSplitView()
+#### tabs.createSplit()
 
 Requires a pair of tabIds referring to distinct tabs, rejects otherwise.
 
@@ -127,14 +126,17 @@ At the bare minimum, two eligible tabs, with tabIds specified in order, should
 be able to join a split view, reflected in `Tab.splitViewId` and two
 `tabs.onUpdated` events.
 
-When tabs are not adjacent, or even in order, the implementation MAY try to
-move the tabs to the given order and join them.
+When tabs are not adjacent, or even in order, the implementation MUST reject
+the method call. This constraint may be relaxed later, see "Future Work".
 
 Upon returning, either the given tabs are in the same split view, or the method
 rejects without having changed the tab state.
 
+A browser supporting more than two tabs in a split view MAY accept up to that
+many tabIds (instead of at most 2).
 
-#### tabs.separateSplitView()
+
+#### tabs.unsplit()
 
 Separates the tabs in a split view, "unsplitting" the split view.
 
@@ -161,6 +163,9 @@ tabs MAY separate tabs automatically.
 Browsers that treat split views as one unit MAY move multiple tabs, exposed via
 `tabs.onMoved`, `tabs.onDetached`, `tabs.onAttached` as needed. In that case,
 the `index` specifies the desired position of the split view after moving.
+
+*Informational note: Chrome separates splits upon move, Firefox preserves the
+split by default, see [discussion on PR 1019](https://github.com/w3c/webextensions/pull/1019#discussion_r3406856534).*
 
 
 ### New Permissions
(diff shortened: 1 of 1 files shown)


proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+- Reversing tabs in a split view.
+
+### Known Consumers
+
+Tab managers, offering tab management functionality through a custom UI, e.g.:
+
+- [Tree Style Tabs](https://addons.mozilla.org/firefox/addon/tree-style-tab/),
+  tracked at https://github.com/piroor/treestyletab/issues/3806
+- [Sidebery](https://addons.mozilla.org/firefox/addon/sidebery/), tracked
+  at thttps://github.com/mbnuqw/sidebery/issues/2327
```
kzar:
```suggestion
  at https://github.com/mbnuqw/sidebery/issues/2327
```

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+- Any other `index` is rejected, because they fail to meet the requirement of
+  the tabs being neighbors.
+
+If `url` is not specified, the browser MAY open a custom split view-specific
+page instead of the default new tab page.
+
+
+#### tabs.createSplitView()
+
+Requires a pair of tabIds referring to distinct tabs, rejects otherwise.
```
sanpii, line 114: When I « Add Split View » in my browser (firefox or chromium), I have only only tab. The other one is an empty tab (`about:opentabs` in firefox). Is it possible to have the same behaviour if I pass only one `tabId`?

And what appends if I pass 3 or more tabs?

Rob--W: > When I « Add Split View » in my browser (firefox or chromium), I have only only tab. The other one is an empty tab (`about:opentabs` in firefox). Is it possible to have the same behaviour if I pass only one `tabId`?

With the existing `tabs.create` API with the new `splitWithTabId` option, described above. Reason for relying on that method is to avoid duplicating the "create new tab" functionality.

> And what appends if I pass 3 or more tabs?

Error, because we cannot create a split view with more than 2 tabs right now.

carlosjeurissen:
@Rob--W considering your concern about future-proofing the API, would it make sense to allow `splitViewId` in `tabs.update()` to attach/detach a specific tab to/from a splitview? For which we then have to determine what should happen to the splitview. It could either resort to the split-view initial state which offers users to do something with the "split", or simply dissolve the splitview, in case it could be a replacement for `tabs.separateSplitView()`.

Rob--W:
> @Rob--W considering your concern about future-proofing the API, would it make sense to allow `splitViewId` in `tabs.update()` to attach/detach a specific tab to/from a splitview? For which we then have to determine what should happen to the splitview. It could either resort to the split-view initial state which offers users to do something with the "split", or simply dissolve the splitview, in case it could be a replacement for `tabs.separateSplitView()`.

With the number of tabs in a split view being limited to 2, adding a tab to a split view automatically implies kicking another tab out of it. The mechanisms by which that should happen are not obvious (should tabs move around? Which one should be kicked out? etc). It may be possible to define something meaningful for limited scenarios, e.g. tabs already adjacent to each other. But then we'd basically duplicate the functionality of `createSplitView` - and I am trying to not duplicate functionality unless necessary. We can consider adding this later if we'd like, but for now I'm trying to keep the proposal minimal to only the parts that are not already available through existing extension APIs.

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+Enabling extensions to offer the built-in functionality as documented at:
+
+- https://support.mozilla.org/en-US/kb/split-view-firefox
+- https://support.google.com/chrome/answer/16971124
+
+Specifically:
+
+- Creating a split view from an existing tab, which creates a new split view
+  consisting of the specified tab, plus a browser-native UI page where the user
+  can choose the tab to adopt in that split.
```
rdcronin, line 46: the browser-native UI page here is unique to Firefox, I think (or at least, we don't have this in Chrome).  Is this something we should include here?

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+
+- Creating a split view from two existing tabs, putting two splits together.
+
+- Separating split views.
+
+- Reversing tabs in a split view.
```
rdcronin, line 52:
```suggestion
- Reversing the order of tabs in a split view.
```

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+ https://issues.chromium.org/issues/510252106
+
+## Specification
+
+### Schema
+
+```javascript
+browser.tabs.create({
+  // ... existing properties, including: url, index, active
+  splitWithTabId: number,   // Existing tab to pair with
```
rdcronin, line 73: bikeshed: We could also call this "splitWith", since tabId is more of the type rather than the name of the value.  And its shorter.  But splitWithTabId does align more with openerTabId.  Thoughts?  (I don't feel strongly.)

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+}) : Promise<Tab>           // Returns new tab in split with given tab
+
+browser.tabs.createSplitView(
```
rdcronin, line 76: createSplitView() is so long.

My instinct here is to actually just go with chrome.tabs.split().  This has good consistency with e.g. chrome.tabs.group().  That said, I know that it might also sound confusing (like you're "splitting" a single tab in half or something.)

Could also be createSplit()?

Again, I don't feel *super* strongly here, but it does kinda stand out as a long / clunky name compared to some others.

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+  tabIds: Array<number>     // Two tabIds.
+) : Promise<number>         // Returns splitViewId
+
+browser.tabs.separateSplitView(splitViewId) : Promise<void>
```
rdcronin, line 80: and chrome.tabs.unsplit() : )

(or "removeSplit()".  But I kinda like just "split()" and "unsplit()")

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### Behavior
+
+#### tabs.create() with splitWithTabId option
+
+First, validate the options and ensure that it would create a tab next to the
+tab specified by `splitWithTabId`. The method SHOULD reject if it cannot create
+a split view (see createSplitView). If a tab was created, the created tab must
+be returned as usual, even if the split view cannot be created.
```
rdcronin, line 91: Is the expectation then that the developer check splitViewId to see if the split view was actually created?

This sounds like something that would be very easy to get wrong.  Do we need to support this case?  Or should we instead say that it's a bug if the browser create a tab that was supposed to be split and then didn't split it?

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+  tabIds: Array<number>     // Two tabIds.
+) : Promise<number>         // Returns splitViewId
+
+browser.tabs.separateSplitView(splitViewId) : Promise<void>
+```
+
+
+### Behavior
+
+#### tabs.create() with splitWithTabId option
```
rdcronin, line 86: kind of obvious, but we should also say that it has to be in the window with the tab it's being split with.

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+- mismatching "groupId" states
+- mismatching "splitViewId" states
+- mismatching "windowId" states
+
+At the bare minimum, two eligible tabs, with tabIds specified in order, should
+be able to join a split view, reflected in `Tab.splitViewId` and two
+`tabs.onUpdated` events.
+
+When tabs are not adjacent, or even in order, the implementation MAY try to
+move the tabs to the given order and join them.
```
rdcronin, line 131: it seems like this is something we should specify more than "MAY".  Should we move, or should we not?

My $0.02: I don't think we should reorder / move them.
* We don't allow tabs that aren't in the same window (or other conditions, as above), so to create a new split, an extension may already need to do some tab management.
* It seems a bit arbitrary to assume the extension wants to do this move, but not make the same assumption that it would want to e.g. change the pin state, etc.  And making all those assumptions is a lot of assumptions.
* Bundling these operations increases the likelihood that the function "half works" -- for instance, it might move the tabs, but not create the split view, violating the "don't change the state if it didn't succeed" bit below.
* In theory, browsers may not require splits to be next to each other in the future (why should they need to be?).  Preemptively moving these means that if we changed that in the future, it causes significant behavior changes for extensions.
* We could potentially change this behavior in the future to _allow_ the case of moving tabs for the split. We can't go the other direction and _stop_ allowing it.

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+`tabs.move` does not receive new options, but is worth mentioning due to their
+potential impact on split views. The defining characteristic of a Split view is
+the adjacency and coupling of (two) tabs.
+
+The `tabs.move` method is specified to move one or more tabs, to end at the
+position given by `index`. This expectation is not always compatible with the
+request of moving a split view.
+
+If the `index` behavior of `tabs.move` weighs more strongly, then the move of
+tabs MAY separate tabs automatically.
```
rdcronin, line 159: MAY here also seems underspecified.

What do each of the browsers do today?  If we already have consistent behavior, is it worth specifying?  (If we don't, then MAY is fine)

rdcronin (changes requested):
Thanks, Rob!  Excited to see this!

proposals/split_view_modification.md
Rob--W: By "browser-native UI page" I meant a page potentially different from the default new tab.

In Firefox it is "about:opentabs", in Chrome is is `chrome://new-tab-page/` (internally "chrome://tab-search.top-chrome/split_new_tab_page.html", opened at
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/browser_commands.cc;l=1608-1612;drc=c9ae8a85e83fc4c4cb5531945ade924b940a816f ). If you want to see for yourself, open the context menu on a tab and use the "Add tab to new split view." option.

proposals/split_view_modification.md
Rob--W: If the type is the same I prefer `splitWithTabId`, as it is also consistent with `openerTabId` like you said.

I initially considered (and actually started with) an object type, e.g. `splitWith: { tabId }`, but ultimately flattened that to minimize complexity in `tabs.create`.

E.g. one of the options I considered for the `splitWith` option was the relative placement, but I dropped that in favor of reusing the existing `index` property.

proposals/split_view_modification.md
Rob--W: > createSplitView() is so long.
> 
> My instinct here is to actually just go with chrome.tabs.split().  This has good consistency with e.g. chrome.tabs.group().  That said, I know that it might also sound confusing (like you're "splitting" a single tab in half or something.)

At the design phase my mind went through all that you wrote here :)
I started with `split()` and `unsplit()`, but decided against "split" because it sounds like the tab would be split in pieces.

> Could also be createSplit()?

I'm not fully opposed to it, but given that the ID is called splitViewId I prefer to be consistent and use "splitView" as the word, so "createSplitView".

For `tabs.create`, I nevertheless went with "splitWithTabId" because "splitViewWithTabId" was a mouthful and could too easily be confused with "splitViewId".

proposals/split_view_modification.md
Rob--W: I like unsplit, which was my first choice, but dropped it because "split()" was not a good option in my opinion.

I don't like "removeSplit" because it is ambiguous and can be mistaken for removing the split view and all of its tabs.

I went with "separate" because that is unambiguous and already the verb that Chrome and Firefox use in split view context menus.

proposals/split_view_modification.md
Rob--W: > Is the expectation then that the developer check splitViewId to see if the split view was actually created?

The expectation is that we reject with an error if a split view cannot be created, expressed as "SHOULD". Since "SHOULD" is not a "MUST", I decided to also specify what the (reasonable) behavior is. `tabs.create`'s primary purpose is to create a tab, and if it creates a tab it should return it, to prevent "leaking" tabs that the extension does not know about. Naturally, if the split view failed to create, then its splitViewId should be `-1`.

> This sounds like something that would be very easy to get wrong.  Do we need to support this case?  Or should we instead say that it's a bug if the browser create a tab that was supposed to be split and then didn't split it?

I am willing to turn the SHOULD into a MUST, in which case the situation does not occur. This imposes higher requirements om the browser implementers, which I'd be fine with, if you are too.

proposals/split_view_modification.md
Rob--W: That is already covered by:

> if it cannot create
a split view (see createSplitView).

And in `createSplitView` one of the failure conditions is:

> mismatching "windowId" states

proposals/split_view_modification.md
Rob--W: I'm onboard with rejecting if they are not adjacent. We can always relax later, which I could call out in the future work section.

proposals/split_view_modification.md
Rob--W: Chrome separates the split, and I don't know if there were much thoughts behind this behavior.

Firefox tightly keeps tabs in a split view together, treating then as one unit unless there are clear indicators that they should not be together. The unity of split views was a Product requirement, and followed from the experience with tab groups, where users were surprised that the tab groups they created were ungrouped. Turns out that there was an extension that was unaware of tab groups and as part of reshuffling, the tabs were taken out of a tab group.

I documented the options (Chrome vs Firefox) along with the decision and expected behaviors for `tabs.move` at https://bugzilla.mozilla.org/show_bug.cgi?id=2016751

Tai-ch0802:
+1 on this proposal — thanks for putting it together, @Rob--W.

I left a detailed real-world use case over at #967 ([comment](https://github.com/w3c/webextensions/issues/967#issuecomment-4699306240)): a Chromium side-panel tab manager that currently has to block drag-reordering of individual split-view tabs because `chrome.tabs.move()` silently resets `splitViewId` to `SPLIT_VIEW_ID_NONE`. Both `tabs.createSplitView()` and the `tabs.move` "move as one unit" behavior described here would unblock us.

proposals/split_view_modification.md
Rob--W: @rdcronin Is Chrome willing to adopt Firefox's behavior, where tabs in a split view move together by default?

A Chrome extension developer expressed their preference for Firefox's behavior (as part of a larger comment describing their use case, at https://github.com/w3c/webextensions/issues/967#issuecomment-4699306240)

> From a consumer's perspective, the most valuable guarantee would be that the "move as a unit" behavior is at least available and discoverable (vs. Chrome's current always-unsplit), so extensions don't have to detect the unsplit after the fact and stitch the split back together.

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+#### Use Cases
+
+Enabling extensions to offer the built-in functionality as documented at:
+
+- https://support.mozilla.org/en-US/kb/split-view-firefox
+- https://support.google.com/chrome/answer/16971124
+
+Specifically:
+
+- Creating a split view from an existing tab, which creates a new split view
```
rdcronin, line 44:
```suggestion
- Creating a split view from a single existing tab, which creates a new split view
```

proposals/split_view_modification.md
rdcronin: Ack

proposals/split_view_modification.md
rdcronin: Ack; I'm fine with this

proposals/split_view_modification.md
rdcronin: I'd slightly prefer createSplit() -- the "view" is somewhat implied here (we don't say "tab view", even though you obviously need a view for the tab) and like you said, it aligns with "splitWithTabId" -- but I'm fine either way

proposals/split_view_modification.md
rdcronin: IMO, if we go with createSplit(), unsplit() then (still) makes sense -- it doesn't strike me as being as confusing as split().

proposals/split_view_modification.md
rdcronin: As we discussed, the ideal case shouldn't create a tab if a split can't be created -- but in all likelihood, it may still be possible (because e.g. creating tabs can be an async operation, and state may have changed in between the creation of the tab and the creation of the split that then makes the split infeasible).  This should be _rare_ and should be an edge case, but it's likely one we need to handle.

Given that, I'd still lean towards saying that if a tab is created but a split view cannot be created, we should error.  I don't want developers to need to double-check API operations to see if, if they say they succeeded, they _really_ succeeded.  If part of the API call fails, we should error and let the developer know.

If they _really_ care about tab creation, they can always determine it in other ways (tabs.onCreated, querying tabs afterwards, etc).

proposals/split_view_modification.md
```diff
@@ -0,0 +1,213 @@
...
+tab specified by `splitWithTabId`. The method SHOULD reject if it cannot create
+a split view (see createSplitView). If a tab was created, the created tab must
+be returned as usual, even if the split view cannot be created.
+
+The returned `Tab.splitViewId` MUST reflect whether the tab successfully joined
+the split view.
+
+Tabs in a split view are neighbors of each other, and as such the new tab MUST
+be adjacent to the specified `splitWithTabId`. The `index` property affects the
+relative positioning:
```
rdcronin, line 98:
```suggestion
relative positioning (the following assumes a left-to-right browser UI):
```

proposals/split_view_modification.md
rdcronin: I'd prefer not to adjust behavior here (and keep breaking the split):
* If the developer wants to move the tab with the split, they can include both tab IDs (and then that should succeed).
* If the developer doesn't "know" about split tabs (backwards compatibility), then IMO breaking the split is safer.  Moving multiple tabs as a byproduct of a single call to move a single tab could look like a bug.
* While browsers, today, require tabs in a split to be adjacent in the tab strip, there's no fundamental reason this would need to remain true.  If that changes in the future, then we could stop breaking the split, and that seems more consistent than stopping moving both tabs (because developers might be relying on keeping splits adjacent).

rdcronin (changes requested):
Thanks, Rob!  Largely looks good, % open comments.

Rob--W:
Chromium has started an implementation of the feature, judging by the activity at https://issues.chromium.org/issues/456257896

Still behind a flag, but:

- `splitWithTabId` in `tabs.create` at https://chromiumdash.appspot.com/commit/af6ed01c33bb25c857b21ec172611d2c9ea1dacc
- Not landed, still under review: `tabs.createSplit` at https://chromium-review.googlesource.com/c/chromium/src/+/8231727 and `tabs.unsplit` at https://chromium-review.googlesource.com/c/chromium/src/+/8257716

The `splitWithTabId` property is directly in my proposal without objections.

The `createSplit` / `unsplit` names differ from the proposed `createSplitView` / `separateSplitView` names. There was still ongoing discussion on it at:
- `createSplit`, originally `createSplitView`: https://github.com/w3c/webextensions/pull/1019#discussion_r3406806876
- `unsplit`, originally `separateSplitView`: https://github.com/w3c/webextensions/pull/1019#discussion_r3406808948

I'm fine with these names and will adjust the proposal accordingly.

rdcronin:
> Chromium has started an implementation of the feature, judging by the activity at https://issues.chromium.org/issues/456257896

Yep -- we chatted about this briefly, Rob : )

> There was still ongoing discussion

Just to clarify, we were building it out with the understanding that names might change (since the feature is behind a flag) -- our intention wasn't to say "these must be the final names".  But, since you're fine with them... ; )

Thanks, Rob!  I'll stamp once the PR's updated.

rdcronin:
We're getting ready to enable support for these in Chromium; @Rob--W ,  can we update and merge this PR?  (Do you have any concerns with us enabling support in Chromium?)

proposals/split_view_modification.md
```diff
@@ -0,0 +1,221 @@
...
+
+
+### Behavior
+
+#### tabs.create() with splitWithTabId option
+
+First, validate the options and ensure that it would create a tab next to the
+tab specified by `splitWithTabId`. The method SHOULD reject if it cannot create
+a split view (see createSplit). If a tab was created, the created tab MAY be
+returned as usual, even if the split view cannot be created.
```
Rob--W, line 91: Github dropped the thread because of a line modification, but there was still an open discussion on the desired behavior when a tab was created but not moved into a split: https://github.com/w3c/webextensions/pull/1019/changes/f48105dda453668daa16cdade5e6f463b607f19d#r3406817268

proposals/split_view_modification.md
Rob--W: The primary purpose of `tabs.create` is to create a tab, so if it succeeds at that, IMO the tab should be returned, so that the extension can clean up the resource if needed.

In the (hopefully!) unlikely event that a tab is created without a split view, there are several options:
- (my expressed preference) return tab with splitViewId `-1`
- (your suggestion) reject with error without mentioning the tab; instead encourage extensions to listen to `tabs.onCreated` and `tabs.query`
- (other option nobody suggested yet) reject with error, close tab to leave no inconsistencies behind.

As an extension developer I'd rather have `splitViewId` returning `-1` (to handle that special case), than to add a boatload more async code to try and intercept this special case. Most likely, an extension developer would not bother, and a browser tab would be left behind.

In response to your feedback, I relaxed "MUST" to "MAY" (diff: https://github.com/w3c/webextensions/pull/1019/changes/77ebbbde2c2f9459ee8cc564202f81a9cb99f3cb ). This allows for either interpretation to be followed. Does that look good to you?

Rob--W (commented):
I have updated the PR. I addressed most feedback; there is still [one open thread (on a previous diff](https://github.com/w3c/webextensions/commit/f48105dda453668daa16cdade5e6f463b607f19d#r3406817268)), about the edge case of `tabs.create` creating a tab but not a split view and the desired behavior.

There is also another [open thread](https://github.com/w3c/webextensions/pull/1019/changes#r3406856534) on the `tabs.move` behavior, but I think that it was a clarification request, which got clarified, but still relevant to other readers, so I left the thread open (and added a link to it from the end of the section).

proposals/split_view_modification.md
rdcronin: I think that leads to a bit of a slippery slope -- yes, the primary point of the API is to create a tab, and it did create a tab... but it was also to create a tab _with the properties I (the developer) specified_.  If we say "as long as a tab is created, we return it and don't throw an error, and you should validate the rest", then it means that developers potentially have to validate every property on the created tab to determine if the browser did the requested action.  Worse, this is something you have to do in every case, because you _don't know_ when there was an error.  By contrast, if we throw an error, then the developer can easily just retry the action (or ignore it and fail gracefully, or something else).  True, a "partially-correct" tab may have been created, but if a developer _really_ cares about that, they can clean it up themselves.  I think I would rather code something like:

```
try {
  const tab = browser.tabs.create({...})
} catch (e) {
  handleTabCreationFailure();
}
```

Than:

```
try {
  const tab = browser.tabs.create({...});
  // Tab was created, but maybe not in the way we wanted.
  if (!verifyTabProperties(tab)) {
    handleTabCreationFailure();
  }
} catch (e) {
  handleTabCreationFailure();
}
```

Of course, this is all case-by-case, and different developers or different use case may have different desired outcomes.  And, this should also be exceedingly rare -- whenever possible, browsers should fail atomically.

I'm okay with changing the must to MAY if we also include the other possible action:

```
In the (rare) case that a tab was created but could not be added to a split view, the browser may return the created tab or reject the promise with an error.
```

Or

```
In the (rare) case that a tab was created but could not be added to a split view, the behavior is undefined and left up to the browser.
```

rdcronin (approved):
Thanks, Rob!

Rob--W:
I have applied the remaining requested change, and renamed the document to use kebab case following the agreement to use kebab case for anything that has not already been merged, at https://github.com/w3c/webextensions/pull/1091

I'll squash and merge the PR if nobody objects.

H4M4CHi-ttr:
Congrats on shipping the Split View API in Chrome 155!

I noticed that relative sizing is still listed under **Future Work**, and I wanted to add a concrete use case for it now that the initial API shape has shipped.

I'm working on an extension that opens secondary/reference content in a native Split View—for example, keeping a search results page, documentation page, or feed in a narrower pane while opening the destination in the larger pane.

For this kind of workflow, the ideal split is often closer to 30/70 or 35/65 than 50/50. Without an API for the initial ratio, the user has to manually resize every newly created split.

Something along these lines would cover the basic use case:

```js
const splitViewId = await browser.tabs.createSplit(
  [sourceTabId, destinationTabId],
  { ratio: 0.35 }
);
```

Here, `ratio` could represent the share assigned to the first tab in `tabIds`, avoiding assumptions about left/right positioning or RTL layouts.

Being able to update an existing split later would also be useful, for example:

```js
await browser.tabs.updateSplit(splitViewId, {
  ratio: 0.35
});
```

but I think initial-size control alone would already unlock useful extension workflows and could be considered separately from runtime resizing.

Since relative size is already mentioned in Future Work, would it make sense to track this in a follow-up issue/proposal after #1019 is merged?

Rob--W:
@H4M4CHi-ttr Please create a new ticket for the discussion, linking https://github.com/w3c/webextensions/issues/967. It's a topic worth discussing, but this is a PR that I will merge now and is not blocked by the discussion you're proposing.

PR #1022, opened 2026-06-04
Update SecureStorage API proposal

nebbles:
Updated SecureStorage API proposal with clearer justification for the need and update schema specification and behaviour.

This has been drafted as a '2' to make it easier to read than diffing on the existing secure storage proposal from 4 years ago.

My thanks to @oliverdunk and @simpleplaces for earlier drafts of this proposal.

### DIFF
--- a/proposals/secure-storage-2.md
+++ b/proposals/secure-storage-2.md
@@ -1,0 +1,375 @@
+# Proposal: browser.secureStorage
+
+A WebExtension proposal to allow the secure storage of data such as sensitive key material, in platform-agnostic way without relying on native installations.
+
+**Author:**
+
+- Ben Greenberg (Aglide) - [@nebbles](https://github.com/nebbles), [email](mailto:ben@aglide.com)
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:**
+
+- Christian R. (1Password) - [@simpleplaces](https://github.com/simpleplaces/), [email](mailto:smp@pengin.systems)
+- Oliver Dunk (1Password) - [@oliverdunk](https://github.com/oliverdunk), [email](mailto:oliver@1password.com)
+
+**Created:** 2026-06-04
+
+**Related Issues:** &lt;urls>
+
+## Champions
+
+## Motivation
+
+### Objective
+
+Extensions such as credential/password managers, messaging apps, and cryptocurrency wallets handle sensitive user data, that are commonly encrypted. A common choice is to only persist encrypted versions of the data to places like localStorage and IndexedDB, or persist to a server after encrypting using a key that is only ever available client-side. The decryption keys needed to use this data at runtime are stored only in extension memory, meaning that to access the data after a browser restart, the user must be prompted for whatever the key is derived from, usually a long and hard-to-type password.
+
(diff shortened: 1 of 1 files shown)


nebbles:
(Re: IPR CI check) I have now linked my GitHub account to my W3C membership account.

proposals/secure-storage-2.md
```diff
@@ -0,0 +1,375 @@
...
+
+**Author:**
+
+- Ben Greenberg (Aglide) - [@nebbles](https://github.com/nebbles), [email](mailto:ben@aglide.com)
+
+**Sponsoring Browser:** Google Chrome
+
+**Contributors:**
+
+- Christian R. (1Password) - [@simpleplaces](https://github.com/simpleplaces/), [email](mailto:smp@pengin.systems)
```
oliverdunk, line 13: I know 1Password had expressed interest over the last year in submitting an updated proposal. Have you coordinated with them, or would this be an alternative proposal to anything they are working on?

proposals/secure-storage-2.md
```diff
@@ -0,0 +1,375 @@
+# Proposal: browser.secureStorage
+
+A WebExtension proposal to allow the secure storage of data such as sensitive key material, in platform-agnostic way without relying on native installations.
+
+**Author:**
+
+- Ben Greenberg (Aglide) - [@nebbles](https://github.com/nebbles), [email](mailto:ben@aglide.com)
+
+**Sponsoring Browser:** Google Chrome
```
oliverdunk, line 9: Was there a discussion around this? If not, I suspect we would be unlikely to be the sponsoring browser, as we previously haven't considered this as something that would be feasible on our medium term roadmap. It might be worth chatting to @kiaraarose as I know Apple have been more interested in shipping this sooner.

proposals/secure-storage-2.md
```diff
@@ -0,0 +1,375 @@
...
+
+This method gets information about the secure storage implementation.
+
+#### browser.secureStorage.set()
+
+Stores a key-value pair in secure storage.
+
+If the ID has already been used, the value will be overwritten.
+
+If `uvRequired` is set, the user will be required to verify their identity before the value can be retrieved. If user verification is not available, this will reject with an error.
```
oliverdunk, line 183: Could you clarify the specific requirements here in more detail? For example, would a PIN set using Windows Hello be sufficient? It's fine if we decide that it is unintentionally left to the browser or operating system but we should be explicit about that.

oliverdunk (changes requested):
Hi Ben! Thanks so much for the work here. As an overarching question, could you clarify why you chose to create a second proposal document over updating the existing one? I'm not opposed to that in principle, but I don't see many differences here. It seems like we could have reached a similar result by replacing the `authentication` enum in the existing proposal with the `uvRequired` flag.

proposals/secure-storage-2.md
nebbles: I've been in contact with the 1Password team at the end of last week. More details below.

proposals/secure-storage-2.md
nebbles: That's good to know - and will certainly update this in future proposal updates

proposals/secure-storage-2.md
nebbles: Thanks for this feedback - I'll make sure to get it into the updated proposal wording.

nebbles:
Hi @oliverdunk thanks for taking a look! I had drafted it as a '2' just for easier reading in the diff view (with the intention of deduplicating the docs later). 

I'm closing this PR for now as, since opening it, I've also been in touch @simpleplaces and @bbruneau and we're going to collaborate on the proposal to align to a single outcome, combining both our use cases together. 

Our plan is to iterate the proposal draft over at https://github.com/simpleplaces/webextensions/blob/secure-key-cache-revamp/proposals/secure-cache.md which was discussed at the April 2026 Face-to-face Meeting, before raising it back here for the WECG.

Issue #1051, opened 2026-07-24
Amendments to Proposal: browser.test API


justinlulejian:
The [Proposal: browser.test API](https://github.com/w3c/webextensions/blob/main/proposals/browser_test_api.md) was created about 1 year ago and has set the stage for standardized testing methods for the [extension WPT tests](https://github.com/web-platform-tests/wpt/tree/master/web-extensions). It is now working quite well with Chrome, Safari, and Firefox all running the tests.

However, as I've created a few extension API tests to exercise the `browser.test` API I've noticed some limitations. In some cases I've needed to massage the test to work with `browser.test` than the other way around. I'd like to propose some new methods and standardize some that are in the ["Future Work" section](https://github.com/w3c/webextensions/blob/main/proposals/browser_test_api.md#future-work) of the proposal to address the limitations.

 # Methods

**`browser.test.assertRejects(promise, expectedError, message)`**
Asserts that a given `Promise` rejects with a specific `error.message`. If it does not, or if the `error.message` doesn't
match `expectedError` (if defined), the test fails.

**Parameters**
- `promise` (Promise)
- `expectedError` (string | RegExp, optional)
- `message` (string, optional)

**`browser.test.runWithUserGesture(fn)`**
Runs the provided function in the context of a user gesture.
> *Note: This might be what was originally referred to by `withHandlingUserInput` in the Future Work section.*

**Parameters**
- `fn` (function)

**`browser.test.succeed(message)`** / **`browser.test.fail(message)`**
Immediately marks the current test as passed or failed, optionally with a custom message. This is helpful in cases where returning a `Promise` or `undefined` is less obvious than explicitly indicating success or failure.

**Parameters**
- `message` (string, optional)

# Properties

**`browser.test.isUserGestureActive`**
Set to `true` when the current extension code is running in the context of a user gesture. Otherwise set to `false`. This is helpful in the case of validating that an event notification is only fired when processing a user gesture.

Rob--W:
@oliverdunk I have not changed the Chrome labels, but I think that you'd probably lean towards [supportive: chrome](https://github.com/w3c/webextensions/labels/supportive%3A%20chrome)?

Next step is to update https://github.com/w3c/webextensions/blob/main/proposals/browser_test_api.md, which is already referenced as the "canonical" documentation, from https://github.com/web-platform-tests/rfcs/blob/main/rfcs/web_extensions.md
There is some ambiguity on what `runWithUserGesture` should return, would be nice to clarify that.

[...]

PR #1057, opened 2026-08-05
proposal: Add amendments to browser.test API (#1051)

justinlulejian:


proposals/browser_test_api.md
```diff
@@ -79,6 +79,32 @@ match `expectedError` (if defined), the test fails.
...
+**`browser.test.assertRejects(promise, expectedError, message)`**
+Asserts that a given Promise rejects with a specific `error.message`. If it does not, or if the `error.message` doesn't match `expectedError` (if defined), the test fails.
+
+**Parameters**
+- `promise` (Promise)
+- `expectedError` (string | RegExp, optional)
+- `message` (string, optional)
+
+**`browser.test.succeed(message)`**
+Immediately marks the current test as passed, optionally with a custom message. This is helpful in cases where returning a Promise or `undefined` is less obvious than explicitly indicating success.
```
xeenon, line 94: This differs from what we do in Safari/WebKit. `browser.test.succeed(message)` is just an alias for `assertTrue(true, message)`. We have `browser.test.notifyPass(message)` as the "marks the current test as passed" method. IIRC, this is what Chrome and Firefox did too last I checked when we added these.

Should WebKit have `succeed` alias to `notifyPass` instead?

proposals/browser_test_api.md
```diff
@@ -79,6 +79,32 @@ match `expectedError` (if defined), the test fails.
...
+
+**Parameters**
+- `message` (string, optional)
+
+**`browser.test.fail(message)`**
+Immediately marks the current test as failed, optionally with a custom message. This is helpful in cases where returning a Promise or `undefined` is less obvious than explicitly indicating failure.
```
xeenon, line 100: Similar to https://github.com/w3c/webextensions/pull/1057/changes#r3724127353. We have `browser.test.fail(message)` as an alias for `assertTrue(false, message)`. `browser.test.notifyFail(message)` is "marks the current test as failed" method.

Should WebKit have `fail` alias to `notifyFail` instead?

proposals/browser_test_api.md
```diff
@@ -91,14 +117,20 @@ The promise:
...
 - Return a promise that rejects
 - Trigger an assertion failure
+- Call `browser.test.fail()`
 
 **Parameters**
 - `tests` (array of functions)
 
+### Properties
+
+**`browser.test.isUserGestureActive`** (boolean)
```
xeenon, line 137: WebKit currently has `isProcessingUserGesture`. We will need to add support for the new name before we can adopt this in WPT.

proposals/browser_test_api.md
```diff
@@ -12,7 +12,7 @@ Enable testing using the `browser.test` runner and assertion API.
 
 **Created:** 2025-07-17
 
-**Related Issues:** web-platform-tests/rfcs/pull/219
+**Related Issues:** web-platform-tests/rfcs/pull/219, #1051
```
Rob--W:
```suggestion
**Related Issues:** web-platform-tests/rfcs/pull/219, https://github.com/w3c/webextensions/issues/1051
```

Markdown files do not linkify `#num`

proposals/browser_test_api.md
```diff
@@ -79,6 +79,32 @@ match `expectedError` (if defined), the test fails.
...
+Immediately marks the current test as failed, optionally with a custom message. This is helpful in cases where returning a Promise or `undefined` is less obvious than explicitly indicating failure.
+
+**Parameters**
+- `message` (string, optional)
+
+**`browser.test.runWithUserGesture(fn)`**
+Runs the provided function in the context of a user gesture.
+
+**Parameters**
+- `fn` (function)
```
Rob--W, line 109: Should we document a return value (whatever the function returns?). I mentioned that at the last meeting and https://github.com/w3c/webextensions/issues/1051#issuecomment-5133679526

I don't mind not returning anything, but then someone always needs to do something like:

```javascript
await new Promise(resolve => {
  browser.test.runWithUserGesture(
    () => resolve(browser.permissions.request(perms))
  );
});
```

Opposed to:

```javascript
await browser.test.runWithUserGesture(
  () => browser.permissions.request(perms)
);
```

FWIW, our (differently named) test helper in Firefox does currently NOT return the value, to discourage the reader from thinking (incorrectly) that returning a promise would extend the user gesture beyond the current run of the event loop.

proposals/browser_test_api.md
```diff
@@ -79,6 +79,32 @@ match `expectedError` (if defined), the test fails.
 - `expectedError` (string | RegExp, optional)
 - `message` (string, optional)
 
+**`browser.test.assertRejects(promise, expectedError, message)`**
+Asserts that a given Promise rejects with a specific `error.message`. If it does not, or if the `error.message` doesn't match `expectedError` (if defined), the test fails.
```
Rob--W, line 83: Let's also document that `assertRejects` returns a Promise that settles when `promise` settles. That is needed to enable us to rely on `await browser.test.assertRejects`.

proposals/browser_test_api.md
justinlulejian: Added a returns section to this method. Chrome usually combines it internally with chrome.test.succeed internally so that's why I omitted it, but I agree it's more ergonomic to the function's result to not have to wrap it in a promise each time. The refactor and impl would be pretty simple to implement in Chrome.

proposals/browser_test_api.md
justinlulejian: I had switched this since it was suggested as a good alternative. Chrome also implements it as `isProcessingUserGesture`. Do other browser have a strong position on this? Otherwise I'm happy saving the effort and keeping it as `isProcessingUserGesture`.

proposals/browser_test_api.md
justinlulejian: In Chrome browser.test.notifyPass/browser.test.notifyFail are for the whole suite of browser.test.runTests, whereas browser.test.succeed in for the specific test case it's run in. For Safari, is browser.test.notifyPass/browser.test.notifyFail for the whole suite of tests or each individual test case? If the latter, then yes you could have browser.test.succeed/browser.test.fail just alias to browser.test.notifyPass/browser.test.notifyFail.

proposals/browser_test_api.md
justinlulejian: See above.

proposals/browser_test_api.md
kiaraarose: We return the result in WebKit

proposals/browser_test_api.md
kiaraarose: i'm fine with `isUserGestureActive`. but as Tim said, we would need to adopt in WebKit and wait for that change to be picked up in Safari Tech Preview before using it in the wpt test

proposals/browser_test_api.md
kiaraarose: > In Chrome browser.test.notifyPass/browser.test.notifyFail are for the whole suite of browser.test.runTests

what about outside the context of runTests? how does notifyPass/notifyFail behave?

kiaraarose: > IIRC, this is what Chrome and Firefox did too last I checked when we added these.

Firefox also does an alias: https://searchfox.org/firefox-main/source/toolkit/components/extensions/child/ext-test.js#299-305

Rob--W: In Firefox we rarely use `browser.test.succeed` (either explicit assertion messages or we log, e.g. with `browser.test.log`). We use `browser.test.fail` to signal (unexpected) failures in the test when an `assertXXX` variant does not make more sense).

In wpt, with explicit control flow through `async function` in `runTests`, I don't think that we need a method to short-circuit the execution. It can result in strange control flow questions:

```javascript
browser.test.runTests([
  async function test_one() {
    browser.test.assertEq(1, 1, "one is one");
    browser.test.succeed("Got there");
    browser.test.assertTrue(true, "What is this?"); // <---
  },
  async function test_two() {
    browser.test.assertEq(2, 2, "two is two");
  },
]);
```

Rob--W (approved):
Looks good to me. I opened a bug in Firefox to track the implementation of the proposed amendments: https://bugzilla.mozilla.org/show_bug.cgi?id=2073186

There is still an open thread on what `succeed`/`fail` should do, I'll let @kiaraarose or @xeenon and you finalize the discussion on that part.

PR #1059, opened 2026-08-10
Define Browser interface

oliverdunk:
Defines the Browser interface so that it can be referenced from other specifications which define WebExtension APIs.

Also, update heading levels for other content to correctly create table of contents.

### DIFF
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -46,13 +46,15 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   <h2 id="browser-interface">{{Browser}}</h2>
 
   <pre class="idl">
-    [Exposed=(Window,Worker), LegacyNoInterfaceObject]
+    [Exposed=(Window,Worker)]
     interface Browser {
       // Intentionally left blank.
     };
   </pre>
 
   Note: This interface is intentionally left blank to allow APIs to be exposed by other specifications.
+
+  Issue(1065): In all known implementations, {{browser}} is currently a plain object and does not have the semantics expected from a WebIDL interface.
 </section>
 
   <section>
--- a/specification/window.browser.bs
+++ b/specification/window.browser.bs
@@ -42,11 +42,23 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
   and are left up to User Agent implementation.
 </section>
 
+<section>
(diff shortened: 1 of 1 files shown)


patrickkettner (approved):

patrickkettner:
Shape LGTM, but there’s linting error that needs sorting

domfarolino:
I spoke with @oliverdunk about this a little offline. It sounds like we want is interface objects annotated with `[LegacyNoInterfaceObject]` as this CL does, and probably `[Replaceable]`-annotated attributes on those objects, so that we can support what I understand extensions needs to support regarding attribute overwriting. But we should be absolutely sure we need this combination of things. For example:

 - Do browser *definitely* not expose `window.Browser` as a non-constructible interface? If they expose this interface, then we don't need `[LegacyNoInterface]` objects. Same goes for other properties of `[LegacyNoInterfaceObject]`. Do existing browsers all match the prototype exposure quirks of LNIO, like https://webidl.spec.whatwg.org/#ref-for-LegacyNoInterfaceObject%E2%91%A0%E2%91%A1 and others?
 - We should also verify that all browsers match the attribute assignment semantics that https://webidl.spec.whatwg.org/#Replaceable gives us. For example, the shadowing and deletion semantics in the example under this section: https://webidl.spec.whatwg.org/#Replaceable. Does that example exemplify the implementation that all browsers ship?

Finally, regarding the failing validation step, if we're hell-bent on preserving `[LegacyNoInterfacObject]` semantics in our spec, then we'll have to add `VALIDATE_WEBIDL: false` to https://github.com/w3c/webextensions/blob/main/.github/workflows/deploy.yml#L52. But it's strongly recommended that we try and align implementations to normal interface semantics, if we can. Over a period from 2018–2021, we removed *all* references to LNIO in all modern specs, which required some implementation changes as I understand it, so it'd be great if we could do the same for extensions. If not, we can live with it. We just to disable webidl validation.

patrickkettner:
LGTM 

For my own edification - what is the benefit of defining replacing `object` with `Browser` if the latter is just an empty object?

oliverdunk:
> For my own edification - what is the benefit of defining replacing `object` with `Browser` if the latter is just an empty object?

Using a named interface means we can extend it with the `partial interface` syntax in our main specification, and Bikeshed will be able to link it all together :)

PR #1091, opened 2026-09-15
Add preferred filename format for proposals

carlosjeurissen:
This PR adds a section to the proposal instructions clearly stating the preferred filename format to use. Considering most of our repository filenames are kebab-case, this is what is included in this PR. However this can be updated if we land on a different filename format.


Rob--W (approved):

xeenon (approved):

dotproto (approved):

carlosjeurissen:
@dotproto thanks for merging!

PR #1092, opened 2026-09-16
Add Node.js CI workflow for testing

nexoracontrol-ops:
This workflow installs node dependencies, builds the source code, and runs tests across multiple Node.js versions.


bershanskiy:
This PR is clearly spam.

